PR #5, opened 2021-05-13
docs: describe proper monorepo usage

### LINKED ISSUE #4: `create-solid` does not support monorepo with yarn workspaces and hoisted deps
Using `create-solid` to setup a new project within a monorepo that uses [yarn workspaces](https://classic.yarnpkg.com/en/docs/workspaces/) and hoisted dependencies (which is the default) does not appear to work.

See reproduction here:
- https://gitlab.com/john.carroll.p/solidstart-monorepo-issue

Running `yarn install` and then `yarn dev` results in the following warning and the app does not serve

```
You appear to have multiple instances of Solid. This can lead to unexpected behavior.
computations created outside a `createRoot` or `render` will never be disposed
```

Usage within a monorepo without hoisted dependencies seems to work fine. However, I would like for all projects within the repo to use the same solidjs dependency.

### npm Workspaces

Testing with [npm v7 workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces) also seems to have the same error, though I'm less familiar with npm workspaces so it's unclear if they are expected to work similarly to yarn workspaces.

jorroll:
This PR, along with solidui/vite-plugin-solid#12, fixes #4 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,3 +11,36 @@ npm init solid@next
 npm install
 npm run dev
 ```
+
+## Monorepo & `project.json` `"workspace"` support
+
+If you are using Solid Start within a monorepo that takes advantage of the `package.json` `"workspaces"` property (e.g. [yarn workspaces](https://classic.yarnpkg.com/en/docs/workspaces/)) with hoisted dependencies (the default for yarn), you must include `solid-start` within the optional `"nohoist"` workspaces property.
+
+- _In the following, "workspace root" refers to the root of your repository while "project root" refers to the root of a child package within your repository_
+
+For example, if specifying `"nohoist"` options from the workspace root (i.e. for all packages):
+
+```json
+// in workspace root
+{
+  "workspaces": {
+    "packages": [
+      /* ... */
+    ],
+    "nohoist": ["**/solid-start"]
+  }
+}
+```
+
+If specifying `"nohoist"` options for a specific package using `solid-start`:
+
+```json
+// in project root of a workspace child
+{
+  "workspaces": {
+    "nohoist": ["solid-start"]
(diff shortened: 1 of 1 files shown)


ryansolid:
To be honest I sort of intend to change how to set the entry. Relying on /node_modules is super hacky. I was doing it for now but I expect this to change before even we reach beta. I can add this for now. But more than likely I will remove it in the next couple weeks.

jorroll:
Makes sense

README.md
```diff
@@ -10,4 +10,39 @@ cd my-app
 npm init solid@next
 npm install
 npm run dev
-```
\ No newline at end of file
+```
```
devinrhode2, line 13: https://github.com/solidjs/solid-start/pull/5#issuecomment-840304696
> To be honest I sort of intend to change how to set the entry. Relying on /node_modules is super hacky. I was doing it for now but I expect this to change before even we reach beta. I can add this for now. But more than likely I will remove it in the next couple weeks.

I wish there was a "git blame" for the future, i.e. what changed this code next?

For now we'll check manually:
1. Search docs ("README.md") for "monorepo": https://github.com/solidjs/solid-start/blame/main/README.md

Conclusion: Monorepo docs are here: https://github.com/solidjs/solid-start#monorepo--projectjson-workspace-support :)

devinrhode2:
Not really an issue with SolidStart, but I figured I'd cross-link an issue I'm having using SolidStart in an Nx monorepo here: https://github.com/nxext/nx-extensions/issues/983

TL;DR: Vite is serving up the initial request for `entry-client.tsx` as un-transpiled source code

PR #14, opened 2021-07-29
fix port config for dev

vsbyvi:
Hi!
Noticed that `3000 port` was hardcoded for `solid-start dev` :(
Here is an option to pass a port from CLI params to Scripts area.

P.S.: This is just a suggestion for improvement, so I'm open for any comments.

### DIFF
--- a/packages/start/components/Scripts.tsx
+++ b/packages/start/components/Scripts.tsx
@@ -10,7 +10,7 @@ function getFromManifest(manifest) {
 
 export default function Scripts() {
   const isDev = import.meta.env.MODE === "development";
-  const { manifest } = useContext(StartContext);
+  const { manifest, port } = useContext(StartContext);
   return (
     <>
       <HydrationScript />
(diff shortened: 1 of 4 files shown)


ryansolid:
I see.. When I moved to the new entry point I hardcoded that.  Yeah this looks good.  I might need to tweak a couple other areas but this makes a lot of sense.

vsbyvi:
On second thought, maybe this solution is a little bit over-engineered and it's easier to just leave a relative links:
```
<script type="module" src="/@vite/client" $ServerOnly></script>
<script
  type="module"
  async
  src="/node_modules/solid-start/runtime/entries/client.jsx"
  $ServerOnly
></script>
```

On the other hand, the way of working with a base href is still unclear to me. I'm struggling with a docker-compose config for a local development (mono rep with many frontend apps), so it's important to pass a port and a base href. But I've noticed that all source routes created from manifest files are absolute-path relative links (with a leading slash and without any base prefix from config) and in this case adding a `<base />` tag to html headers makes no impact. So imho it's important to understand the way of working with a base href now to maybe reuse this way of passing prop to add a base prefix to source links. What do you think?

ryansolid:
I agree. I went with #18 which did the simpler approach. Thanks.

This is mostly a dev concern. The Router can set different base url references. I might just need to expose some things.

PR #34, opened 2022-01-04
Missed dependency in solid-start module

AdvTechnoKing:
solid-start using rollup-route-manifest without rollup

image

### DIFF
--- a/packages/start/package.json
+++ b/packages/start/package.json
@@ -17,6 +17,7 @@
   ],
   "dependencies": {
     "node-fetch": "^2.6.6",
+    "rollup": "^2.0.0",
     "rollup-route-manifest": "^1.0.0",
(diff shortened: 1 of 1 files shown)


atk:
But it is using vite, which loads rollup already.

AdvTechnoKing:
I use pnpm, which does not load modules that are already on the computer, so it only puts the modules that are specified in package.json.

AdvTechnoKing:
It is not recommended to depend on modules of other packages, because they can be changed at any time

atk:
> I use pnpm, which does not load modules that are already on the computer, so it only puts the modules that are specified in package.json.

Other people are using npm or yarn. It should work either way. Also, it could be a conscious decision to let vite manage the version of rollup.

PR #36, opened 2022-01-08
feat: support extensions provided to solid plugin used for file system routing, eg. mdx

nksaraf:
The pull request has a few additions (if you want me to break this up more.. tell me).

1. `pnpm` as package manager instead of npm workspaces
2. `/examples` directory for live examples to play with and add to for different use cases. We start with a `basic-ts`, and `with-mdx-ts` examples
3. Routing configuration based on virtual modules. The initial motivation is to allow more configurable routing, for example, `vite-plugin-solid` accepts a list of extensions to enable the JSX transform. The use cases are things like MDX and single-file components. We want to be able to write our top level routes using such files as well, eg. `/src/pages/about.mdx`. To enable this, we change the pages and data loading from plain glob imports to imported virtual modules that we can dynamically control the contents of using the vite plugin and the configuration we get. This PR adds an initial implementation that uses the `extensions` option passed to `vite-plugin-solid`, to also enable those extensions as valid page files.

I think the routing can be more configurable ALA https://remix.run/docs/en/v1/api/conventions#routes. I am working on this on top of this infrastructure. I can now control how the routes are resolved. We can add a `routes` config in the vite-plugin and then while generating the `solid-start/routes` virtual module, we populate it accordingly after we evaluate the user's routing config from the options passed to the vite-plugin. 

The `examples` add a lot of noise to the PR, they are essentially the current `/template/app-ts` from `create-solid`


ryansolid:
Thank you so much. This is great stuff.

PR #40, opened 2022-01-10
fix URL to file path conversion on windows in plugin.js

fictitious:
on windows, `fileUrl.pathname` for absolute file URLs produces a pathname with leading `/` before the drive letter, like

```
 /C:/Users/me/src/solid-next/node_modules/solid-start/plugin.js
```

which is invalid path on windows, see the discussion about this in https://github.com/nodejs/node/issues/23026

Fortunately, `readFileSync` accepts file URLs as well, so there's no need for URL -> path conversion

### DIFF
--- a/packages/start/plugin.js
+++ b/packages/start/plugin.js
@@ -42,7 +42,7 @@ export default function StartPlugin(options) {
           ].join("|")})");`;
         } else if (id === SOLID_START_ROUTES_MODULE_ID) {
           return fs
-            .readFileSync(path.dirname(new URL(import.meta.url).pathname) + "/routes.js", "utf8")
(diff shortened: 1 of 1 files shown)


fictitious:
btw I just found out that there's `fileURLToPath` used here for the same purpose in [start-node/index.js](https://github.com/solidjs/solid-start/blob/main/packages/start-node/index.js). I don't know which is better - `new URL(relativePath, import.meta.url)` seems shorter, and seems like all node.js file API functions accept URLs.

nksaraf:
I was not aware that node js file system functions take URLs coz I could have sworn I did this way of getting the path coz the URL wasn't working

Issue #45, opened 2022-01-14
HEAD: solid-start/routes.js: ReferenceError: exports is not defined


btakita:
```
ReferenceError: exports is not defined
    at processRoute (file:///home/brian/work/portfoliome/censible-core/vendor/solid-start/packages/start/routes.js:65:44)
    at file:///home/brian/work/portfoliome/censible-core/vendor/solid-start/packages/start/routes.js:82:5
    at Array.reduce (<anonymous>)
    at getRoutes (file:///home/brian/work/portfoliome/censible-core/vendor/solid-start/packages/start/routes.js:80:29)
    at async Context.load (file:///home/brian/work/portfoliome/censible-core/vendor/solid-start/packages/start/plugin.js:24:26)
    at async Object.load (/home/brian/work/portfoliome/censible-core/node_modules/.pnpm/vite@2.7.12/node_modules/vite/dist/node/chunks/dep-0351185a.js:36961:32)
    at async doTransform (/home/brian/work/portfoliome/censible-core/node_modules/.pnpm/vite@2.7.12/node_modules/vite/dist/node/chunks/dep-0351185a.js:51996:24)
```

nksaraf:
This should be fixed by #44

PR #46, opened 2022-01-14
fix: ReferenceError: exports is not defined

btakita:
https://github.com/solidjs/solid-start/issues/45

This PR includes setting the dependency:

`"solid-start": "workspace:0.1.0-alpha.53"`

& adding `*.log` to `.gitignore`


nksaraf:
Yeah this was my bad on the exports stuff. I've also fixed it in my PR #44 and the dependency for solid-start-node

packages/start/routes.js
```diff
@@ -53,27 +53,26 @@ export async function getRoutes({
...
-        full
-      );
+      if (!parentRoute) {
+        routesList.push({
+          src: src,
+          _id: id,
+          path: toPath(id) || "/",
+          componentSrc: src,
+          type: "PAGE",
+          dataSrc: data[full] ? data[full] : exports.includes("data") ? src + "?data" : undefined
```
nksaraf, line 63: ```dataSrc: data[full] ? data[full] : undefined```

packages/start/routes.js
```diff
@@ -53,27 +53,26 @@ export async function getRoutes({
...
+          // loaderSrc: exports.includes("loader") ? src + "?loader" : undefined
+        });
+      } else {
+        processRoute(
+          parentRoute.children || (parentRoute.children = []),
+          src,
+          id.slice(parentRoute._id.length),
+          full
+        );
+      }
```
nksaraf, line 75: putting this inside here.. but prevent non-`.tsx,.ts,.jsx,.js` routes to be includes... eg. `.mdx`, the fix is that dataSrc shouldn't use the exports.. and instead be what I pointed out earlier

PR #48, opened 2022-01-15
feat: server(...) fn to compile away server side code into fetch calls on the client

nksaraf:
DO NOT MERGE!

add server pragma to remove server side code on the client. It introduces a new API for server side only code in SolidStart. The idea is that be it loading data or mutating data.. the core API that interacts with the SolidJS app will be an isomorphic API like createResource or (upcoming) createMutation/createAction. These will interact with promises and no specific data APIs. The server-side only code will be called inside these promises.

I used a bit of code from Next JS where they manually do the dead code elimination after compiling away the server calls. So variables and variables that were referenced and are no longer referenced after some code is compiled away... those will be removed.

Its in a much better to try now without failing the dev. Havent figured out the build step yet.


ryansolid:
We are so going to trample on each other. I've been working on getting generic Request interfaces but I haven't touched the actions know that we will gut them.

nksaraf:
Small writeup about my thoughts the API being proposed:

Since SolidStart is a monolith, and controls both the server and the client, much like other meta-frameworks, we have to provide to ways to write server side functionality like database access, auth, etc. I think we have an opportunity to reimagine the API for server-side only code that is serving data and handling data updates from our components.

### User experience
- **Consistency** Try to never show inconsistent state
- **Prefetching** for routes, both data and code for the data
- **Sane Defaults** that provide the expected experience of a user on the web
- **Progressive enhancement** made easy: Not just for no Javascript but before javascript arrives and hydrates (inspiration: Remix)
- **App-like** feel wherever possible, eg. Caching, offline support, managed loading states
- **Customizability**: control the user experience around the data based on what the developer feels is appropriate
	  - **SWR** for showing state data and refetching in the background
	  - **Transitions** where we show the previous view until the data is there and then you switch the view when data has come in (this is how browsers handle native navigations)
	  - **Optimistic updates** during mutation, where looking at the request being sent to the server, you anticipate the future of certain resources and then reconcile as the backend has done the mutation successfully or with errors.
	  - **Suspense** integration for showing loading states where desired

### Developer experience
- **Colocation** with the routes and Components that use the data.
- **Isomorphic**, the API needs to work in isomorphic code, so the solution should ensure that no server side code is shipped to the client without losing any functionality
- **Easy migration** from an existing client-side or custom SSR app using SolidJS, SolidAppRouter, etc
- **Patterns** introduced that share a mental model between client side apps and SSR'ed apps
- **Type inference** without any additional typing
- **Expose primitives** without too many opinions, allow room for exploration
- **Composobility**, the solution should be able to compose across files and components much like hooks. Not lose DX as you grow bigger and have to refractor things into more directories and files. 
- **Sane defaults** for providing expected UX with high configurability (inspiration: React Query)
- **Middleware** like functionality to manipulate the request and response for adding features without having to do it for every time I have to use the functionality (both client and server-side)
	  - auth
	  - validation
	  - logging
	  - normalized cache (for GraphQL)
	  - subscriptions
	  - deduping

**Integrations** Make it easy to integrate into the best tech that there is for data-related problems:
	  - GET caching
	  - server-side code on the edge
	  - Redis caching
	  - GraphQL
	  - External APIs
	  - Webhooks-based functionality

### Existing solutions:
- API Routes 
- Trpc with some way of exposing the server (eg. API Routes or node server)
- Magic server imports, eg. BlitzJS
- Importing from a certain kind of file, eg. from a specific folder, or with a specific extension

PR #50, opened 2022-01-15
fix unescaped periods in route toIdentifier regex

### LINKED ISSUE #49: [Bug] Routes are not resolved correctly because of an incorrect regex
Some files in the `pages` directory are incorrectly resolved as routes.

For example:

`src/pages/rockets.tsx` --> `/rock.tsx`

This is because the periods (`.`) in the below regex are not escaped.

https://github.com/solidjs/solid-start/blob/596bc2e8c290cb5dc02f9b994a16836e0554ad69/packages/start/routes.js#L22-L24

emme1444:
Escape periods in route toIdentifier regex, which leaves some routes incorrectly resolved.

Fixes #49

### DIFF
--- a/packages/start/routes.js
+++ b/packages/start/routes.js
@@ -20,7 +20,7 @@ export async function getRoutes({
 
   const dataRoutes = fg.sync(`${baseDir}/**/*.data.(ts|js)`, { cwd });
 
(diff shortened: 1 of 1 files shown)


nksaraf:
Looks good to me! We should check if there are any concerns related to windows paths that could create problems.. not sure if there are any concerns.. but just making sure.

ryansolid:
Thanks.

PR #51, opened 2022-01-16
fix ENOENT when spawning vite and imports path resolution on windows

fictitious:
1. On windows, node binaries are usually started using cmd or powershell scripts, which can't be spawned without `shell: true`

2. In routes.js, the intent is to import file using its absolute path. Technically correct way for this is to use file URLs, but vite resolver does not support `file:///` URLs. Also, vite does not seem to recognize absolute paths in windows format. Using absolute posix paths is the only way I could find to get this to work.

### DIFF
--- a/packages/start/routes.js
+++ b/packages/start/routes.js
@@ -1,5 +1,6 @@
 import fg from "fast-glob";
 import fs from "fs";
+import path from "path";
 import { init, parse } from "es-module-lexer";
 import esbuild from "esbuild";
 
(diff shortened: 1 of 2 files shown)


packages/start/routes.js
```diff
@@ -108,9 +109,9 @@ export function stringifyRoutes(routes) {
           i =>
             `{\n${[
               /.data.(js|ts)$/.test(i.dataSrc ?? "")
-                ? `data: ${addImport(process.cwd() + "/" + i.dataSrc)}`
+                ? `data: ${addImport(path.posix.resolve(i.dataSrc))}`
```
nksaraf, line 112: Does this resolve relative to process.cwd() or is configurable?

fictitious: yes `path.resolve` uses `process.cwd` as default if it's not given an absolute path as a base: https://nodejs.org/api/path.html#pathresolvepaths

> If, after processing all given path segments, an absolute path has not yet been generated, the current working directory is used.

PR #61, opened 2022-01-23
fix path resolution for manifest on windows (fix #60)

### LINKED ISSUE #60: windows: route path mismatch with manifest
this is what manifest looks like on windows in `getAssetsFromManifest` :

```
getAssetsFromManifest: route= [ { originalPath: '/', pattern: '', path: '/', params: {} } ]
manifest:  {
  '/C:/Users/me/src/test-solid-start/2022-01-23/src': [
    { type: 'script', href: '/assets/index.1ce99872.js' },
    { type: 'script', href: '/assets/client.99f63873.js' },
    { type: 'style', href: '/assets/index.adcc5497.css' }
  ],
  '*': [ { type: 'script', href: '/assets/client.99f63873.js' } ]
}
```

that `C:/Users/..` path is just '/' on linux

as a result, on windows, no script and style tags are added to the `<head>`

fictitious:
this uses `normalizedPath` from vite to avoid mismatch with file paths which are normalized by vite.

it makes vite an actual runtime dependency, but I think it's ok
to leave it as it is - both dev an peer dependency

### DIFF
--- a/packages/start/plugin.js
+++ b/packages/start/plugin.js
@@ -1,4 +1,5 @@
 import path from "path";
+import { normalizePath } from "vite";
 import manifest from "rollup-route-manifest";
 import solid from "vite-plugin-solid";
 import inspect from "vite-plugin-inspect";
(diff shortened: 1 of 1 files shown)


ryansolid:
Yeah that is fine. As far as I'm concerned this is still build time.

PR #65, opened 2022-01-30
fix windows paths in dev mode

MrFoxPro:
It's only for dev, I did not test it in build mode.
Please, do not forget use node path methods when working with paths

### DIFF
--- a/packages/start/runtime/devServer.js
+++ b/packages/start/runtime/devServer.js
@@ -15,7 +15,7 @@ export function createDevHandler(viteServer) {
     try {
       if (req.url === "/favicon.ico") return;
 
-      const entry = (await viteServer.ssrLoadModule(path.resolve("/src/entry-server"))).default;
+      const entry = (await viteServer.ssrLoadModule(path.resolve("./src/entry-server"))).default;
 
       const webRes = await entry({
         request: createRequest(req),
--- a/packages/start/server/babel.js
+++ b/packages/start/server/babel.js
@@ -3,6 +3,7 @@
 // This is adapted to work with any server() calls and transpile it into multiple api function for a file.
 
 import { INLINE_SERVER_ROUTE_PREFIX } from "./constants.js";
+import nodePath from 'path'
 
 function decorateServerExport(t, path, state) {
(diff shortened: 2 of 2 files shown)


MrFoxPro:
It's only for dev, I did not test it in build mode.
Please, do not forget use [node path methods](https://nodejs.org/api/path.htm) when working with paths

packages/start/runtime/devServer.js
```diff
@@ -15,7 +15,7 @@ export function createDevHandler(viteServer) {
     try {
       if (req.url === "/favicon.ico") return;
 
-      const entry = (await viteServer.ssrLoadModule(path.resolve("/src/entry-server"))).default;
```
LXSMNSYC, line 18: This former line is actually from Vite's SSR example, except the part for the `path.resolve`.
Screen Shot 2022-01-30 at 9 45 06 PM

What happens if we remove `path.resolve` instead?

packages/start/server/babel.js
```diff
@@ -143,11 +144,11 @@ function transformServer({ types: t, template }) {
                   let statement = path.findParent(p => program.get("body").includes(p));
                   let serverIndex = state.servers++;
                   let hasher = state.opts.minify ? hashFn : str => str;
-                  const hash = hasher(
-                    state.filename.replace(state.opts.root, "").slice(1) + "/" + serverIndex
-                  );
+                  const fName = state.filename.replace(state.opts.root, "").slice(1)
 
-                  const route = `${INLINE_SERVER_ROUTE_PREFIX}/${hash}`;
```
LXSMNSYC, line 150: Wasn't this for RPC routing? I don't think we need to use `path` here.

MrFoxPro: hash will contain os-specific delimiters, so path will be like `a\b\c/d/e`. It is useful to use this methods to have correct path joining

packages/start/runtime/devServer.js
ryansolid: I added path resolve to handle both js and ts entry points

PR #80, opened 2022-02-07
example: add tailwindcss integration

muslax:
Forked from "app" example
- with single css
- added about and 404 routes


nksaraf:
Thanks for adding this example!

PR #84, opened 2022-02-19
Fix TypeScript errors in generated project

JoosepAlviste:
Hey!

Previously, running `npx tsc` in a freshly generated project would result in a few type errors.

I added stricter type declarations, especially to `StartContext` from looking at the existing code. I tried my best to understand the shapes of all objects but I'm not totally sure if the type definitions are 100% correct.

Additionally, I added a declaration file for the `~/root` file to exist so that it wouldn't be necessary to use `@ts-expect-error` when importing it in this library's code. This also fixes an error in the generated project because `@ts-expect-error` would fail if the file actually exists and there is no TypeScript error.

One thing I didn't have a good solution for is in the `<Meta>` component where the `ssr()` function returns `{ t: string }` but the `<Assets>` component expects a JSX element as its `children`. So, I just added the `@ts-expect-error` comment there. There's probably a more correct solution but I couldn't think of anything better at the moment.

There are still some errors when running `tsc` inside the `solid-start` project though.

### DIFF
--- a/examples/auth/package.json
+++ b/examples/auth/package.json
@@ -7,13 +7,16 @@
     "db:dev": "prisma migrate dev",
     "db:reset": "prisma migrate reset",
     "db:studio": "prisma studio",
-    "test": "npm-run-all -p test:unit:*",
+    "build-and-start": "npm run build && npm run start",
+    "test": "npm run test:unit && start-server-and-test build-and-start http://localhost:3000 test:e2e",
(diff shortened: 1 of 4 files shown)


nksaraf:
My bad on having closed this. We did actually incorporate your changes, https://github.com/solidjs/solid-start/commit/e4c978d3905eb1938f56e2e084a7adb05f63d4de, I just screwed up the merge a bit. My apologies

PR #86, opened 2022-03-08
update locations, sync server redirects

ryansolid:
This PR breaks apart "/components" and moves stuff around. Introduces no render entries.


packages/start/entry-server/entries.ts
```diff
@@ -0,0 +1,49 @@
+import { renderToStringAsync, renderToStream } from "solid-js/web";
+import { RequestContext } from "./StartServer";
+
+export function renderAsync(fn: (context: RequestContext) => any, options?: any) {
```
nksaraf, line 4: Should probably type this as JSX.Element

nksaraf: And options as:
```
{
    timeoutMs?: number;
    nonce?: string;
    renderId?: string;
}
```

packages/start/entry-server/entries.ts
```diff
@@ -0,0 +1,49 @@
...
+    context.responseHeaders.set("content-type", "text/html");
+
+    return new Response(markup, {
+      status: 200,
+      headers: context.responseHeaders
+    });
+  };
+}
+
+function handleRedirect(context) {
```
nksaraf, line 20: Is this supposed to be unimplemented?

packages/start/entry-client/StartClient.tsx
```diff
@@ -1,21 +1,16 @@
...
-const [request, setRequest] = createSignal(new Request("http://localhost:3000/"));
 
 export default () => {
   return (
     <StartProvider
-      context={{
-        responseHeaders: new Headers(),
-        request: request()
-      }}
+      context={{}}
```
nksaraf, line 13: This has typescript issues, maybe mark as any or @ts-expect-error, but I think we should stub the values out with reasonable defaults. Though probably even better to make it a proxy and throw on the Client if the user tries to access it on the client.

packages/start/entry-client/StartClient.tsx
```diff
@@ -1,21 +1,16 @@
 import { MetaProvider } from "solid-meta";
 import { Router } from "solid-app-router";
-import { StartProvider } from "./StartContext";
+import { StartProvider } from "../entry-server/StartContext";
 // @ts-expect-error
```
nksaraf, line 4: this can now be deleted, should have done it earlier but yeah can be removed now.

nksaraf (commented):
a few fixes here and there, I can clean it up while merging if you want

packages/start/entry-server/entries.ts
ryansolid: It was causing infinite loops. So I left it for now.

PR #91, opened 2022-03-16
example with vitest

atk:


nksaraf:
Do we want to encapsulate some of the logic in a plugin and make it part of every template. It'll really go a long way to make it seamless to start testing.. this setup is not incrementally easy to add to an existing app?
Your thoughts?

atk:
I'd be glad to help doing that, but my knowledge of the plugin structure is rather limited. Also, this is a full example with jsdom, solid-testing-library and jest-dom (including a postinstall script to fix a type reference). Not everyone might want to use all of this. Maybe we can discuss on discord how this should play out.

nksaraf:
Yeah I can help with the plugin structure.. I think for the plugin, if this stuff (jsdom, jest-dom), etc are required for the testing, then we should include in our abstraction.. when would people not want it?

atk:
If you only test server-side rendering or primitives or run the test inside the browser where there already is a DOM (all a matter of configuration), you don't need jsdom. Also, some people like testing-library and jest-dom extensions, others don't.

ryansolid:
Yeah.. although I think there is an argument here for sane defaults.. Mind you I didn't see any testing built into Svelte Kit or Remix.. they must be doing that as an add on. CRA had it which was nice.. I don't know about Next. It's ok to be opinionated here if it is easy to remove.

atk:
I'm fine with this, but we should at least document the moving parts then.

nksaraf:
Lets merge this and then work on an abstraction?

ryansolid:
Yeah I agree. Abstraction will take time and probably not as high priority.

Issue #90, opened 2022-03-16
Files in public folder in build (Node Adapter)


modmanx:
Hello, we deployed new application using solid-start and after doing npm run build, we cannot open files in /public folder. It always gives us fallback route index.js. What could be the cause? We checked other examples (like hackernews-ts) and did not find any difference.

ZeAmanuel28:
Can you mention to which Hosting Provider you are trying to deploy because Solid Start has adapters that can change the output of your code. Did you try to install these adapters? If not try to see these:
https://github.com/solidjs/solid-start/tree/main/packages
https://m.youtube.com/watch?v=O2kFXlNSdn0

If your node.js version is lower than node 14, try to install the latest or the LTS version.

modmanx:
Hello, we are currently doing Dockerfile build on Cloud Run. We have same problem locally if we do `npm run build` and `cd` to dist and run `node index.js`

We are using solid-start-node.

modmanx:
Hello again, we managed to fix the issue by moving files in public/ to public/assets, as assets seems to be hard-coded in connection with polka (i have not yet figured out why polka is used)

```
const server = polka().use("/assets",
		comp,
		assets_handler
	).use(comp, render);
```

ryansolid:
Yeah when I first made the adapters I missed making them configurable. More of that stuff has come in over time. I just grabbed Polka because Luke Edwards makes good stuff and I figured since this was abstracted away the need for Express wasn't necessary. But maybe there are reasons to go with Express anyway given it is more known and used.

modmanx:
I understand ;) I was just not sure if I missed something because https://github.com/solidjs/solid-start/tree/main/examples/hackernews-ts does not have public files in assets folder and it works just fine here - https://hackernews.ryansolid.workers.dev/

ryansolid:
This should be fixed in alpha.69

PR #93, opened 2022-03-19
fix start-node not finding public folder

modderme123:
These are the patches I used for solid-start-node in https://github.com/solidjs/solid-site/pull/261

edit: Removing the second rollup call is not doing what I expected, need to rethink

fixes https://github.com/solidjs/solid-start/issues/90


ryansolid:
I fixed this one.. although in a different way. Still have the extra rollup step though.

PR #98, opened 2022-03-26
Added auto detection for adapters

ch99q:
Added zero-configuration detection for adapters.

Now the only thing needed to get started with solid is to drop off `solid()` and install the adapter you need. 🚀

## Implementation
By looking through the current directory for node_modules, then if it's not found, keep looking up the path for the directory. When node_modules are found, look for any directory starting with "solid-start-", ignore the default "solid-start-node"; Then, in case a package is found, apply that name and default to "solid-start-node".

### DIFF
--- a/packages/start/plugin.js
+++ b/packages/start/plugin.js
@@ -300,7 +300,12 @@ function detectAdapter() {
   let adapters = [];
   fs.readdirSync(nodeModulesPath).forEach((dir) => {
     if (dir.startsWith("solid-start-")) {
-      adapters.push(dir);
+      const pkg = JSON.parse(
+        fs.readFileSync(path.join(nodeModulesPath, dir, "package.json"))
+      );
+      if (pkg.solid && pkg.solid.type === "adapter") {
+        adapters.push(dir);
+      }
     }
   });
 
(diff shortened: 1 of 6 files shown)


packages/start/plugin.js
```diff
@@ -285,6 +286,30 @@ function solidStartConfig(options) {
...
 
+function find(locate, cwd) {
+  cwd = cwd || process.cwd();
+  if (cwd.split(path.sep).length < 2) return undefined;
+  const match = fs.readdirSync(cwd).find((f) => f === locate);
+  if (match) return match;
+  return find(locate, path.join(cwd, ".."));
+}
+
+function detectAdapter() {
```
nksaraf, line 297: Is there a way to ensure that this doesn't mistakenly use a package like `solid-start-form-lib` as the adapter? People might come up with package names like that. We can have a system of reading the package.json in any such package and if it has a specific config that tells us it's an adapter, we use it? What do you think?

ch99q: Sounds like a good idea, to place a "solid.adapter": true in the package, I'll try to update the pull request!

nksaraf (approved):

Issue #102, opened 2022-04-16
pnpm build && pnpm start - TypeError: Cannot read properties of undefined (reading 'default')


btakita:
The route jsx files are transpiled after the `const routes = $ROUTES;` routes definition. Explicitly importing the routes tsx files before importing Routes fixes the issue.

```js
  p.then(mod => resolved = mod.default);
```

`TypeError: Cannot read properties of undefined (reading 'default')`

btakita:
For some reason, the transpiled output `Promise.resolve().then(index$1)` does not act the same as `queueMicrotask`. It immediately executes & `index$1` is undefined since it is defined later in the transpiled file.

btakita:
The problem was a top level `await`, which stalled the initial pass in the server index.js file. Since the route modules were defined after the routes array, `index$1` was not defined when the lazy route was run. If there is a top-level await, the route modules will need to be defined before the routes array.

nksaraf:
Looks like stale issue for which we don't a solution yet. If the problem persists, you could reopen this..

PR #103, opened 2022-04-16
pre-import route modules before defining routes array

btakita:
fixes https://github.com/solidjs/solid-start/issues/102


btakita:
I realized that the imports break lazy loading in the routes. I'll try a `queueMicrotask` instead

btakita:
Abandoning this PR due to identifying a top-level await & fixing the issue in my project & not immediately knowing how to add the route modules earlier in SSR.

PR #111, opened 2022-05-05
add ErrorBoundary and Suspense wrappers to tailwind and vitest examples

katywings:
As per [Discord discussion](https://discordapp.com/channels/722131463138705510/910635844119982080/971093152570175488), this adds ErrorBoundary and Suspense to the tailwind and vitest examples and therefore makes these examples consistent with the changes in 67eaa63cfbcb0e94d04617e76eeed354cab017ed.


atk:
Please also add it to the uvu example, which is basically the same as for vitest, but with another test runner.

katywings:
@atk No worries 🙂🙏

ryansolid:
Thank you.

PR #112, opened 2022-05-07
fix windows path handling

ghalle:
I noticed that the vite server was restarting every time I edited a file. I tracked it down to the chokidar file watcher inside `routes.js` giving back a windows style path containing backslashes, while the existing path inside the `routes` object are url/posix style paths only containing forward slashes.

This is a simple fix that converts the file path to a posix file path.


ryansolid:
Thank you

PR #113, opened 2022-05-09
[start-vercel] Update to Vercel Build Output API v3

TooTallNate:
This PR updates the `solid-start-vercel` adapter module to output the [Vercel Build Output API v3](https://vercel.com/docs/build-output-api/v3) format, which is the successor of the currently implemented v2 File System API.


ryansolid:
Thank you!

PR #118, opened 2022-05-19
fix hmr

ghalle:
https://github.com/solidjs/solid-start/commit/a27a3a9f13e506f2a11b80b4f7ea2b91d5cffb76 broke hmr for routes without a dataPath. 

To simplify the issue imagine we have an empty config, we end up comparing `{dataPath: undefined}` vs `{}` which wasn't an issue previously because `JSON.stringify` doesn't support undefined. But the new check using `dequal` considers those 2 objects to be different.

I opted to change the code to not destructure `dataPath` out of the config, so if it is not present it won't create an object with `{dataPath: undefined}`

### DIFF
--- a/packages/start/routes.js
+++ b/packages/start/routes.js
@@ -162,15 +162,14 @@ export class Router {
       if (this.routes[id]) {
         // get old config, we want to compare the oldConfig with the new one to
         // detect changes and restart the vite server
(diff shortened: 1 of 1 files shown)


ryansolid:
Thanks

PR #122, opened 2022-05-23
[WIP] Add api router that supports route params

ghalle:
I needed this for my project so I quickly wrote this. It supports parameters and wildcards. Implemented using a simple tree and recursion.

I'm passing the parameters as a second argument to the handler, could be passed in the context instead.
The naming of certain things also need a review (getApiHandler returning an object with a handler is a bit awkward).

Feel free to modify this or to use something else.

### DIFF
--- a/packages/start/api/index.ts
+++ b/packages/start/api/index.ts
@@ -1,15 +1,122 @@
+import { RequestContext } from "../server/types";
+
 // the line below will be replaced by the compiler with a configuration of routes
 // based on the files in src/routes
 // we have to keep this var so that we can find it even if vite precompiles the code
 // @ts-ignore
 var api = $API_ROUTES;
 
-export function getApiHandler(url: URL, method: string) {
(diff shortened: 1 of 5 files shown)


ghalle:
There's an issue with parameterized / wildcard api routes being matched over non-api routes since the router is unaware of those routes.

Example if you have:
`src/routes/[...api].ts`
`src/routes/index.tsx`

and you navigate to `/`
the less specific api route will incorrectly be matched

One way to fix it would be to also pass the regular routes to the api router and add some sort of skip property to the route. We could then match against it and see that we should skip, returning right away and letting the next middleware handle it.

ghalle:
I went ahead and implemented the solution explained in the previous comment and changed the router to use a score based strategy like `solid-app-router` so the route resolution should be the same for API routes and regular routes.

I also fixed a few more issues related to the index nested layout. If there is an index layout and an API route is inside the index folder it will drop the `/index` like regular routes. The console messages showing the routes also correctly reflect this now and no longer include inaccessible layout pages.

ryansolid:
Thanks

PR #124, opened 2022-05-26
fix: run build in sequence

MarvinRudolph:
Running the build tasks in parallel fails when the vite.config is a `.ts` file. This avoids the missing config file error when generating the output.

Vite generates a `vite.config.ts.js` when the project has `"type": "module"` set in the `package.json` which seems to fail when multiple build tasks are running:
https://github.com/vitejs/vite/blob/main/packages/vite/src/node/config.ts#L728


ryansolid:
Thanks.

PR #133, opened 2022-06-11
Fix README example for server function

alxarch:


alxarch:
Closing, docs seem still WIP, fixes like this can be annoying

PR #135, opened 2022-06-16
server-func: separate browser and node esm exports

### LINKED ISSUE #130: importing solid-start/server causes server side dependencies to be loaded by vite
Steps to reproduce:
1. Have any npm packages that only work with node imported somewhere (pg, argon2, etc..)
2. Import something from `solid-start/server` somewhere.

Minimal example: https://github.com/ghalle/solid-start-test

```
X [ERROR] No loader is configured for ".node" files: node_modules/argon2/lib/binding/napi-v3/argon2.node

    node_modules/argon2/argon2.js:12:12:
      12 │ } = require("./lib/binding/napi-v3/argon2.node");
         ╵             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

4:41:01 PM [vite] error while updating dependencies:
Error: Build failed with 1 error:
node_modules/argon2/argon2.js:12:12: ERROR: No loader is configured for ".node" files: node_modules/argon2/lib/binding/napi-v3/argon2.node
```

The code still mostly works despite the errors but the CSS doesn't get included in the head of the document anymore.

If you comment out the `solid-start/server` related code, everything is back to normal.

ghalle:
Fixes #130

The if checks that were meant to separate the browser and server code weren't being statically removed, I tried multiple variation including the one mentioned in https://vitejs.dev/guide/ssr.html#conditional-logic but nothing worked. I'm not sure if this is a vite bug or not.

Looking for another solution I decided to go with esm [conditional exports](https://nodejs.org/api/packages.html#conditional-exports).

I added an `exports` field to the `package.json` so that vite can load a different entrypoint for the browser and server when importing `"solid-start/server"`

I moved all the server functions related code into their own subfolder and split up the original `serverFunctions.ts` file into a browser and server file.

At first I tried to leave it at that and only split the server-functions import internally, but esm conditional export only work internally if you import using the package name instead of using relative paths and that caused issues with vite. So I then moved the conditional export to the `solid-start/server` entrypoint as that is what is being imported externally by the users of the framework.

I created `server/browser.ts`, `server/server.ts` and `server/shared.ts`, should be pretty self-explanatory.

I also had to create `server/index.ts` as we are still using `"moduleResolution": "node"` and typescript doesn't resolve esm exports with that resolution mode.

I deleted the code below, not sure if it was needed for anything. I ran the `withAuth` example without issue without it.
https://github.com/solidjs/solid-start/blob/235caefe58627eee874c05c6f29bd789b4b369b2/packages/start/server/serverFunction.ts#L56-L65


nksaraf:
Thank you for figuring this out! This is really helpful!

PR #136, opened 2022-06-17
fix createServerResource inside .data files

ghalle:
The babel transform for `createServerResource` was missing for .data files.


nksaraf:
Thank you for figuring this out too! Life-saver!

PR #137, opened 2022-06-17
chore(start): use 'del' for 'DELETE' api route

cstrnt:
Right now it's not possible to have an api route / handler with a delete method because you can't create a variable / export called `delete` in JS/TS because it's a reserved keyword.

Sveltekit has the same issue and they use `del` instead [reference](https://kit.svelte.dev/docs/routing#endpoints-post-put-patch-delete).

This PR brings the same behavior to solid-start


nksaraf:
This is really helpful, was in the backlog.. thanks for figuring it out!

PR #143, opened 2022-06-22
fix hmr for pages with routeData

ghalle:
Currently if you have a page that exports a `routeData` function like this:

```javascript
export function routeData() {}

  return <div></div>;
}
```
the page will reload whenever you make a change to the component.

This PR fixes that by adding a new babel plugin for ?data modules. The plugin creates a hash of the whole source code of the module and then injects some HMR code into the module to compare the hashes, if the hashes don't match the page is reloaded.

I left the code in template form so it's easier to review and tweak for now, but it can be converted into babel-types function calls like the other plugins for some extra performance.

### DIFF
--- a/packages/start/plugin.js
+++ b/packages/start/plugin.js
@@ -8,6 +8,7 @@ import { stringifyApiRoutes, stringifyPageRoutes, Router } from "./routes.js";
 import c from "picocolors";
 import babelServerModule from "./server/server-functions/babel.js";
 import routeData from "./server/routeData.js";
+import routeDataHmr from "./server/routeDataHmr.js";
 import routeResource from "./server/serverResource.js";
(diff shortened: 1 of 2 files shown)


nksaraf:
I thought the current behaviour was that we reloaded when `routeData` changed and we wanted to do HMR. But this PR is enforcing a page reload when we change `routeData`. I might be a little confused..

ghalle:
> Is it actually preventing reloads when the Component changes but routeData doesn't?

Yes exactly that.

nksaraf:
Nice!! Is it tested and working like you would expect?

ghalle:
Yes, this brings it in line with having the `routeData` in a .data.ts file.

PR #144, opened 2022-06-22
Fix css coming in late during dev

nksaraf:


nksaraf:
Fixed in https://github.com/solidjs/solid-start/commit/b83372e2b5095d1a0c2e40f861cf6b237970833f

PR #147, opened 2022-06-22
fix plugin order and simplify hmr logic

ghalle:
Follow up to #143 

I noticed that the page wasn't reloading if I modified a `routeData` function using server functions.

The server transforms happening in the plugins before were hiding the change and causing the hash to be the same even though the code changed.

Changing the plugin order so that the `routeData` separation and `routeDataHmr` happens first fixes that.

I also realized that `import.meta.hot.data` wasn't behaving how I thought it would and it ended up being useless so I got rid of it.

### DIFF
--- a/packages/start/plugin.js
+++ b/packages/start/plugin.js
@@ -219,19 +219,19 @@ function solidStartFileSystemRouter(options) {
         return babelSolidCompiler(code, id.replace("?data", ""), (source, id) => ({
           plugins: [
             [
+              routeData,
+              { ssr, root: process.cwd(), minify: process.env.NODE_ENV === "production" }
+            ],
+            !ssr &&
+              process.env.NODE_ENV !== "production" && [routeDataHmr, { ssr, root: process.cwd() }],
+            [
               routeResource,
               { ssr, root: process.cwd(), minify: process.env.NODE_ENV === "production" }
             ],
             options.ssr && [
               babelServerModule,
               { ssr, root: process.cwd(), minify: process.env.NODE_ENV === "production" }
-            ],
-            [
-              routeData,
-              { ssr, root: process.cwd(), minify: process.env.NODE_ENV === "production" }
-            ],
-            !ssr &&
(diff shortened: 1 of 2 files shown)


nksaraf:
I think changing the plugin order is not that safe because I think we want server functions to actually happen before the `routeData` function stuff because you could have server functions both in `routeData` and the `Component` and we don't want to lose the indexing across both of these since they are part of the same `filename`, the index of the `server` functions is consistent across both.

For ex.
```tsx

export default function Component() {
  const s1 = server(...)
  return <></>
}

export function routeData() {
  const s2 = server(...)
}
```

We want `routeData`'s server function to have index 2 even after compilation which wont happen if we split out the file before processing the server functions.

I too checked the `import.meta.hot.data` stuff that it was not working, and the fix for me was to assign the newHash to `import.meta.hot.data` inside the `accept` function.

ghalle:
You make a good point about server function being used in both the component and the `routeData`.

Do you have any suggestions as to how we could fix this?

nksaraf:
I actually wanna check if the ordering stuff is still relevant. I didn't change the URL to include `index` + another identifier for the server fn.. eg. `/_m/0/routeData`. 

But if we don't find a relevant identifier, we end up using `/_m/0/default`, which we can have conflicts still. And then we name the variables we create based on the index , so `$$serverFunction0` etc. Don't know if that can cause problems. I think it's probably a good idea to add tests for this stuff so we can know when we break it. Since ur PR didn't mess up with the https://github.com/solidjs/solid-start/blob/main/test/sessions-test.ts which actually does have server functions in Component and `routeData`, but they do probably have different identifiers attached so they don't see this problem.

ghalle:
One way of fixing it would be keeping the plugin order the same as it was and have logic in the server function transform to create a hash of the function code that can be appended somewhere in the source, the presence of this hash will then cause the code to be different if it changes.

nksaraf:
I was having the same idea as this and I think this could work! Maybe we try implementing it?

ghalle:
Closing, this is continued in #149.

PR #149, opened 2022-06-23
make the page reload when routeData changes

ghalle:
Supersedes https://github.com/solidjs/solid-start/pull/147

The `routeDataHmrFix` will no longer prevent the page from reloading if some code inside of a server function used in the `routeData` changed.

This is accomplished by modifying the plugin order allowing the `routeDataHmrFix` to compare the hashes of the code prior to the server function transformation.

To be able to alter the plugin order without introducing new problems, a new plugin had to be created:
- Adds a new babel plugin to add metadata to the server functions to make sure they can be uniquely identified after the code gets split up by the `routeData` transform.


nksaraf:
The CI is failing.. could you see why? You can run the tests locally by running `pnpm run test` in the monorepo

ghalle:
Was missing a check to only add the import if it's actually needed, looks like that caused some issues.

ghalle:
I think the `serverResource` transform should probably be moved above the `renameServer` transform so that the server calls can all be renamed but doing so causes an error with the sessions-test. 

The error is
```
ReferenceError: server is not defined
```

The logic added to the server transform to add back the import should cover this so I'm not sure why it's happening.

ghalle:
Should be ok to merge now.

ghalle:
Ended up finding an off by one error using something like this:

```typescript
export function routeData() {
  server((a: number, b: number, c: number) => {
    console.log("handler0", a, b, c)
  })(1, 2, 3)

  return createServerData(async (_, { request }) => {
    console.log("handler1")
  })
}
```

The first `server` call was getting called with the request parameter from the `createServerData` call.

ghalle:
Rebased this, also not quite sure why github displays the diff for `babel.js` like that, there isn't actually that much change, the vscode built-in diff view displays it correctly.

ryansolid:

PR #167, opened 2022-07-02
fix: compression on event stream

### LINKED ISSUE #162: Disable compression for event stream response in node adapter
`app.use(compression({filter: (req) => req.headers['content-type']?.startsWith('text/event-stream')}))`

Similar issue found in remix: https://github.com/remix-run/remix/issues/3232

G3root:
fixes #162

### DIFF
--- a/test/event-stream-test.ts
+++ b/test/event-stream-test.ts
@@ -1,4 +1,4 @@
-import { test } from "@playwright/test";
+import { expect, test } from "@playwright/test";
 import { createAppFixture, createFixture, js } from "./helpers/create-fixture.js";
 import type { AppFixture, Fixture } from "./helpers/create-fixture.js";
 import { PlaywrightFixture } from "./helpers/playwright-fixture.js";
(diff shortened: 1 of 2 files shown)


test/event-stream-test.ts
```diff
@@ -0,0 +1,98 @@
...
+                return;
+              }
+            },
+          });
+          return new Response(stream, {
+            headers: { "Content-Type": "text/event-stream" },
+          });
+        }
+        
+        export default () => {
```
nksaraf, line 58:
```suggestion
        export default function Page() {
```

Just for consistency with other tests

test/event-stream-test.ts
```diff
@@ -0,0 +1,98 @@
...
+          let ref;
```
nksaraf, line 59: Do we need this `ref` stuff, can you do this with a signal again and see if it works, I think it will, and if it doesn't thats an issue too..

nksaraf (changes requested):
Just a few small changes

test/event-stream-test.ts
G3root: i ran the tests in Gitpod. using a signal seems to hang the playwright , when using a ref it work as expected.

nksaraf:
Can you change your PR to point to `solidjs:dev` instead, we are staging changes there before pulling them in main

PR #176, opened 2022-07-04
fix link tags for nested routes

### LINKED ISSUE #171: CSS loaded from `root.tsx` missing for index nested routes during production SSR
Repro: https://github.com/ghalle/solid-start-index-layout-css

1. build and start
2. Go to http://localhost:3000/about

This only applies to index nested routes other than the `/` route.
This only applies if the CSS is loaded from the `root.tsx` moving it to `routes/index.tsx` fixes it.

ghalle:
Fixes #171 

Previously only link tags for the topmost routes were being added to the document. Now all link tags for the ancestors routes will be added. This fixes a CSS flashing issue during prod SSR for nested routes.

A check for index nested routes was also added to correctly match the manifest key, this fixes an issue where the CSS was not getting loaded at all during prod SSR for index nested routes.

### DIFF
--- a/packages/start/plugin.js
+++ b/packages/start/plugin.js
@@ -1,14 +1,14 @@
 import fs from "fs";
 import path from "path";
-import { normalizePath } from "vite";
-import manifest from "rollup-route-manifest";
-import solid from "vite-plugin-solid";
-import inspect from "vite-plugin-inspect";
-import { stringifyApiRoutes, stringifyPageRoutes, Router } from "./routes.js";
 import c from "picocolors";
-import babelServerModule from "./server/server-functions/babel.js";
+import manifest from "rollup-route-manifest";
+import { normalizePath } from "vite";
+import inspect from "vite-plugin-inspect";
+import solid from "vite-plugin-solid";
+import { Router, stringifyApiRoutes, stringifyPageRoutes } from "./routes.js";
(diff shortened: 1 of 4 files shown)


nksaraf:
Do you think we could have a strategy for testing this in our playwright test suite.. maybe we detect if the correct css is present based on the route you visit.. I think playwright should have the APIs for this..

Any ideas or thoughts about this?

ghalle:
Yes as long as there is a way to check if a certain link tag is in the head of the document before any other JavaScript executes it should be fairly simple. Although I've just noticed that sometimes vite merges all the css into one `entry-client.hash.css` instead of keeping it as separate files but it seems to be deterministic so as long as we have a setup where that doesn't happen it should be fine.

nksaraf:
We should inspect the html from a raw request itself, instead of in playwright's headless page. then no javascript executes, and we just check the html, like here: https://github.com/solidjs/solid-start/blob/25406e4ff746066b11123fb511b0d41ba2b0e12b/test/rendering-test.ts#L70

Do you think you could add tests for this in this PR?

ghalle:
Yes, been working on it, but vite's CSS code splitting seems very unreliable, it randomly stops splitting some CSS files when I modify a completely unrelated file.

I'll try again tomorrow.

PR #178, opened 2022-07-05
fix non-index routes containing the word index

ghalle:
Previously a route like the following `src/routes/non-index.tsx` would get converted into `/non-` instead of `/non-index`


ghalle:
Testing for this should be covered by the tests for #171 when they are complete, but we could also have a separate test for this.

nksaraf:
Does this PR still make sense after merging #180 . It looks like this commit is part of that PR..

ghalle:
Closing, this is included in #180.

PR #180, opened 2022-07-08
Pathless Layout Routes

ghalle:
Idea taken from: https://remix.run/docs/en/v1/guides/routing#pathless-layout-routes

Implemented the same as described on the remix page.

Removed support for index nested routes.

### DIFF
--- a/packages/start/root/FileRoutes.tsx
+++ b/packages/start/root/FileRoutes.tsx
@@ -1,10 +1,10 @@
 /// <reference path="../server/types.tsx" />
 
 // @ts-expect-error
-var routeData = $ROUTE_DATA;
+var routesConfig = $ROUTES_CONFIG;
 
-export const fileRoutes = routeData.routes;
-export const routeLayouts = routeData.routeLayouts;
+export const fileRoutes = routesConfig.routes;
+export const routeLayouts = routesConfig.routeLayouts;
 
 /**
  * Routes are the file system based routes, used by Solid App Router to show the current page according to the URL.
--- a/packages/start/root/FileRoutes.tsx
+++ b/packages/start/root/FileRoutes.tsx
@@ -1,15 +1,17 @@
 /// <reference path="../server/types.tsx" />
 
 // @ts-expect-error
-var routes = $ROUTES;
+var routeData = $ROUTE_DATA;
+
+export const fileRoutes = routeData.routes;
+export const routeLayouts = routeData.routeLayouts;
 
 /**
  * Routes are the file system based routes, used by Solid App Router to show the current page according to the URL.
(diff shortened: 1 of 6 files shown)


ryansolid:
Love it.

nksaraf:
Looks like the test failures are from some flaky tests by me, so we can ignore that..

packages/start/routes.js
```diff
@@ -1,11 +1,11 @@
...
-import { init, parse } from "es-module-lexer";
-import esbuild from "esbuild";
 import chokidar from "chokidar";
 import debug from "debug";
 import { dequal } from "dequal";
+import { init, parse } from "es-module-lexer";
+import esbuild from "esbuild";
+import fg from "fast-glob";
+import fs from "fs";
+import path, { join } from "path";
```
nksaraf, line 8: Is this some VSCode extension or prettier plugin that changes the order of the imports? Maybe we should add it as `.vscode/extensions` so everyone follows it

packages/start/root/FileRoutes.tsx
```diff
@@ -1,15 +1,17 @@
 /// <reference path="../server/types.tsx" />
 
 // @ts-expect-error
-var routes = $ROUTES;
+var routeData = $ROUTE_DATA;
```
nksaraf, line 4: nit: Can we change $ROUTE_DATA to $ROUTES_CONFIG, since RouteData already means something else in solid-start, doesnt help to overload the term.. and the singular ROUTE_DATA doesnt communicate the meaning properly..

nksaraf (changes requested):
Also can we rename `test/link-test.ts` to `test/css-link-test.ts`, because it can be confusing with Router links which is more common use of `link`

packages/start/routes.js
ghalle: I think that's a built-in vscode settings I enabled

```json 
"editor.codeActionsOnSave": {
  "source.organizeImports": true
},
```

I think it used to not work with prettier before but they must have fixed it recently.

packages/start/root/FileRoutes.tsx
ghalle: You are right I struggled with coming up with a good name for that one.

Your change has been applied.

ghalle:
> Also does this test #178, you suggested there that we could add tests for that here..

I meant that there are a few tests in the css link tests already testing against pages named like this `/non-index`. But I renamed one of the test to make it a bit more explicit.

packages/start/routes.js
nksaraf: Ohh awesome will enable it by default for the repo!

PR #181, opened 2022-07-10
Add Lambda adapter

thdxr:
This supports both Lambda and Lambda@Edge

Still experimental but gives me enough to start implementing a construct in SST


packages/start-lambda/README.md
```diff
@@ -0,0 +1,20 @@
+# start-netlify
```
suciptoid, line 1: netlify -> lambda

thdxr:
Closing in favor of #477

PR #182, opened 2022-07-12
fix list of published files

ghalle:
The `data` folder was missing, I also reordered everything to match the usual file listing order.

Removed `types.d.ts` and `router` has they don't exist.


nksaraf:
Thanks!

PR #188, opened 2022-07-17
Feat/fix with vitest example

benbender:
Fixes the `with-vitest` example. 

- Removes and replaces the need of patching `@testing-library/jest-dom`.
- conditionally updates the package-resolving to the `browser`-field only if vitest runs.


benbender:
Closed. Opened against the wrong branch (should be dev, not main)

PR #189, opened 2022-07-17
feat: update & cleanup with-vitest example

benbender:
Fixes the with-vitest example.

Removes and replaces the need of patching @testing-library/jest-dom.
conditionally updates the package-resolving to the browser-field only if vitest runs.


atk (approved):

examples/with-vitest/vite.config.ts
```diff
@@ -23,11 +22,8 @@ export default defineConfig({
...
   plugins: [solid()],
-  build: {
-    target: "esnext",
-    polyfillDynamicImport: false
-  },
+  
   resolve: {
-    conditions: ["development", "browser"]
-  }
+		conditions: process.env["VITEST"] ? ["browser"] : [],
```
wjq990112, line 27: why not use `process.env.VITEST` like [official docs](https://vitest.dev/guide/#configuring-vitest)?

benbender: Basically: `process.env` is typed via an index-signature in `@types/node`. I'm using very strict ts-settings. TS forces me to :)

References: 
- https://www.typescriptlang.org/docs/handbook/2/indexed-access-types.html
- https://www.typescriptlang.org/tsconfig#noPropertyAccessFromIndexSignature

examples/with-vitest/vite.config.ts
```diff
@@ -1,33 +1,34 @@
...
+      environment: "jsdom",
+      transformMode: {
+        web: [/\.[tj]sx?$/]
+      },
+      setupFiles: "./vitest.setup.ts",
+      // solid needs to be inline to work around
+      // a resolution issue in vitest:
+      deps: {
+        inline: [/solid-js/]
+      }
```
nksaraf, line 22: maybe we should do this as part of our vite plugin..

examples/with-vitest/vite.config.ts
```diff
@@ -1,33 +1,34 @@
...
-  test: {
-    globals: true,
-    environment: "jsdom",
-    transformMode: {
-      web: [/\.[tj]sx?$/]
+export default defineConfig(({ mode }) => {
+	// Load env file based on `mode` in the current working directory.
+	// Set the third parameter to '' to load all env regardless of the `VITE_` prefix.
+	const env = loadEnv(mode, process.cwd(), '');
+
```
nksaraf, line 10: this function stuff is added complexity.. i wonder if we can do without this.. or have the solid-start plugin take care of some of this if its a good idea to do anyway

benbender: It's the correct way of loading the env-vars in `vite`. We could simplify, but I opted for "correct" over "simple" here as it's an example...

Generally my feeling is that the root-cause should be correctly handled in `solid-testing-library` and it should work in node-envs (which would evaporate the need of most of the setup...)

examples/with-vitest/vite.config.ts
benbender: Tbh, I'm not completely sure why it is needed in the first place, but without it, the imports from `solid-js/web` aren't working correctly...

nksaraf: yeah I think this was figured out by some vitest magician lol.. but its needed.. I think we can do this inside a plugin we include with solid-start.. can you see if putting this whole config in a plugin works that people can install and we will include it by default in solid-start

examples/with-vitest/vite.config.ts
nksaraf: okay makes sense but can we do this also inside a plugin so it doesnt need to be complicated for the user to adopt.. the example is one thing.. but the method should be to incrementally adopt.. so just installing a vite plugin is the easiest thing we can have (and maybe having it as part of solid-start by default)

examples/with-vitest/vite.config.ts
benbender: Hmm, don't think it should live inside a plugin for `solid-start` as none of this is specific to `solid-start`. The main culprit is the "import-magic" for ssr in `solid-testing-library` and imho it should be solved there and not hidden here.

It should be possible though…

examples/with-vitest/vite.config.ts
benbender: I'm feeling like this is leaving the scope of this pr. It might be worthwhile though to revisit ssr-handling in `vite-plugin-solid` for environments which aren't the browser and have trouble with the "magical" import-logic of `solid/web`.

I would propose to merge this and open another issue/discussion where it feels appropriate to those who are more familiar with the underlying logic than I am atm.

examples/with-vitest/vite.config.ts
atk: This is because vitest has two mechanisms of loading imports: using vite or using node. Unfortunately, solid's exports lead to them being imported via both over different routes, which breaks reactivity.

Luckily, a PR to fix this is already open.

PR #198, opened 2022-07-21
Fix typo in server/README.md

benbender:


boehs:
Why was this closed?

benbender:
@boehs I closed it myself because it was opened against the wrong branch ("main" instead of "dev").

PR #205, opened 2022-07-26
fix: solid-start merge babel options

lxsmnsyc:
Probably an oversight. `solid-start`'s vite plugin completely omits the `babel` option for `vite-plugin-solid`.

As for the server transform, I'm not sure where to put it, but for now I'll just use `push`.

### DIFF
--- a/packages/start/plugin.js
+++ b/packages/start/plugin.js
@@ -437,13 +437,13 @@ export default function solidStart(options) {
 
         if (options.ssr) {
           // NOTE push or shift?
-          baseOptions.push([
+          processedOptions.push([
             babelServerModule,
             { ssr, root: process.cwd(), minify: process.env.NODE_ENV === "production" }
           ])
         }
         
-        return baseOptions;
+        return processedOptions;
       },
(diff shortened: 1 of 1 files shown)


nksaraf:
We also have another section where we use the solid `babel` compiler and override it with options:

```typescript
    async transform(code, id, transformOptions) {
      const isSsr =
        transformOptions === null || transformOptions === void 0 ? void 0 : transformOptions.ssr;
      let babelSolidCompiler = (code, id, fn) => {
        // @ts-ignore
        return solid({
          ...(options ?? {}),
          babel: fn
        }).transform(code, id, transformOptions);
      };
 ```

Could you see if we could do something about this as well?

Also, thanks for helping with this!

LXSMNSYC:
@nksaraf which part is this?

nksaraf:
This is the `transform` function from `function solidStartFileSystemRouter(options)`

lxsmnsyc:
Okay so I've fixed it, not sure with the conflicts

PR #210, opened 2022-07-30
docs: provide instructions for pnpm

kkharji:


nksaraf:
Hey! thanks for adding this. But we are not going to merge this since we are going to be removing the docs from here and continuing the effort at this repository: https://github.com/solidjs/solid-docs-next/tree/main/content/start

We would love it if you could contribute there!

PR #215, opened 2022-08-03
fix entry-client url on Windows

ghalle:
Since on windows the URL was starting with `C:\` the browser was treating it as a local file request.

This PR simply adds `/@fs/` in front of the absolute file path.

This will create URLs like the following:

Windows
`/@fs/C:\...\examples\bare\src\entry-client.tsx`

Linux
`/@fs//home/.../examples/bare/src/entry-client.tsx`

The vite dev server seems to handle the extra slash in the Linux URL just fine and the browser seems to take care of converting the backslashes to slashes.

I tested this on Windows and WSL Ubuntu, with an entry-client and without.

### DIFF
--- a/packages/start/root/Scripts.tsx
+++ b/packages/start/root/Scripts.tsx
@@ -32,7 +32,7 @@ export default function Scripts() {
               <script
                 type="module"
                 async
-                src={import.meta.env.START_ENTRY_CLIENT}
+                src={"/@fs/" + import.meta.env.START_ENTRY_CLIENT}
                 $ServerOnly
               ></script>
             </>


nksaraf:
Ahh thanks for this! I wonder why the cloudflare workers are failing.. they have this `required login` error from wrangler.. but cloudflare pages is fine.. maybe do a push to rerun the tests?

nksaraf:
There seems to be something specific about this change causing this.. can you for a commit revert the change and push to trigger tests with the old stuff.. just to confirm this

nksaraf:
Looks like its still failing.. I wonder if it has to do with permissions and you not getting access to the cloudflare tokens.. hmm.. let me check

PR #222, opened 2022-08-05
solid-start: + solid-start-vercel dependency

btakita:
update dependencies


ryansolid:
I think this PR is probably out of date by this point.

PR #225, opened 2022-08-06
Consistent styles in all Examples

### LINKED ISSUE #223: `durable-objects-websocket` example doesn't have tailwindcss/windicss
The example `durable-objects-websocket` doesn't have tailwind/windicss deps installed, but it does use classes and styles from it.

lightyaer:
Fixes #223 

Inside the example [durable-objects-websocket](https://github.com/solidjs/solid-start/tree/main/examples/durable-objects-websocket)

- Adds windicss as dep 
- Add `windi.config.ts`
- Add base styles and windicss-devtools (in case of attributify mode)

Let me know if this is not what we wanted, and need tailwindcss or something else.
I'm happy to make the change.
Thanks in advance


nksaraf:
Heyy really thanks for this PR.. but i think what we might do, and you can update your PR for this.. is just styling our examples with root.css files so people can move off them easily. We will have a windicss starter, but putting it in a websocket example makes it tough to make the examples flexible. Does that make sense?

lightyaer:
Yeah, that totally makes sense.
I looked up the `with-auth` example, and thought windicss is the way to go except when we have `with-tailwindcss` example.

Anyways, I can get this done.

Should we also go ahead and implement this in other examples, if so, can it be in this PR or do we want separate issues and PR's for each example?

lightyaer:
Hey, I have updated the PR to incorporate all stylings in a `root.css` file.

I used element selectors instead of class, so people can just delete the css or keep it without touching jsx.

Also, Let me know you thoughts on this.

> Should we also go ahead and implement this in other examples, if so, can it be in this PR or do we want separate issues and PR's for each example?

nksaraf:
You can do it in this PR.. I think most of the other have some basic styling like this but if they dont it'll be really nice if you could add it!

lightyaer:
Changes done here are, 
- Make styles consistent across all examples except tailwind example.
- Import components from `solid-start/meta` and 'solid-start/root' across examples instead of using bare html meta tags
- In `durable-objects-websocket`, the return from routeData has a utility type `ReturnType` which seems unnecessary and the data returned isn't typed in this case, removing the utility type makes the data typed and doesn't have sqiggly lines.

examples/bare/src/root.tsx
```diff
@@ -1,12 +1,14 @@
 // @refresh reload
 import { Suspense } from "solid-js";
 import { Body, ErrorBoundary, FileRoutes, Head, Html, Meta, Routes, Scripts } from "solid-start";
+import { Title } from "@solidjs/meta";
```
lightyaer, line 4: Is this the right place to import from? 
Wonder why its not clubbed into `solid-start/root`

examples/durable-objects-websocket/src/root.tsx
```diff
@@ -2,18 +2,19 @@
 import { Routes } from "@solidjs/router";
 import { Suspense } from "solid-js";
 import { ErrorBoundary } from "solid-start/error-boundary";
-import { FileRoutes, Links, Meta, Scripts } from "solid-start/root";
+import { Body, FileRoutes, Head, Html, Links, Meta, Scripts } from "solid-start/root";
+import { Title } from "@solidjs/meta";
```
lightyaer, line 6: Should we import in this fashion or import everything from `solid-start`?

examples/durable-objects-websocket/src/routes/index.tsx
```diff
@@ -17,7 +17,7 @@ export function routeData() {
 }
 
 export default function Home() {
-  const user = useRouteData<ReturnType<typeof routeData>>();
+  const user = useRouteData<typeof routeData>();
```
lightyaer, line 20: The ReturnType here causes lack of types for user

lightyaer (commented):
more info on changes made

examples/bare/src/root.tsx
lightyaer, line 4: Should the title be imported from '@solidjs/meta' or `solid-start`

nksaraf: lets import everything from `solid-start` in the examples

examples/durable-objects-websocket/src/routes/index.tsx
nksaraf: Yeah we can remove it

nksaraf:
Also would it be possible to do one last thing? Make sure all the examples have a consistent name in `package.json`? eg. `example-hackernews`, `example-with-auth`, `example-bare`

lightyaer:
@nksaraf 
Changes made are, 
- Consistent names in package.json
- Fix imports in `root.tsx`

PR #233, opened 2022-08-08
Append http headers instead of overwrite

samcfinan:
This implementation appends headers instead of overwrites if a header with the same name is set.

I'm considering a change to the API where `overwrite: true` could be passed if we want to force overwriting.

Would love to get some eyes on this.


nksaraf:
I think these APIs follow the pattern of our `@solidjs/meta` components, i.e they should overwrite by default, and have an API with `append: true` for appending since that's the unusual behaviour with the header instead of the default behaviour when you assign a header.

So

`<HttpHeader name="x" value="y" />` = `responseHeaders.set(x, y)`
`<HttpHeader name="x" value="y" append />` = `responseHeaders.append(x, y)` (we need to think about clean up here)

samcfinan:
Updated!

nksaraf:
Do you think we can handle the `cleanup` side of this as well.. checkout the `onCleanup` function below the code you edited where we cleanup `set` headers. Could you handle `append` headers there too?

ryansolid:
Thanks

PR #244, opened 2022-08-11
Add status code to server error

### LINKED ISSUE #243: Support passing status code to ServerError
When throwing a `ServerError` it would be great to be able to set the `status` of the `Response`.

For example I would like to throw a `401 Unauthorized` error, e.g. like this
```js
throw new ServerError('Unauthorized', { status: 401 })
```

This allows that I don't need to build my own `Response` or `ServerError`.

dertieran:
I just went ahead and implemented it, but feel free to close this PR if you don't want this feature.

Closes #243 

Not sure if the passed status code should be validate, mainly `status >= 400`.


nksaraf:
Can we add some integration tests for this.. we can create a route with a component that renders the error message and status code in the JSX (you can catch the error in an ErrorBoundary)

dertieran:
Sure I just couldn't get the tests working on my system so I didn't add one 🙈
But I will check if I can get it to work 👍🏻

nksaraf:
What was the problem with the tests? maybe I could help u debug.. check this guide out:  https://github.com/solidjs/solid-start/blob/main/CONTRIBUTING.md

dertieran:
I'm running Arch Linux so the install command did not work
```
pnpm --filter solid-start-tests install:playwright

> solid-start-tests@ install:playwright
> playwright install --with-deps
BEWARE: your OS is not officially supported by Playwright; installing dependencies for Ubuntu as a fallback.
Installing dependencies...
Switching to root user to install dependencies...
sh: line 1: apt-get: command not found
Failed to install browsers
Error: Installation process exited with code: 127
```

But I got it working by running the command that got logged in the console, **but I needed to run it inside of the `test` folder.**

```
npx playwright install
```

I also needed to comment out this line, otherwise the test was always skipped.
```js
test.skip(process.env.ADAPTER !== "solid-start-node");
```
*(Not sure if this is needed, I just saw it in the other tests.)*

I added the test in a new file so I can easily test it separately, I hope this is fine.

dertieran:
I also just noticed that you *can't* throw a `ServerError` in an api route, I would have expected it to just work.

Also is there a way to make the page return with the error status code?
I tried `deferStream` but it still returned with a `200`.
*(Not very important, just was curious about it.)*

PR #249, opened 2022-08-14
Data loading docs

nas5w:
Initial outline for the "Data loading" section of the start docs


ryansolid:
Stuff moved forward from here so just cleaning up. Thanks for all your work.

PR #274, opened 2022-09-16
Fixed refresh bug as it would never match response 304 Not Modified

davycoe:
### Cloudflare Pages Adapter

- When refreshing the page currently `entry.js` does not match response 304 and will fail to load assets until a subsequent refresh.

- I've now added `(resp.status === 200 || 304)` so if nothing has changed on the page it will now return assets on page reload.

### DIFF
--- a/packages/start-cloudflare-pages/entry.js
+++ b/packages/start-cloudflare-pages/entry.js
@@ -5,7 +5,7 @@ export const onRequestGet = async ({ request, next, env }) => {
   // Handle static assets
   if (/\.\w+$/.test(request.url)) {
     let resp = await next(request);
-    if (resp.status === 200) {
+    if (resp.status === 200 || 304) {
       return resp;
     }
   }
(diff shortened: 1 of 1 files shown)


davycoe:
this now seems resolved in alpha 105 :+1:

ryansolid:
Oh really so we don't need this PR anymore. Is it safe to close? Thanks for looking into it.

davycoe:
I'm so sorry for messing you around, this is very much required and 105 made no difference.

yhdgms1:
I don't think it will work the way it's supposed to.

```ts
// let's imagine `resp.status` is replaced by `500`

Boolean(500 === 200 || 304) // true
Boolean(500 === 200 || 500 === 304) // false
```

Shouldn't it be `(resp.status === 200 || resp.status === 304)`?!

davycoe:
> I don't think it will work the way it's supposed to.
> 
> ```ts
> // let's imagine `resp.status` is replaced by `500`
> 
> Boolean(500 === 200 || 304) // true
> Boolean(500 === 200 || 500 === 304) // false
> ```
> 
> Shouldn't it be `(resp.status === 200 || resp.status === 304)`?!

yeah, you're right!

Issue #273, opened 2022-09-15
Html not being properly formatted


JonahLund:
At first everything worked great, but then for some reason when i upgraded solid-start everything broke

Running the dev server just shows the raw html output

![image](https://user-images.githubusercontent.com/108274890/190525801-012e20f4-3a7b-4211-aa69-6ce7b1d74c39.png)

So then i thought, huh thats weird, probably some mis-matched version or something. 
Firstly i tried to revert the update, didnt work.
Then i tried to create a new template with create-solid, didnt work.
Then i thought maybe there is some cache that messes things up, so i deleted all cache, reinstalled node, but still didnt work.

This made me confused, so i decided to reset my entire pc thinking that things would work again, because why wouldnt it. So i reset my pc, including all files, fresh windows installation, reinstalled everything, created a new project with the bare template, but nope. Still the same issue.

ghalle:
Lock your solidjs version to `1.5.4` for now, seems like something is broken with `1.5.5` that causes all the `<` to be escaped.

package.json
````json
"solid-js": "1.5.4"
````

JonahLund:
> Lock your solidjs version to `1.5.4` for now, seems like something is broken with `1.5.5` that causes all the `<` to be escaped.
> 
> package.json
> 
> ```json
> "solid-js": "1.5.4"
> ```
> 
> This issue is caused by [ryansolid/dom-expressions@0f8aa58](https://github.com/ryansolid/dom-expressions/commit/0f8aa58b9201de98263c47b64927f03461fe80f0), commenting out [these lines](https://github.com/ryansolid/dom-expressions/blob/8221452594c2eb2d3d593b07759c05047dba17fa/packages/dom-expressions/src/server.js#L339-L343) fixes the issue so it's something to do with that.

Thanks, that solved it!

PR #276, opened 2022-09-16
use ssrElement instead of raw html strings

btakita:
I thought this was a fix for https://github.com/solidjs/solid-start/issues/273 but it proved not to be.

However, in this PR, `ssrElement` replaces the deprecated `ssrSpread`.


nksaraf:
Thanks for finding the issue and adding a fix!!!

I think this is a duplicate of https://github.com/solidjs/solid-start/commit/a0a1c63b1a4adaa1e0a92d0aaa4ba81d03d555ed
So this should be duplicate change.

PR #290, opened 2022-09-29
Add support for files in FormData

yonran:
Pass through enctype=multipart/form-data when submitting an action's Form so that files can be uploaded

In the node adapter, when the request body is multipart/form-data, convert parts that are files to File instead of decoding all the parts as UTF-8 strings. Also add support for multiple values of the same name.


ryansolid:
Thank you

PR #291, opened 2022-09-29
Fix request being undefined for serverData/Actions (#289).

### LINKED ISSUE #289: `createServerData$` request is undefined when invoked from the client side
Originally reported by tomchristensen on discord.

> How does createServerData$ within routeData work in the context of client-side navigation? It seems like the request object is only passed to the fetcher (in createServerData$) on the initial page request, but subsequent client-side navigation to the same page runs the fetcher with undefined for the request object
Example: https://stackblitz.com/edit/github-fk8exj?file=src%2Froutes%2Findex.tsx - on loading the index page, the console logs the request object, but if you navigate to "About" and then back to "Index", the console logs undefined for the request

ghalle:
Fixes #289 

Apply a transform that moves the 2nd parameter into the body of the function and is initialized to the value of `this`
```typescript
createServerData$((_, { request }) => {})
```
```typescript
createServerData$(_ => {
  let { request } = this;
});
```

### DIFF
--- a/packages/start/server/serverResource.js
+++ b/packages/start/server/serverResource.js
@@ -30,9 +30,24 @@ function transformRouteData({ types: t }) {
                 }
               },
               CallExpression(callPath, callState) {
+                function moveEventParam(fn) {
+                  if (fn.params.length >= 2) {
+                    if (!t.isBlockStatement(fn.body)) {
+                      fn.body = t.blockStatement([t.returnStatement(fn.body)]);
+                    }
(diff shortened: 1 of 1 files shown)


nksaraf:
See comment in #289. I don't think this transform will help too much and issue is more to do with `vite` than our server functions. See this commit for fix: https://github.com/solidjs/solid-start/commit/e94ea868216b5f475ca5f9e7034fbe1d9fbb6f64

PR #311, opened 2022-10-13
fix: respect base path, fixes issue#309

atk:


ryansolid:
Thanks

atk:
Happy to help. It gave me the chance to look a bit deeper into solid-start and it was a good experience.

Issue #310, opened 2022-10-13
Multi action result stuck in pending if there was a newer submission


pawelblaszczyk5:
When there are multiple submissions to multi action that are pending, only the latest one will ever resolve. All the previous one will be stuck in pending for ever, even if the corresponding request finished.

Reproduction: https://stackblitz.com/edit/vitejs-vite-kuuenu?file=src/routes/index.tsx

You can quickly click the submit button few times and observe that only the latest one will finish

pawelblaszczyk5:
I found this line in the code and it seems to be responsible for this problem:
https://github.com/solidjs/solid-start/blob/main/packages/start/data/createRouteAction.tsx#L69

I've got an idea what it's responsible for in standard createRouteAction. Though I'm not sure whether it should be there in multi action

pawelblaszczyk5:
Solved in #313

ryansolid:
Hmm.. testing this I remember why I had it this way. To prevent excess thrashing. It basically waits for all settled to refetch the data. Transitions will hold them anyway... we should be clearing pending at this point is probably the issue.

PR #313, opened 2022-10-13
Track requests per submition in multi action

pawelblaszczyk5:
Fix for #310 
Exact solution we discussed on Discord, did some manual testing around retrying etc., seems to work correctly 😃


ryansolid:
Thanks

PR #332, opened 2022-10-18
`server$.fetch` forward headers during SSR

ghalle:
Forward the cookie header from the original page request to the API route.

We could consider forwarding more headers or even letting the user configure which header to forward.

Sidenote:
I was not able to get the tests to run locally, going to test using the CI.

### DIFF
--- a/packages/start/api/internalFetch.ts
+++ b/packages/start/api/internalFetch.ts
@@ -9,21 +9,56 @@ export const registerApiRoutes = routes => {
   apiRoutes = routes;
 };
 
-const FORWARDED_HEADERS = ["cookie"];
+// Headers from the original page GET request that should not
+// be copied over to the fetch request
+const SKIPPED_HEADERS = ["Accept"];
+
+// https://fetch.spec.whatwg.org/#forbidden-header-name
+const FORBIDDEN_HEADERS = [
+  `Accept-Charset`,
+  `Accept-Encoding`,
+  `Access-Control-Request-Headers`,
+  `Access-Control-Request-Method`,
+  `Connection`,
+  `Content-Length`,
+  `Cookie`,
+  `Cookie2`,
+  `Date`,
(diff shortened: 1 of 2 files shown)


ghalle:
Ended up changing the code to instead forward all headers (except `Accept` let me know if you think of any other ones).
This should allow us to access headers like [CF-Connecting-IP](https://developers.cloudflare.com/fundamentals/get-started/reference/http-request-headers/#cf-connecting-ip) for example without having to cover every non-standard header that might exist across different hosting solutions.

packages/start/api/internalFetch.ts
```diff
@@ -8,16 +9,67 @@ export const registerApiRoutes = routes => {
...
+];
+
+export async function internalFetch(route: string, init: RequestInit = {}) {
   if (route.startsWith("http")) {
     return await fetch(route, init);
   }
 
-  let url = new URL(route, "http://internal");
-  const request = new Request(url.href, init);
+  const headers = new Headers(useServerContext().request?.headers);
```
nksaraf, line 46: this `useServerContext()` is probably not going to work correctly. I want to solve this a different way. We need `server.$fetch` to actually go through the whole handler pipeline so people from an API routes can request a UI page and kind of pretend to have rewritten that url.. maybe we can merge and we can build on top... but basically we need an internal fetch that closes over the current call to fetch, like have a reference to the `handler` so that we can run an internal fetch through the whole handler, not just the API routes

nksaraf:
Oooo! this is awesome. We needed this!

boehs:
Thank you ❤️

packages/start/api/internalFetch.ts
ghalle: That sounds good to me, do you have any idea how else we could access the headers other than `useServerContext()`?

ryansolid:

PR #333, opened 2022-10-19
Fix circular dependency by isolating FileRoutes

ghalle:
This had been annoying me for a while, it would cause all sort of issues with imports returning `undefined` during SSR.

To fix the circular dependency `FileRoutes` must now be imported from its own module. We could potentially move it somewhere else as long as it's still isolated.


ryansolid:
I get the need. But are there other options? Inconsistent import is awkward. I'm sure we could find another path but it's still awkward.

One thing I have been gravitating towards is that the top level exports in Start are isomorphic and the Server specific ones have a path.. But this would be an exception to that. I wonder if there is other re-arranging we can do.

ghalle:
The `FileRoutes` can be anywhere as long as that exact import path is never referenced in any of the routes. Otherwise, we end up with a cricular dependency from FileRoutes -> route -> FileRoutes. 

FileRoutes should only ever be referenced once in the `root.tsx` so it should be a pretty minor inconvenience for the user.

At least this is the most straightforward solution that works with vite not sure if there is any other trickery to deal with circular imports.

ghalle:
We could write a babel plugin to rewrite the import and put a dummy `FileRoutes` in the top level import for typescript.
Not ideal but it would solve the inconsistent import.

nksaraf:
Yeah I have got trippd by this too, and always tried to keep rearranging th imports. I would like the export to be available on `solid-start` if we can do that kind of simply.. maybe a simple babel plugin can do it

ghalle:
Pushed a new version with a babel plugin, it should handle every type of import.
https://astexplorer.net/#/gist/6536b9927fb5aad25a91b49a9c85756e/6e5f9d7a07124ef2d814e707927230af5ceb8c07

PR #335, opened 2022-10-20
Movies example

nksaraf:
https://solid-movies.deno.dev


ryansolid:
This will never get merged on its own. But rather as part of dev branch.

PR #343, opened 2022-10-21
better trpc docs

OrJDev:
More detailed documentation on how to use tRPC on solid, 
also rename lowercase get/post to uppercase GET/POST on graphql api routes.

### DIFF
--- a/docs/core-concepts/api-routes.md
+++ b/docs/core-concepts/api-routes.md
@@ -9,21 +9,21 @@ active: true
 
 <table-of-contents></table-of-contents>
 
-While we think that using  [`createServerData$`][createServerData] is the best way to write server-side code for data needed by your UI, sometimes you need to expose API routes.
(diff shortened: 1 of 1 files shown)


docs/core-concepts/api-routes.md
```diff
@@ -162,95 +158,123 @@ SolidStart makes it easy to implement a GraphQL API. The `graphql` function take
...
+```bash
+npm install @trpc/server@next @trpc/client@next zod
 ```
 
+Then, install the required Solid packages
 
-## Exposing a tRPC Server route
+```bash
+npm install @tanstack/solid-query solid-trpc@next solid-start-trpc@latest
+```
```
nksaraf, line 194: I would like the trpc basic example to show how to integrate the API routes with trpc directly, coz the point is to show how to use API routes. We might have a more extensive guide for integrations, but this doc is for the API routes functionality. Want to show how to do the raw integration if people want to do that. We can make a note about these libraries towards the end

PR #375, opened 2022-10-28
fix prisma esm __dirname

QuentinDutot:
Prisma internally uses __dirname which breaks the solid-start start command with solid-start-node

```
dist/server.js => import_path.default.join(__dirname, "../query-engine-darwin");

ReferenceError: __dirname is not defined in ES module scope
```

### DIFF
--- a/packages/start-node/server.js
+++ b/packages/start-node/server.js
@@ -2,11 +2,14 @@ import compression from "compression";
 import { once } from "events";
 import fs from "fs";
 import { readFile } from "fs/promises";
-import { join } from "path";
+import { dirname, join } from "path";
+import { fileURLToPath } from "url";
 import polka from "polka";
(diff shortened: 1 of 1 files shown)


pawelblaszczyk5:
I'm not sure whether it should be fixed here. It's a problem with ESM vs CJS. Probably not only __dirname can make problems here. You can fix it like that: 
https://github.com/pawelblaszczyk5/planotes/blob/4b24f1d74735856bc438ba72c4ac036e9d198a73/apps/web/vite.config.ts#L8

nksaraf:
Yeah I think @pawelblaszczyk5's answer is probably the fix here. That `__dirname` is wrong for most files so its not stable at all.

QuentinDutot:
It does the trick, thanks

PR #386, opened 2022-10-31
start/vite: Configurable hydration event names

atk:
At the moment, hydration only works for input and click events. Other events are not covered. In my case, I want to detect other events, so I want to have this configurable.


ryansolid:
Yeah this makes sense. Although wondering if this should be a props sort of thing rather than a Vite plugin sort of thing since this is a Solid core feature that we are just masking. It is interesting as it is just for event playback for pre hydration capture.. Any reason in favor of doing this via config vs props?

atk:
Sorry, I missed this comment. This was more a question of convenience. Where to put the props?

PR #404, opened 2022-11-07
Fix prisma example

edivados:
Example does not work.

- Call to `getUser` in login.tsx is missing db as first argument
- Production build fails to start without externalizing `@prisma/client`

Trying to start production build without externalizing the prisma client will result in:
prisma


ryansolid:
Thanks

PR #411, opened 2022-11-09
Added robots.txt, meta description and sitemaps

paoloricciuti:
This PR adds robots.txt and meta tag description to improve on the SEO of web dev measure.

It also add an index sitemap and a series of sitemaps for base, movies and tv shows.

This is primarly for education reasons so the actual indexing it's not superimportant. On that note given every sitemap needed the same headers and needed access to the baseURL i've implemented an Higher order function to facilitate the creation of multiple sitemaps. I don't know if this might result in being confusing for a newcomer so if you prefer i can undo this abstraction and follow a more linear path.

### DIFF
--- a/examples/movies/public/robots.txt
+++ b/examples/movies/public/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Allow: /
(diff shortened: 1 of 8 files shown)


ryansolid:
The branches are too out of date now. We are entering the new stage of Beta.

Issue #417, opened 2022-11-10
Missing Dependency in `with-websocket` Example


baristikir:
# Overview
Seems like the `with-websocket` example is missing the `@solid-start/meta` dependency which causes the dev server to crash.

## Output
Error: Cannot find module '@solidjs/meta' imported from '.../node_modules/solid-start/entry-server/StartServer.tsx'

## Reproduce
Simply create a new solid project and run `pnpm i && pnpm dev`
```bash
pnpm create solid
# Template
> with-websocket
```

PR #418, opened 2022-11-10
Fix `with-websocket` Example

baristikir:
Added missing dependency for `with-websocket` example.

Related to https://github.com/solidjs/solid-start/issues/417


ryansolid:
Thanks

Issue #419, opened 2022-11-10
Unable to read `.well-known` directory in routes


jtag05:
`fast-glob` ignores dotfiles by default so the following line will ignore a `.well-known` directory and any other dotfiles when using the `FileRoutes`

https://github.com/solidjs/solid-start/blob/842f2751795a181ec3a9975f18a008344192535a/packages/start/fs-router/router.js#L46

I'm not sure if this was an intentional design decision or not, but I'll put up a PR momentarily in the event that this functionality would be useful to people other than myself

PR #420, opened 2022-11-10
Add dotfile support to FileRoutes

jtag05:
As mentioned in #419 this would enable having dotfile and `.well-known` paths in file based routing


boehs (approved):

ryansolid:
Thanks

PR #431, opened 2022-11-12
Add "with-pwa-csr" example

Vexcited:
Hello, I want to add an example where we use Client Side Rendering with vite-plugin-pwa (no index.html, only root.tsx)

This case needs some kind of "hack" in the vite.config.ts to work, that's why I am making a PR for it here, hoping it can help people like me!


kammeph:
Hi @Vexcited, do you know if I can use the vite-plugin-pwa with server side rendering too?

Vexcited:
> Hi @Vexcited, do you know if I can use the vite-plugin-pwa with server side rendering too?

I honestly didn't tried, this PR is for Client Side Rendering only. It should work out of the box for SSR but needs to be checked.

Vexcited:
And by the way this PR is useless now since CSR works out of the box **ONLY if you have a `/src/root.tsx` (for dev) __and__ a `/index.html`.**

anderssjoberg97:
I found some pointers here @Vexcited 

https://github.com/vite-pwa/docs/pull/26/files

PR #437, opened 2022-11-14
rename .tsx to .ts for non-jsx file

emilbonnek:
This session.tsx file in the with-auth example doesn't use any jsx so I think it should be named session.ts


ryansolid:
Thanks

PR #447, opened 2022-11-16
fix(example:with-prisma): Fixed a breaking bug in with-prisma example

Lalit3716:
The app was breaking because we needed to pass `db` as well as `request` object to the getUser function called in `login.tsx`.

I updated the file and now the app starts without any bug!


ryansolid:
Thanks

PR #449, opened 2022-11-16
Pass file type through to Files in FormData

soorria:
Currently, the MIME types of any uploaded files aren't passed through to the File constructor, so `file.type` is always `''` (empty string).

This fixes that, by passing through what the `multipart/form-data` parser gets, and updates the server file action integration test to check uploaded file MIME types.


ryansolid:
thanks

PR #459, opened 2022-11-18
Remove the need to `nohoist`

amitdahan:
# Summary

Due to a few assumptions made within `solid-start` - usage within a monorepo with hoisting (npm or yarn) enabled can lead to a few things breaking, which is why `nohoist` is the recommended approach.

This PR attempts at removing the need for `nohoist` by handling the following issues I could find.
(There may be more I'm not aware of, these are just the instances I could find)

#### Automatic adapter detection
Previously, we searched for all `solid-start-*` packages with `{ solid: { type: 'adapter' } }`  in `node_modules`.

This can lead to undesired inclusion of adapters in bigger monorepo situations.

This PR changes said behavior to instead only look for known adapter packages in the user's `package.json` deps and devDeps, and later resolve them from his own CWD, which should resolve successfully regardless of hoisting conditions.

#### Running or referencing `vite`
Previously, we assumed `vite` always lives in `node_modules/vite`.

This PR adjusts this to either `require('vite')` or `require.resolve('vite')` from user's CWD.

#### Running `wrangler` in `solid-start-cloudflare-workers` and `solid-start-cloudflare-pages`
Similarly, `wrangler` was also assumed to live in `node_modules/wrangler`.


ryansolid:
I needed to remove this to a branch as it broke the repos doc site. You can find it on the no-hoist branch. Can you help me fix this up so the docs site still works?

amitdahan:
Sure! Will ping when I get it going again

amitdahan:
Finally got around to it - #676

PR #462, opened 2022-11-19
fix vitest example

atk:
Updated the example and fixed the configuration so that the tests will run again.


ruler501 (approved):
Fixed the issues for me.

ryansolid:
Thanks

PR #472, opened 2022-11-23
Add an example using Xata

fabien0102:
Add an example using [Xata](https://xata.io/) as database.

## How to test

```bash
$ pnpm --filter example-with-xata run xata:init
$ pnpm --filter example-with-xata run dev
```


boehs:
This is probably a better candidate for Xata's repository

fabien0102:
Yeah no problem, this was more for discovery, since you had an example with prisma, but no hard feeling 😉

PR #477, opened 2022-11-24
Add AWS Lambda adapter

fwang:
The `solid-start-aws` adapter allows to deploy the server function to both AWS Lambda and Lambda@Edge. This adapter takes a standard Lambda event payload, and returns a standard Lambda response.

For context, we created a `SolidStartSite` construct in [SST](http://github.com/serverless-stack/sst) that allows users to deploy SolidStart apps to AWS. An example looks like:
```ts
new SolidStartSite(stack, "web", {
  path: "path/to/my/solid/start/app"
});
```

If the PR gets merged, we will be able to launch the `SolidStartSite`. After that, I can submit a PR adding examples to the README.md/doc if fit. 

Thanks for reviewing the PR!

^ btw, #181 can be closed


ryansolid:
Thank you. Sorry it has taken so long. I don't have the setup to test this so been pushing it back. But I trust you to be invested to help out with any bug fixes so lets go!

aheissenberger:
@fwang how do you deal with the assets in the root directory of the s3 bucket when using lambda@edge? There is no way to catch them with a simple behavior as I use for the asset folder so I change the adapter to handle this by scanning the list/client root on build and add the list of files to the lambda@edge script to forward this requests to the s3 bucket.

I know you use SST but I use only the server less framework - maybe this can be optional if it is not needed with SST.

```
diff --git a/node_modules/solid-start-aws/entry-edge.mjs b/node_modules/solid-start-aws/entry-edge.mjs
index 8b0faf7..bb6a81b 100644
--- a/node_modules/solid-start-aws/entry-edge.mjs
+++ b/node_modules/solid-start-aws/entry-edge.mjs
@@ -1,8 +1,16 @@
 import "solid-start/node/globals";
 import manifest from "../../dist/client/route-manifest.json";
+import staticFileList from "./static-files.json"
 import server from "./entry-server";
 
 export async function handler(event) {
+  const request = event.Records[0].cf.request;
+  //console.log(`uri:${request.uri}`)
+  if (staticFileList.includes(request.uri.substring(1))) {
+    request.headers.host[0].value=request.origin.custom.domainName
+    console.log('pass thru to S3', JSON.stringify(request))
+    return request
+  }
   const response = await server({
     request: createRequest(event),
     env: { manifest },
diff --git a/node_modules/solid-start-aws/index.mjs b/node_modules/solid-start-aws/index.mjs
index 93b4e79..0fae627 100644
--- a/node_modules/solid-start-aws/index.mjs
+++ b/node_modules/solid-start-aws/index.mjs
@@ -1,7 +1,7 @@
 import common from "@rollup/plugin-commonjs";
 import json from "@rollup/plugin-json";
 import nodeResolve from "@rollup/plugin-node-resolve";
-import { copyFileSync, renameSync } from "fs";
+import { copyFileSync, renameSync, readdirSync, writeFileSync } from "fs";
 import { join } from "path";
 import { rollup } from "rollup";
 import { fileURLToPath } from "url";
@@ -29,6 +29,10 @@ export default function ({ edge } = {}) {
         join(__dirname, edge ? "entry-edge.mjs" : "entry.mjs"),
         join(config.root, ".solid", "server", "index.mjs")
       );
+      if (edge) {
+        const files = readdirSync(join(config.root, "dist", "client"))
+        writeFileSync(join(config.root, ".solid", "server", "static-files.json"),JSON.stringify(files))
+      }
       const bundle = await rollup({
         input: join(config.root, ".solid", "server", "index.mjs"),
         plugins: [
@@ -51,6 +55,7 @@ export default function ({ edge } = {}) {
 
       // closes the bundle
       await bundle.close();
+
     }
   };
 }

```

PR #479, opened 2022-11-25
start: fix types

atk:
I tried to fix the types of the .ts files. Next step would be to move .js files to .ts one after another.


quantuminformation (approved):

ryansolid:
Thanks a ton

PR #484, opened 2022-11-26
Fix wrong description of the entry-server.tsx docs

luca-rath:
I just noticed this small incorrectness while reading through the docs. But anyways, awesome project 🔥


ryansolid:
Thank you

PR #485, opened 2022-11-27
Expose responseHeaders to createServerData$

oscartbeaumont:
This PR exposes a way to set HTTP headers on the response within the `createServerData$` function.

This makes it possible to do:
```typescript
export function routeData() {
  return createServerData$((_, { responseHeaders }) => {
    responseHeaders.append("x-demo", "123");

    return "mydata";
  });
}
```

This will set the headers both on the HTML document (with SSR) and on the fetch request when a client-side navigation is done.

## Existing solutions?

Currently, you can use `useRequest().responseHeaders` to set headers on the response for the HTML document, however, if the user does a client-side navigate a fetch request is done to get that data and when that is done `useRequest()` is an empty object which can't be used to set the response headers.

## Why is this useful?
It is very useful to be able to set cookies on the response. For example, I am trying to do the [double submit cookie](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#double-submit-cookie) technique for CSRF protection while integrating SolidStart with NextAuth.js.


pilcrowOnPaper:
This would make porting [Lucia](https://github.com/pilcrowOnPaper/lucia-auth) (auth library) to SolidStart much easier

marcolanaro:
Is there anything I can do to help getting this merged? 😊🙏
I believe it's an important piece for making production ready apps (auth management).

ryansolid:
We need some more consideration here. I intended to write about this in a state of solid start discussion topic which I spent 6 hours writing yesterday before losing it. I will write it again today if I have a chance. 

Concern here is that server data works in 2 modes. Both initial render when paired with HTML and as individual requests on route updates. In the latter case setting headers this way makes a lot of sense. However, in the former you can have multiple of these all competing for each other and we've adopted a JSX based way of setting headers during rendering so it isn't as simple I think as just setting headers here I think. At minimum this requires some review.

boehs:
> spent 6 hours writing yesterday before losing it.

Owch.

oscartbeaumont:
Thanks for giving this a look @ryansolid. I can understand your concern.

After using this PR I noticed I never considered setting response headers in a normal `server$` function. If we stick with the same API I think we would wanna extend it to add something like `server$.responseHeaders`.

I hope we can come up with something to support this use case. One of the workarounds to this issue right now is to use an API route and a fetch request on the page but this requires JS which is annoying because one of my favorite parts of solid-start is how easy it is to make a UI that can run without JS and then progressively hydrate.

If you get a better idea of your plan around this and need someone to help building it just let me know.

For anyone who is blocked by this issue I have included a patch-package below that I have been using with `pnpm patch`. Obviously, the API may change for the final version of this feature and expect this to break between updates of `solid-start` but it might help if you were blocked on this issue like I was.

<details>
  <summary>Patch</summary>

`package.json`:
```json
{
  ...
  "pnpm": {
    "patchedDependencies": {
      "solid-start@0.2.5": "./patch/solid-start@0.2.5.patch"
    }
}
```

`patch/solid-start@0.2.5.patch`
```diff
diff --git a/data/createRouteData.tsx b/data/createRouteData.tsx
index 27de7ee5bb5f269ae5b8bbde5a2be4aa3c323851..934d32c292627c47a45e4fd296f8efef1ff697e5 100644
--- a/data/createRouteData.tsx
+++ b/data/createRouteData.tsx
@@ -12,7 +12,7 @@ import { createStore, reconcile, unwrap } from "solid-js/store";
 import { isServer } from "solid-js/web";
 import { useNavigate } from "../router";
 import { isRedirectResponse, LocationHeader } from "../server/responses";
-import { ServerContext } from "../server/ServerContext";
+import { ServerContext, useRequest } from "../server/ServerContext";
 import { FETCH_EVENT, ServerFunctionEvent } from "../server/types";
 
 interface RouteDataEvent extends ServerFunctionEvent {}
@@ -71,11 +71,13 @@ export function createRouteData<T, S>(
     try {
       let event = pageEvent as RouteDataEvent;
       if (isServer) {
+        const req = useRequest();
         event = Object.freeze({
           request: pageEvent.request,
           env: pageEvent.env,
           $type: FETCH_EVENT,
-          fetch: pageEvent.fetch
+          fetch: pageEvent.fetch,
+          responseHeaders: req.responseHeaders
         });
       }
 
diff --git a/server/middleware.ts b/server/middleware.ts
index 28f60691ba7d9572bb32369697cf814f1e05610c..366a8eb4413ffa5233aca7fcebb06a45d9b491cb 100644
--- a/server/middleware.ts
+++ b/server/middleware.ts
@@ -33,11 +33,13 @@ export const inlineServerFunctions: ServerMiddleware = ({ forward }) => {
         });
       }
 
+      let responseHeaders =  new Headers();
       let serverFunctionEvent = Object.freeze({
         request: event.request,
         fetch: internalFetch,
         $type: FETCH_EVENT,
-        env: event.env
+        env: event.env,
+        responseHeaders
       });
 
       const serverResponse = await handleServerRequest(serverFunctionEvent);
@@ -55,22 +57,31 @@ export const inlineServerFunctions: ServerMiddleware = ({ forward }) => {
       ) {
         const formData = await formRequestBody.formData();
         let entries = [...formData.entries()];
-        return new Response(null, {
-          status: 302,
-          headers: {
-            Location:
-              new URL(event.request.headers.get("referer")).pathname +
-              "?form=" +
-              encodeURIComponent(
-                JSON.stringify({
-                  url: url.pathname,
-                  entries: entries,
-                  ...(await serverResponse.json())
-                })
-              )
-          }
-        });
+
+        responseHeaders.append("Location", new URL(event.request.headers.get("referer")).pathname +
+          "?form=" +
+          encodeURIComponent(
+            JSON.stringify({
+              url: url.pathname,
+              entries: entries,
+              ...(await serverResponse.json())
+            })
+        ));
+
+         return new Response(null, {
+           status: 302,
+           headers: responseHeaders
+         });
+      } else if (!serverResponse) {
+          return new Response(null, {
+            headers: responseHeaders
+          })
       }
+
+      for (const [name, value] of responseHeaders) {
+        serverResponse.headers.append(name, value);
+      }
+
       return serverResponse;
     }
 
diff --git a/server/types.tsx b/server/types.tsx
index 8437e66a9674446692a8f52fd496cb23bc078801..fcea9c291fd0d084d1e750a8372a4764ec5310e5 100644
--- a/server/types.tsx
+++ b/server/types.tsx
@@ -64,6 +64,7 @@ export interface FetchEvent {
 export interface ServerFunctionEvent extends FetchEvent {
   fetch(url: string, init: RequestInit): Promise<Response>;
   $type: typeof FETCH_EVENT;
+  responseHeaders: Headers;
 }
 
 export interface PageEvent extends FetchEvent {
```
</detail>

pilcrowOnPaper:
@ryansolid 
> Both initial render when paired with HTML and as individual requests on route updates. In the latter case setting headers this way makes a lot of sense. However, in the former you can have multiple of these all competing for each other and we've adopted a JSX based way of setting headers during rendering so it isn't as simple I think as just setting headers here I think.

My understanding is that you *can* already set headers on initial render, using `useRequest`. Or, is the concern about having multiple ways of setting headers within a single `serverRouteData`?

marcolanaro:
Hey @ryansolid, first of all I want to say thank you for all you are doing.
I see your point, it's not trivial.
I like @oscartbeaumont approach: hook base api are great to interact with the lifecycle of the application. A hook valid only on route change scenario could make sense if named explicitly.
On the other hand I see how [json response](https://github.com/solidjs/solid-start/blob/main/packages/start/server/responses.ts#L13) is 99% there. In that case #233 would be related.

I'm happy to help so let e know when you'll came up with a clear intent.

ryansolid:
Worth adding Remix's solution to the conversation: https://remix.run/docs/en/v1/route/headers

They appear to ignore loader/action headers by default when rendering the page but they expose them so you can merge them into your page headers and then for page headers most nested level wins. Most nested wins goes back to the idea that our header tag should replace rather than append.

But we don't really have a means right now of exposing headers from any of these and awkwardly in this case we don't block. So route data might resolve after page headers have been called. Not to mention we can have multiple per route section. Without guarantee of order what replaces what is trickier. Remix's approach of catching all the headers seems more manageable. However even then they recognize how messy this is just reading that.

nksaraf:
When I had thought about this use-case: we believed that headers for the page and for the client-side fetch should be explicit. so that you can control what headers get applied to the page. 

So you're right: the best way to set the page headers are using `useRequest().responseHeaders` or the `HttpHeader` component. And to set headers for the client-side fetch, you should respond with a `Response` object with headers. Use our `json` helper for this. This needs a bit of work to make it proper, where we can understand that `json` with headers is still a JS object (so the consumer doesn't need to parse to use).

So:

```ts
export function routeData() {
  return createServerData$((_, { responseHeaders }) => {
    responseHeaders.append("x-demo", "123");

    return json("mydata", { headers: { "x-demo": 123 } });
  });
}
```

Might seem a bit repetitive but becomes its explicit, its more composable.

Spiralis:
> And to set headers for the client-side fetch, you should respond with a `Response` object with headers. Use our `json` helper for this.

@nksaraf What you are saying here is referring to the future solution? Because at the moment the json returned Response is not working well. I mean, My use-case I want to both return data and apply the headers. And, when returning a Response the data is somewhat harder to extract. One has to reverse back to awaits or .then()'s - and the use of .json(). Plus, the headers are not applied.

I think maybe I too _might_ have scenarios that could benefit on separating when to apply them, based on whether it is a page header (SSR) or a fetch.

Is this change happening? Or are there other changes that breaks this?

pilcrowOnPaper:
@ryansolid Can I please get an update on this? Even just having a way to set cookies in `routeData` would be great.

GiyoMoon:
Hey, I'm currently using a patch to add this feature to my project and I came across an issue. If `createServerData$` is used inside a component and called again after the initial load of the page (For example due to a change of the key), the response headers cannot be set. I'm not that familiar with solid-start's codebase but it looks like the issue causing this is in `createRouteData.tsx`, where the response headers are only set if it's `isServer`. And in that case, `isServer` is false and the responseHeaders cannot be set.

I will create a small reproduction repo later and add it to this comment.

<details>
  <summary>I'm using this patch for solid-start `0.3.2`</summary>
./patches/solid-start@0.3.2.patch

```patch
diff --git a/api/middleware.ts b/api/middleware.ts
index 3829de2352cf9c65146a6560461e89cd62083b85..bb1ad7d5fef2fbf71150453537276d1f77f4bcc9 100644
--- a/api/middleware.ts
+++ b/api/middleware.ts
@@ -6,6 +6,7 @@ export const apiRoutes: Middleware = ({ forward }) => {
   return async (event: FetchEvent) => {
     let apiHandler = getApiHandler(new URL(event.request.url), event.request.method);
     if (apiHandler) {
+      let responseHeaders = new Headers();
       let apiEvent = Object.freeze({
         request: event.request,
         clientAddress: event.clientAddress,
@@ -13,10 +14,15 @@ export const apiRoutes: Middleware = ({ forward }) => {
         params: apiHandler.params,
         env: event.env,
         $type: FETCH_EVENT,
-        fetch: event.fetch
+        fetch: event.fetch,
+        responseHeaders
       });
       try {
-        return await apiHandler.handler(apiEvent);
+        const resp = await apiHandler.handler(apiEvent);
+        responseHeaders.forEach((value, name) => {
+          resp.headers.append(name, value);
+        });
+        return resp;
       } catch (error: any) {
         if (error instanceof Response) {
           return error;
diff --git a/data/createRouteData.tsx b/data/createRouteData.tsx
index 076c2b061144f0b270da970283c9bb28ce63a12d..85da785affe85f7522369fd8b547697825b5c8ea 100644
--- a/data/createRouteData.tsx
+++ b/data/createRouteData.tsx
@@ -77,7 +77,8 @@ export function createRouteData<T, S = true>(
           clientAddress: pageEvent.clientAddress,
           locals: pageEvent.locals,
           $type: FETCH_EVENT,
-          fetch: pageEvent.fetch
+          fetch: pageEvent.fetch,
+          responseHeaders: pageEvent.responseHeaders
         });
       }

diff --git a/server/types.tsx b/server/types.tsx
index a2b4b66ddb573a6a594456a8a24611b670a1213a..8106da347d8b0ccb7c33adbc7cd6cbb39623b8bb 100644
--- a/server/types.tsx
+++ b/server/types.tsx
@@ -82,6 +82,7 @@ export interface FetchEvent {

 export interface ServerFunctionEvent extends FetchEvent {
   $type: typeof FETCH_EVENT;
+  responseHeaders: Headers;
 }

 export interface PageEvent extends FetchEvent {
```
package.json

```json
"pnpm": {
  "patchedDependencies": {
    "solid-start@0.3.2": "patches/solid-start@0.3.2.patch"
  }
}
```
</details>

edivados:
This seems to work, it's just inconvinient to get the data out of the response inside the component. But I may not understand the downsides to this...

```jsx
export function routeData() {
  return createServerData$((_, { fetch }) => {
    const request = useRequest();
    const randomNumber = Math.floor(Math.random() * 10);
    const cookieString = `test=${randomNumber}; Path=/`;
    if (request.responseHeaders) {
      // Sets headers during page render (ssr) 
      request.responseHeaders.set('x-random', randomNumber.toString());
      request.responseHeaders.append('set-cookie', cookieString);
    }
   // Sets headers during request (refetchRouteData)
    return json(`Content ${randomNumber}`, { 
      headers: new Headers({ 
        'set-cookie': cookieString, 
        'x-random': randomNumber.toString()
      })
    });
  })
}

  const response = useRouteData<typeof routeData>();
  const data = createResource<string, Response>(() => response.latest, async (res) => res.json());
....
```

@GiyoMoon I will try your patch later. I am assuming responseHeader on the event object will be undefined during ssr and this only covers request (refetchRouteData for example)?

GiyoMoon:
It looks like part of the `resourceFetcher` function in `createRouteData.tsx` is executed on the client if `createServerData$` is refetched after the site already went through ssr and is loaded on the client. And in that case, `event.responseHeaders` is undefined/it throws an error if I try to access it.

Part of `createRouteData` I'm talking about:
```ts
let event = pageEvent as RouteDataEvent;
if (isServer && pageEvent) {
  event = Object.freeze({
    request: pageEvent.request,
    env: pageEvent.env,
    clientAddress: pageEvent.clientAddress,
    locals: pageEvent.locals,
    $type: FETCH_EVENT,
    fetch: pageEvent.fetch,
    responseHeaders: pageEvent.responseHeaders
  });
}
let response = await (fetcher as any).call(event, key, event);
```

GiyoMoon:
Alright I created an example:
```
git clone https://github.com/GiyoMoon/solid-start-expose-headers.git
cd solid-start-expose-headers
pnpm install
pnpm dev
```
I added a console.log to the patch for debug purposes.
Navigate to https://localhost:3000. You should see `Inside createRouteData, on the server: true` in the server console. But when pressing the button, it recalls `createServerData$` and it prints `Inside createRouteData, on the server: false` in the **browser console**. Is this intended behaviour? Regardless, it cases the `event.responseHeaders` to be undefined which breaks the feature of this pull request :/

edivados:
@GiyoMoon This should cover server requests, if you want to try: https://github.com/edivados/solid-start-expose-headers/tree/edivados-patch-1

GiyoMoon:
Great, that seems to work! Thanks for taking the time to check it out :)
Just for my understanding, aren't we losing some headers here?:
https://github.com/edivados/solid-start-expose-headers/blob/edivados-patch-1/patches/solid-start%400.3.2.patch#L82

Before the patch, `data.headers` was used to initialize the header object and now it is not used anymore.

edivados:
> Great, that seems to work! Thanks for taking the time to check it out :) Just for my understanding, aren't we losing some headers here?: https://github.com/edivados/solid-start-expose-headers/blob/edivados-patch-1/patches/solid-start%400.3.2.patch#L82
> 
> Before the patch, `data.headers` was used to initialize the header object and now it is not used anymore.

uhhh oops 😅

GiyoMoon:
I think this should do it:
https://github.com/GiyoMoon/solid-start-expose-headers/blob/main/patches/solid-start%400.3.2.patch#L75

ryansolid:

Issue #495, opened 2022-12-01
Unstable Islands only work with double quote imports


ryansolid:
Single quote imports are missed by the compilation. While this isn't the final API. Definitely unexpected trap.

orenelbaum:
https://github.com/solidjs/solid-start/pull/499

orenelbaum:
this issue can be closed

PR #499, opened 2022-12-02
fix: islands only work with double quote imports

orenelbaum:
https://github.com/solidjs/solid-start/issues/495
needed to update the regex

### DIFF
--- a/packages/start/vite/plugin.js
+++ b/packages/start/vite/plugin.js
@@ -691,16 +691,18 @@ function islands() {
     transform(code, id, ssr) {
       if (code.includes("unstable_island")) {
         let replaced = code.replaceAll(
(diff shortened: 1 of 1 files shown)


packages/start/vite/plugin.js
```diff
@@ -691,16 +691,18 @@ function islands() {
     transform(code, id, ssr) {
       if (code.includes("unstable_island")) {
         let replaced = code.replaceAll(
-          /const ([A-Za-z_]+) = unstable_island\(\(\) => import\("([^"]+)"\)\)/g,
-          (a, b, c) =>
-            ssr
+          /const ([A-Za-z_]+) = unstable_island\(\(\) => import\((("([^"]+)")|('([^']+)'))\)\)/g,
```
atk, line 694: `(("([^"]+)")|('([^']+)'))` could also be expressed as `((["'`])(.+?)\1)` at the expense of an additional argument for the replacer.

PR #500, opened 2022-12-02
fix unstable islands on windows

orenelbaum:
I replaced the backslashes in the windows path with forward slashes


ryansolid:
I have to ask.. is doing path conversions like this our best option?

atk:
There's no native path conversion in node, unfortunately.

orenelbaum:
I don't know if that's how it's supposed to be done, but it seems to work on both Windows and Linux

dev-rb:
I implemented a similar fix in the movies branch. Using the vite `normalizePath` function.
https://github.com/solidjs/solid-start/blob/movies/packages/start/islands/vite-plugin.js

peerreynders:
> I don't know if that's how it's supposed to be done

Going by the node documentation I'd assume something like this:

```JS
// file: index.mjs
import { parse, posix, sep} from 'node:path';

const toPosix = (original) => 
  original.split(sep).join(posix.sep);

const full = process.argv[1];
const { root, dir } = parse(full);
const noRoot = dir.slice(root.length);

console.log(toPosix(full));
console.log(toPosix(dir));
console.log(toPosix(noRoot));
```

- [`path.sep`](https://nodejs.org/api/path.html#pathsep)
- [`path.posix`](https://nodejs.org/api/path.html#pathposix)

PR #510, opened 2022-12-04
Engine requirement to meet nodejs/undici is >=16.8

yurisich:
See https://github.com/nodejs/undici/blob/68c269c4144c446f3f1220951338daef4a6b5ec4/index.js#L95 Using anything less than 16.8 will not start with this error:

```
class NodeRequest extends BaseNodeRequest {
                          ^            
                        
TypeError: Class extends value undefined is not a constructor or null
````


ryansolid:
Thank you

PR #520, opened 2022-12-05
Undefined when converted to string returns "undefined"

stereobooster:
Also while I'm here. Do you think we should pass original NODE_OPTIONS, like

```ts
NODE_OPTIONS: `--experimental-vm-modules ${inspect ? "--inspect" : ''} ${process.env.NODE_OPTIONS || ''}`
```
?

### DIFF
--- a/packages/start/bin.cjs
+++ b/packages/start/bin.cjs
@@ -131,7 +131,7 @@ prog
         stdio: "inherit",
         env: {
           ...process.env,
-          NODE_OPTIONS: `--experimental-vm-modules ${inspect ? "--inspect" : undefined}`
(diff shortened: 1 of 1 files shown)


patdx:
> Also while I'm here. Do you think we should pass original NODE_OPTIONS, like
> 
> ```ts
> NODE_OPTIONS: `--experimental-vm-modules ${inspect ? "--inspect" : ''} ${process.env.NODE_OPTIONS || ''}`
> ```
> 
> ?

Actually that is what PR #399 is trying to do at the moment too:

https://github.com/solidjs/solid-start/pull/399/files

PR #540, opened 2022-12-17
Add `@latest` to `npm init solid`

sapphi-red:
`npm init` uses a local cached package if existed.
So an outdated version could be called with `npm init solid`.

This PR changes all `npm init solid` to `npm init solid@latest` to prevent this. (pnpm always fetches the latest version and doesn't need `@latest`.)


ryansolid:
Thanks

PR #544, opened 2022-12-17
Fix `pnpm run docs:build`

ryanseipp:
Fixes the docs:build script used to deploy the docs site. The error occurring was twoslash expects type definitions for modules imported in code examples. `docs/core-concepts/api-routes.md` includes an example of exposing a GraphQL API, and imports from the `graphql` package.


ryanseipp:
Open to opinions on where the best place to add this dependency is.

ryansolid:
Thanks

Issue #562, opened 2022-12-21
Navigating with query strings does not trigger re-fetch


knpwrs:
In all three of these cases:

1. Using `A` from `solid-start`:

```tsx
<A href="?after=foo">Next page</A>
```

2. Calling `navigate` from `useNavigate`:

```tsx
const navigate = useNavigate();

// snip

<a
  href="?after=foo"
  onClick={(e) => {
    e.preventDefault();
    navigate('?after=foo');
  }
>
  Next Page
</a>
```

3. Calling `setSearchParams` from `useSearchParams`:

```tsx
const [,setSearchParams] = useSearchParams();

// snip

<a
  href="?after=foo"
  onClick={(e) => {
    e.preventDefault();
    setSearchParams({ after: 'foo' });
  }
>
  Next Page
</a>
```

No data is refetched. I have a more complete example for the first snippet utilizing `A`:

```tsx
export function routeData({ params, location }: RouteDataArgs<{ id: string }>) {
  const after = location.query['after'];
  console.log('routeData', { after });

  return createServerData$(
    async ([, id, after = null], { request }) => {
      console.log('serverData'. { after });
      // fetch and return result
    },
    { key: () => ['channels', params['id'], after] },
  );
}

export default function ChannelRoute() {
  const data = useRouteData<typeof routeData>();

  return (
    <>
      { /* snip */ }
      <A href={`?after=foo`}>Next Page</A>
    </>
  );
}
```

When I load the page with no query string, I get the following log statements on the server:

```
routeData { after: undefined }
serverData { after: null }
```

Then when I click the link containing ?after=foo, I get no other logs on server or client.

What I'm trying to accomplish is pagination utilizing query strings rather than a route parameter. Is there a way to accomplish this? I've also tried using `refetchRouteData` but it appears to not support this use case -- I think it wants to invalidate and refetch a specific key, but it doesn't seem to want to invalidate the previous key and fetch a new key (where the last part of the key is a new query param).

atk:
Why don't you use URL parts instead of query strings?

Also, while I'm here, might I interest you in our [community primitive for numbered pagination](https://github.com/solidjs-community/solid-primitives/blob/main/packages/pagination/README.md)?

knpwrs:
I could use URL segments, though query strings seem like the canonical pattern across the web. I just checked the first three websites I thought of off the top of my head:

- Reddit: https://www.reddit.com/?count=25&after=t3_zrfuq2
[...]

PR #567, opened 2022-12-22
Add example of refetching based on a query parameter

knpwrs:
I didn't realize until #562 that `key`s in `createRouteData` and `createServerData$` were reactive. This PR documents that.


ryansolid:
Thank you

Issue #480, opened 2022-11-26
server env variables require additonal configuration / packages


OrJDev:
Currently enviroment variables that are not prefixed with `VITE_` will not be loaded,
variables that are prefixed with `VITE_` will be exposed to the client (`import.meta.env`),
what causing it so there is no such thing as `server env variable` in SolidStart.

this could easily be fixed using dotenv:
```ts
import solid from "solid-start/vite";
import dotenv from "dotenv";
import UnoCSS from "unocss/vite";
import { defineConfig } from "vite";
// @ts-expect-error no typing
import vercel from "solid-start-vercel";
  
export default defineConfig(() => {
  dotenv.config();
  return {
    plugins: [
          solid({ ssr: false, adapter: vercel({ edge: false }) }),
          UnoCSS(),
        ],
  };
});
```
but I think that it should not be handled in our end as SolidStart is a meta framework, and should be loading server variables by default.

ryansolid:
I agree. I imagine you have an idea of what this should look like. Is it just adding dotenv? Open to following your lead here as I've never really looked at this part in other meta frameworks. I've heard that SvelteKit does well here but that's about it.

OrJDev:
Ye you just have to add dotenv, and call the config function. As you can see in the example above it is that simple.

OrJDev:
Check out how nuxt does it

https://github.com/nuxt/nuxt.js/blob/dev/packages/config/src/load.js

OrJDev:
just make sure to merge and test: examples/bare is where i tested it

https://github.com/solidjs/solid-start/pull/585

OrJDev:
Reopened, close when my pr gets merged please

apatrida:
Are these build-time or runtime variables?  i.e. if you do a production build and have not yet defined `VITE_SOMETHING` then will `VITE_SOMETHING` be available on the client if it is only defined before running the build?

SveltKit supports both static (build time) and dynamic (runtime) variables and docker containers, k8s will want to use runtime and not build time since secrets maybe are not baked into the container and are provided later.

JakkuSakura:
A note here:
app.config.ts requires dotenv yourself

PR #585, opened 2022-12-24
support server env variables

OrJDev:
.env:
```
VITE_TEST=1
SERVER_TEST=2
```
code on client:
 ```ts
  console.log("client", import.meta.env.VITE_TEST); // equals 1 - as expected
```
  
credits to https://github.com/nuxt/nuxt.js/blob/dev/packages/config/src/load.js 

code on server:
```ts
export const routeData = () => {
  return createServerData$(() => {
    // @ts-ignore
    const t = process.env.SERVER_TEST;
    console.log({ t }); // equals 2 - as expected
    return t;
  });
```
};  

closes https://github.com/solidjs/solid-start/issues/480


OrJDev:
on client:

    ```const t = process.env.SERVER_TEST; // not defined - as expected```

OrJDev:
on server:

```ts
    const s = process.env.VITE_TEST; // not defined - as expected because it was loaded on client
```

ryansolid:
Thank you. Although not going to default bare to `ssr:false`.

OrJDev:
Ahh yes i just hated that it prints it multiple times, ofc you can delete the entire env thing from the examples it was just to test

apatrida:
now configure environment variables, do a production build, and see VITE ignore them at runtime as they appear that `VITE_` variables are only read and used at build time.  Above examples seem like you are only running and testing in dev mode.

PR #587, opened 2022-12-24
fix typescript

xania:
fix typescript complaining about headers having wrong type type at fetch

Type '{ "User-Agent": string; } | { "User-Agent"?: undefined; }' is not assignable to type 'HeadersInit | undefined'.
  Type '{ "User-Agent"?: undefined; }' is not assignable to type 'HeadersInit | undefined'.
    Type '{ "User-Agent"?: undefined; }' is not assignable to type 'Record<string, string>'.
      Property '"User-Agent"' is incompatible with index signature.
        Type 'undefined' is not assignable to type 'string'.


ryansolid:
Thanks

PR #594, opened 2022-12-26
fix: make island router replace outermost diffed route

XiNiHa:
This PR fixes a bug that the islands router always replaces the innermost route of the whole nested routes, which is supposed to replace the outermost diffed route.


ryansolid:

Issue #519, opened 2022-12-05
`ssr: false` has no effect?


stereobooster:
I'm new to this lib, so maybe I don't understand something. I have

```ts
export default defineConfig({
  plugins: [solid({ ssr: false })],
});
```

But components still render on the server. I get this error

```
An error occured while server rendering /:

	window is not defined
```

viridia:
I encountered this as well. What I eventually ended up doing is:

```
if (typeof window === 'undefined') {
```

...which is a way to test for the existence of `window` without getting a dereference error.

orenelbaum:
Setting SSR to false does mean that CSR would be used as far as I know. So it probably still runs the SSR but doesn't use it.

stereobooster:
> which is a way to test for the existence of window without getting a dereference error.

Yes I know how `typeof` works. I could as well do

```ts
import { isServer } from "solid-js/web";

if (!isServer) ...
```

I'm not looking for workaround. I just think it's counterintuitive

orenelbaum:
It's definitely something that needs to be solved even if there's an easy workaround.

orenelbaum:
Interesting observation - only module scope code is run on the server. I think that it has to do with a Babel plugin Solid Start runs.

adriancu:
I'm getting the "window is not defined" from within the pikaday.js calendar library, which I'm using in a Solid Component. (SolidStart, with a bare template and SSR no, and tailwind) 

The Component,  DateForm is as follows...

```jsx
import { createSignal, onMount } from "solid-js"
import Pikaday from 'pikaday'

export function DateForm() {

  const [formDt, setFormDt] = createSignal("")

  onMount(() => {
    const pika = new Pikaday({
      field: document.getElementById('formpika'),
      onSelect: function(dt) {
        setFormDt(pika.toString())
      }
    })
  })

  return (
    <>
      <div class="flex flex-col max-w-screen-md">
        <div class="flex flex-col items-center justify-center">
          Hello, choose date:
        </div>
        <div>
          <div>Date: { formDt }</div>
          <input type="text" id="formpika" />
        </div>
      </div>
    </>
  )
}
```
The window usage is within pikaday. Is there pehaps a work around to get and set a window reference, before calling pikaday?

Vexcited:
Try to use a dynamic import inside the onMount like `const Pikaday = await import("pikaday");`.

adriancu:
[...]

PR #595, opened 2022-12-26
fix csr dev server evaluates user modules

orenelbaum:
- fixes https://github.com/solidjs/solid-start/issues/519
I don't think that this is going to break anything but I haven't run tests cause I'm not sure how to make playwright work.


ryansolid:
Hmm.. it did seem to break a bunch of the tests. I will have to review this more closely.

ryansolid:
Ok I've done some work here. The async imports break the tests. I managed to solve this at the Vite plugin level. But the real challenge here is I think not importing fileRoutes breaks API and server function routes. So I'm not sure that will fly even for dev. I don't have a good answer here at least for the server functions as someone could put something top level that is client only and if they don't wrap it in a `!isServer` then we are pulling it in. API routes that is much less likely to happen though. Not sure if @nksaraf has thoughts. My first temptation though is to at least merge this and my work heading in this direction even if it doesn't quite solve the full problem yet.

ryansolid:
This is actually merged but I botched it a bit. Just closing here. You will see the merge commit in the repo.

PR #597, opened 2022-12-27
fix: pass only pathname to getMatchedBranch

XiNiHa:
`getMatchedBranch()` doesn't work properly with location strings that have search params included. However, the island router is currently receiving a plain string location that includes search params.

This PR fixes it by making the location prop to be an URL and passing only the pathname when calling `getMatchedBranch()`.


ryansolid:

Issue #421, opened 2022-11-11
Make TypeScript the default while scaffolding a starter app


nexxeln:
When you use the CLI to initialize a project, the default option for 'Use TypeScript' is false. So when you just press enter, it will select JavaScript
<img width="852" alt="image" src="https://user-images.githubusercontent.com/95541290/201262797-3d10f869-c0dc-44da-876a-8e1f1e975e9b.png">

My proposal is to make TypeScript the default. Happy to make a PR.

PR #602, opened 2022-12-28
Make TypeScript the default when running create-solid (#421)

obask:
tested manually by running a bin file produced by `pnpm --filter create-solid build`


ryansolid:
Thank you

PR #605, opened 2022-12-29
Add ScrollIndicator for docs sidebar nav

leoj3n:
https://user-images.githubusercontent.com/990216/209974748-b3d5714a-6d13-44f6-bbae-80ed0b2dc025.mov

JS size before PR (71.2 kB):

image

JS size after PR (76.6 kB, +5.4 kB):

image


leoj3n:
Hey! 👋  Was hoping to have some conversation about this PR, to get some expert opinions... I'll do my best to explain.

First of all, please note that my PR also includes some important changes to classes within the Nav component that improve the contrast/accessibility of hovered/active links in the nav sidebar that would definitely be worth extracting into a separate PR if this is not going to be refused to be merged.

Second, and more importantly to my continued understanding, I want to get some opinions about how I implemented this ScrollIndicator component within the isldand-enabled docs site for solid-start...

Initially, I thought to try to just pass around refs starting from the scope of the Nav component, but quickly found out that this would not work with the island/router setup, because the Nav component is not an island... I thought maybe the Nav component could become an island too and that may allow refs to be passed starting from that scope but I was worried to try that because of these Qs:

1) Could the Nav component actually be turned into an unstable island?
2) Would this affect performance/optimization negatively considering all the work Nav does?

It just seemed like something I didn't really want to try at the time, as I was just getting familiar with the islands concept. Should I have tried taking that path...?

My goal was to make few modifications to the existing code while also making the new ScrollIndicator component as generic as possible... I tried a number of approaches before settling on using Context... For instance, I had one rendition that observed mutations and received a string prop indicating a query selector to use to find the active element... Although I assume that MO are inefficient and it felt maybe too disconnected from the whole solidjs reactive system... I also looked into using events so the children links could emit when they were active... However I felt it was probably best to work with refs where possible, and then get the state synced between active links and ScrollIndicator using context; this is the route I ended up taking for the PR.

The only obviously clumsy part of this approach is that I had to do `containerEl.parentElement.parentElement`... This is to get past the wrapping `<solid-island>` element which `solid-start` seems to need:

```html
<solid-island data-props="{&quot;class&quot;:&quot;pt-8 pb-20 md:pb-8 px-8&quot;}" data-component="/components/ScrollIndicator.tsx?island" data-island="/components/ScrollIndicator.tsx?island" data-when="load">
```

You might ask why not wrap the Nav component instead of making ScrollIndicator the immediate child of `<nav>` within the Nav component? It's because of this wrapping `<solid-island>` element; we would have to change the primary layout CSS of the solid-start docs site in multiple places outside of the Nav component, or try and inject the tailwind classes onto this wrapping element; but anything I could think to try along these lines just felt wrong due to Nav being a primary part of the layout. Note that if Nav were turned into an island, this wrapping problem would still exist, as the Nav component and it's first child are primary layout elements (part of some grid layout; notice `row` and `col` on ` <nav class="min-w-[300px] col-start-1 row-start-2 ...`).

To reiterate, I needed to be able to get a reference to this `<nav>` first child element, but can't use ref because the Nav component is not in an island, and wrapping it or making it an island would force us to target the injected `<solid-island>` with CSS which feels wrong. I tried playing with `Portal` but this did not yield any usable solution either.

I guess this leads to a pretty interesting question for a core author...

3) Would it be possible to not wrap islands with this `<solid-island>` wrapper element?

That would be a great help, but I have a feeling it might not be tenable and that this wrapper has to exist for various reasons... Can someone please confirm this as not possible? This may be the most critical question at the moment.

4) Should I have used events instead of context?
5) Should/could I have made the Nav component itself an island so that can just pass refs around? How to solve the `<solid-island>` wrapper element issue (totally rework the css to avoid using grid?)?

Also note that I had to update the `.npmrc` to explicitly make pnpm update the lockfile so CI could run... I couldn't find any local settings on my machine that were setting the lockfile setting to false and pnpm docs say it should default to true... Not sure why it was an issue but explicitly setting the lockfile setting to true in the `.npmrc` fixes it and being explicit here shouldn't be an issue if you were to merge that change (perhaps it will help with other PRs as well, idk).

I'm going to tag @nksaraf on this as I see he has done most of the authoring on the solid-start docs site.

Thanks and sorry if this is inquiry is not as concise as it could be, but kind of hard to explain. I understand islands is still being figured out. Hope to get a reply! Let me know if I can clarify anything here; hope to get some input on how to more optimally work with islands.

ryansolid:
Yeah you are touching largely on why I haven't stopped to look at this. Mostly patterns around Islands are still being figured out and some details are sort of tricky.

1. It could be. This starts us down the path of potentially nesting unstables in unstables, and sort of points a big flaw in the current API design. It is the boundary that matters but how conscious of what is server should we be when authoring. We effectively have 2 versions of `<A>` one that reaches the client and one that doesn't. This sort of inconsistency is troublesome.

2. It could. Which is why the fact that for 1. we haven't made a clear delineation is problematic. There are patterns for hoisting that can keep things relatively isolated but we haven't really documented them or figured out how to codify them into best practices. 

3. This can be addressed. It is better to use comments to wrap ranges but more expensive to find them. Elements are much easier to deal with. We went with elements at this point because it was basically no effort. Just a query selector. To do comments we will need custom tree walkers. Completely doable, just haven't done it yet.

4. Context is fine in theory, but it has some holes right now. Mostly that future navigations that render on the server might not have access to it at server render time. We need to be smarter here, and actually look at doing more client rendering hybrids on routing after initial page load, but haven't implemented it yet. Right now where things are at events might be easier. But ultimately I'd like to be able to use context.

5. Yeah similar to 1 above. Building this site using Islands was a good learning exercise and gives us a vehicle to get Docs content out there and from the community. Structurally this will require overhaul as we refine the APIs.

ryansolid:

leoj3n:
Hi @ryansolid — Thanks for the very robust initial reply from the beginning of 2023.

Hate to ask you to reply to yourself over a year later, but can you give an update on these bullet points? You make some more overarching points that would be helpful for me to understand how things have progressed. I would like to know what is/is not still relevant, as it looks like the files I was PRing on are still pretty similar, still using unstable island, etc. but I also see a new branch that doesn't https://github.com/solidjs/solid-start/blob/vinxi-0.2.0/docs/app.tsx#L118

For instance, based on things you brought up in the original reply: Are unstable islands still a relevant thing? Are they still experimental but being worked towards, or have they been made a stable feature? Do you walk for comments instead of using a div now? What about the "holes" in the context api? Etc...

NOTE:
I do see some hints here https://github.com/solidjs/solid-router?tab=readme-ov-file#migrations-from-09x but I don't entirely understand. I see no more [`<IslandA>`](https://github.com/leoj3n/solid-start/blob/76b92d48b264547a4287fd213841ac27759f169f/docs.root.tsx#L201), but instead [`<A>` imported from Router](https://github.com/solidjs/solid-start/blob/fca4cec50e9cfcc008848416fd990c82e2079318/docs/app.tsx#L194) ...Is this `<A>` working similarly to`<IslandA>`? I see there is still [`clientOnly`](https://github.com/solidjs/solid-start/blob/main/packages/start/src/shared/clientOnly.tsx#L7) in `solid-start` but not used in `solid-router`.

_1. Could the Nav component actually be turned into an unstable island?_
> 1. It could be. This starts us down the path of potentially nesting unstables in unstables, and sort of points a big flaw in the current API design. It is the boundary that matters but how conscious of what is server should we be when authoring. (( **We effectively have 2 versions of `<A>` one that reaches the client and one that doesn't.** )) This sort of inconsistency is troublesome.

_2. Would this affect performance/optimization negatively considering all the work Nav does?_
> 2. It could. Which is why the fact that for 1. we haven't made a clear delineation is problematic. (( **There are patterns for hoisting that can keep things relatively isolated but we haven't really documented them or figured out how to codify them into best practices.** ))

_3. Would it be possible to not wrap islands with this `<solid-island>` wrapper element?_
> 3. This can be addressed. It is better to use comments to wrap ranges but more expensive to find them. Elements are much easier to deal with. We went with elements at this point because it was basically no effort. Just a query selector. (( **To do comments we will need custom tree walkers. Completely doable, just haven't done it yet.** ))

_4. Should I have used events instead of context?_
> 4. Context is fine in theory, but it has some holes right now. Mostly that future navigations that render on the server might not have access to it at server render time. (( **We need to be smarter here, and actually look at doing more client rendering hybrids on routing after initial page load, but haven't implemented it yet.** )) Right now where things are at events might be easier. But ultimately I'd like to be able to use context.

_5. Should/could I have made the Nav component itself an island so that can just pass refs around? How to solve the `<solid-island>` wrapper element issue (totally rework the css to avoid using grid?)?_
> 5. Yeah similar to 1 above. Building this site using Islands was a good learning exercise and gives us a vehicle to get Docs content out there and from the community. Structurally this will require overhaul as we refine the APIs.

As an aside, since making the PR, a CSS feature I recently learned about `display: contents` on `<solid-island>` would have perhaps helped, since it looks like the use of an element for unstable island was causing me issues, but I will have to look closer to be sure if that's helpful or not.

PR #608, opened 2022-12-30
Fix and improve with-vitest template

jchatard:
Add some commands to `package.json`:

- pnpm test: "vitest run",
- pnpm test-watch: "vitest",
- pnpm test-ui: "vitest --ui",
- pnpm coverage: "vitest run --coverage"

I fixed a bug in Solid's vite plugin, which prevented Vitest --ui from starting.
`vite.resolvedUrls` is always null when running Vitest --ui server. So I added a check to that.

I updated the template so that is works out of the box.
And move test config, to `vitest.config.ts`instead of of `vite.config.ts`. This avoids configuration conflicts as some have reported here and there.

Hope this helps :-)

### DIFF
--- a/examples/with-vitest/jest.config.js
+++ b/examples/with-vitest/jest.config.js
@@ -1,0 +1,3 @@
+module.exports = {
+  preset: "solid-jest/preset/browser",
+};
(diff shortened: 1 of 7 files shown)


ryansolid:
Thanks

Issue #612, opened 2023-01-01
With Vercel prerender ISR gets applied to API routes


ryoid:
If you have ISR enabled with the `solid-start-vercel` adapter the API routes get prerendered likes pages as well, which I don't think is expected behaviour.

```ts
solid({
  adapter: vercel({
    prerender: {
      expiration: 1000
    },
  }),
})
```

This is due to everything being bundled into a single `render.func` and the `render.prerender-config.json` is applied to it.

Build output:
```
.vercel
├── output
│  ├── config.json
│  ├── functions
│  │  ├── render.func
│  │  │  └── index.js
│  │  └── render.prerender-config.json
```

PR #613, opened 2023-01-01
split vercel api routes function

ryoid:
This creates a seperate serverless function `api.func` to handle API routes. 

An effect of this is fixing https://github.com/solidjs/solid-start/issues/612

Build output
```
.vercel
├── output
│  ├── config.json
│  ├── functions
│  │  ├── api.func
│  │  │  └── index.js
│  │  └── render.func
│  │     └── index.js
```

Config
```ts
{
  "version": 3,
  "routes": [
    ...
    // Add route for api handler
    {
      "src": "/api/hello",
      "dest": "/api"
    },
    {
      "src": "/.*",
      "dest": "/render"
    }
  ]
}
```

Not sure if this is the best approach, but would at least make ISR work with API routes. We could limit it to only be done when prerender is enabled, but I think it should be fine even when it is not enabled.

If there is a way to get a server output with only API routes it would make more sense (and without API routes for `render.func`)

### DIFF
--- a/packages/start-vercel/index.js
+++ b/packages/start-vercel/index.js
@@ -84,6 +84,20 @@ export default function ({ edge, prerender } = {}) {
 
       // Generate API function
       const apiRoutes = config.solidOptions.router.getFlattenedApiRoutes()
+      const apiRoutesConfig = apiRoutes.map(route => {
+        return { 
+          src: route.path.split('/')
+            .map(path => 
+              path[0] === ':'
+                ? `(?<${path.slice(1)}>[^/]+)`
+              : path[0] === '*'
+                ? `(?<${path.slice(1)}>.*)` 
+              : path
+            )
+            .join('/'),
+          dest: "/api" 
+        }
+      })
       if (apiRoutes.length > 0) {
         let baseEntrypoint = "entry.js";
         if (edge) baseEntrypoint = "entry-edge.js";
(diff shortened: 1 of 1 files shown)


packages/start-vercel/index.js
```diff
@@ -82,6 +82,57 @@ export default function ({ edge, prerender } = {}) {
...
+        copyFileSync(join(__dirname, baseEntrypoint), entrypoint);
+
+        const bundle = await rollup({
+          // Same as render
+          input: entrypoint,
+          plugins: [
+            json(),
+            nodeResolve({
+              preferBuiltins: true,
+              exportConditions: edge ? ["worker", "solid"] : ["node", "solid"]
```
ryoid, line 99: Maybe there should be a separate option in the config for using edge API routes.

A better solution would definitely be having it on a [per function basis](https://nextjs.org/docs/api-routes/edge-api-routes#basic), but that would require regenerating a new file for each function.

ryoid:
I have an experimental version available that splits each api routes into individual serverless functions https://github.com/ryoid/solid-start/pull/1 not sure if this breaks stuff though.

ryansolid:
Yeah this is interesting, because it touches on something I've been thinking about. How to effectively break apart these monoliths for optimal deployment and how to expose it. My temptation is to just split in half like you've done here for now. But this definitely an important area of consideration.

ryoid:
I think having it more granular helps when there are different runtimes, such as edge that might not work/make sense for all functions. Or when everything bundled together is larger than the platform's serverless size limits (unsure how common this is). Could also have it as an opt in similar to what [SvelteKit](https://github.com/sveltejs/kit/tree/master/packages/adapter-vercel#usage) has. Splitting it in half seems like a good first step.

Agree that if it does become named exports there needs to be a standardised pattern.

PR #621, opened 2023-01-03
server env variables - support vite mode variables

OrJDev:
this will allow adding server env variables that are based on vite mode:

- .env - which is the only thing that is supported right now
- .env.local
- .env.production / .env.development


OrJDev:
Related [discord support thread](https://discord.com/channels/722131463138705510/1059754857411649577)

PR #626, opened 2023-01-05
Use safest defaults for @rollup/plugin-commonjs

### LINKED ISSUE #623: Broken build when using mongoose - rollup plugin configuration
I've made a reproduction here;

https://github.com/birkskyum/solid-mongoose-issue

Importing mongoose in a route like this breaks the build:

```ts
import { Title } from "solid-start";
import Counter from "~/components/Counter";
import {version} from "mongoose"

  console.log(version)
  return (
    <main>
      <Title>Hello World</Title>
      <h1>Hello world!</h1>
      <Counter />
    </main>
  );
}
```

The error is like this:

```
./dist/server.js:107060 var decimal128$3 = driver$1.get().Decimal128; ^

TypeError: Cannot read properties of null (reading 'Decimal128') at ./dist/server.js:107060:34 at ModuleJob.run (node:internal/modules/esm/module_job:193:25) at async Promise.all (index 0) at async ESMLoader.import (node:internal/modules/esm/loader:530:24)

Node.js v18.12.1
```

It's supposedly related to the bundling order as mentioned in this mongoose [ticket](https://github.com/Automattic/mongoose/issues/12335#issuecomment-1299427402). It would be interesting to see if the error persist in vite to 4 /w rollup 3.

birkskyum:
Fixes #623 

Vite has an option to inject configuration parameters to @rollup/plugin-commonjs in the vite.config.ts like this:

```ts

export default defineConfig({
  build: {
    **commonjsOptions: {}**
  },
});
```

Unfortunately that doesn't work in solid-start, because the node adapter takes control over the 3 default rollup plugins like this ( separate issue #628 ):

https://github.com/solidjs/solid-start/blob/df206edb975de4941e368c92ec9e25c825592005/packages/start-node/index.js#L45-L52

Arguably it would be ideal to have a way of injecting the parameters from vite.config.ts in here, but as long as there is no such way, the way to adjust the parameters of commonjs plugin for users of SolidStart, it's wisest to use the safest default option that is least likely to cause errors, know as strictRequires (more info [here])(https://github.com/rollup/plugins/tree/master/packages/commonjs#strictrequires)): 

{ strictRequires: true }

It's [considered by a maintainer](https://github.com/rollup/rollup/issues/4787#issuecomment-1371842223) of the plugin, that this might be the better default config.

### DIFF
--- a/packages/start-aws/index.mjs
+++ b/packages/start-aws/index.mjs
@@ -37,7 +37,7 @@ export default function ({ edge } = {}) {
             preferBuiltins: true,
             exportConditions: ["node", "solid"]
           }),
-          common()
+          common({ strictRequires: true })
         ]
       });
       await bundle.write({
--- a/packages/start-cloudflare-pages/index.js
+++ b/packages/start-cloudflare-pages/index.js
@@ -148,7 +148,7 @@ export default function (miniflareOptions) {
             preferBuiltins: true,
             exportConditions: ["worker", "solid"]
           }),
-          common()
+          common({ strictRequires: true })
         ]
       });
 
--- a/packages/start-cloudflare-workers/index.js
+++ b/packages/start-cloudflare-workers/index.js
@@ -160,7 +160,7 @@ export default function (miniflareOptions = {}) {
             preferBuiltins: true,
             exportConditions: ["worker", "solid"]
           }),
-          common()
+          common({ strictRequires: true })
         ]
       });
       // or write the bundle to disk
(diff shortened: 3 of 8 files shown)


ryansolid:
Only consideration is if this affects code splitting. I think given this is a post process after the Vite build we should be fine. Like everything that would be shaken has been. We will give it a shot. But might need more consideration if this change breaks other things. Luckily less and less code outside of Node is impacted by this.

PR #627, opened 2023-01-05
Replace @vinxi/vite-plugin-inspect with vite-plugin-inspect

birkskyum:
Currently package/start depend on two vite-plugin-inspect packages (it's the same [github repo](https://github.com/antfu/vite-plugin-inspect)):

- @vinxi/vite-plugin-inspect
- vite-plugin-inspect (This is the newer npm package)

This PR simplifies things by:
- removing @vinxi/vite-plugin-inspect
- updating vite-plugin-inspect from 0.6.1 to 0.7.12 (Vite 4 compatible since ^[0.7.10](https://github.com/antfu/vite-plugin-inspect/releases/tag/v0.7.10))

### DIFF
--- a/packages/start/package.json
+++ b/packages/start/package.json
@@ -134,6 +134,6 @@
     "@solidjs/meta": "^0.28.0",
     "@solidjs/router": "^0.6.0",
     "solid-js": "^1.6.2",
-    "vite": "^3.0.4"
+    "vite": "^3.1.8"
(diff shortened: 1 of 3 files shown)


ryansolid:
Thanks. We had played with forking it to add some unique features but probably not worth maintaining at this point. We might consider bringing it back later.

PR #631, opened 2023-01-07
Add support for optional params

ghalle:
This PR adds support for optional params to the file system routing.

The syntax chosen is `[[param]]`, the first set of square brackets indicate that it is a parameter and the second set indicate that it is optional.
This syntax was chosen because `?` is not a valid character in filenames on Windows.

#### Usage example:

> **_src/routes/[[lang]]/blog/test.tsx_**

| path | params |
| ------------- | ------------- |
| /blog/test  | {} |
| /en/blog/test  | { lang: "en" }  |

### DIFF
--- a/packages/start/api/index.ts
+++ b/packages/start/api/index.ts
@@ -8,7 +8,30 @@ import { MatchRoute, Method, Route } from "./types";
 // @ts-ignore
 var api = $API_ROUTES;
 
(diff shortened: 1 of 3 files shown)


ryansolid:
Thanks. This has been overdue. I was thinking about it just the other day.

PR #633, opened 2023-01-07
Better MiniflareOptions docs

AlexErrant:
Took me some chin scratching to figure out Miniflare, `solid-start dev`, and `.env` files, so I documented it.


ryansolid:
Thank you

PR #644, opened 2023-01-14
feat: Add adapter include dir

ryoid:
Noticed that people were running into issues ([Related discord thread](https://discord.com/channels/722131463138705510/1063330181793714246)) using Prisma as they need to copy some files to the deployment directory. Not sure what other libraries require similar.

An adapter agnostic solution to better handle this use-case this could be to provide an adapter config option `include` for files to include.
```ts
solid({
    adapter: vercel({ 
      // Glob string or string[] of files to include
      include: ["node_modules/prisma/*engine*.node", "prisma/schema.prisma"]
    })
})
```

Note: It would flatten the directory, not sure if there is a use-case where the directory structure needs to be preserved (e.g. `prisma/schema.prisma` -> `.vercel/output/functions/render.func/schema.prisma`)

An alternative could be to stick with the current approach where users add a `postbuild` script that copies the files into `.solid/server` and the adapters copy any files found there into the deployment directory.


OrJDev:
this was used by using the vercel nft api - in https://github.com/solidjs/solid-start/commit/a0c71fc86412aa2b8b3fbe2703354a51af4c2e30

can be closed.

ryansolid:

PR #648, opened 2023-01-16
Bump vite dependency

SirSerje:
## Reason: 
Currently `"ts-jest-template-solid"` requires `"vite":"^3.1.8"` if I'll try to add solid-start (which is obviously required for routing and other necessities)

## Fix:
bump `vite` to 4.0.4

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "turbo": "^1.5.6",
     "typescript": "4.7.4",
     "undici": "^5.11.0",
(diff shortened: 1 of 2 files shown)


ryansolid:
This actually only ups the dependency of the docs site in the main repo. I merged this and realize didn't actually touch anything else, and updating makes the test fail. I'd be more than open to a PR that updates vite across the packages and fixes the tests. But this one isn't doing what you are expecting.

PR #676, opened 2023-01-21
Remove the need to `nohoist` (2nd attempt)

amitdahan:
# Summary

A 2nd attempt at applying #459, this time without breaking things :-) 

Basically my `requireCwd` was broken because it was using a directory as the base for `createRequire`, instead of a file path.
To fix this, I joined `cwd()` with `dummy.js`, similarly to what was [done in `import-from`](https://github.com/sindresorhus/import-from/blob/6063941cfac93086ae7f5a2e9ec2ac23f4bee401/index.js#L9).

---

**Side note:**
I noticed adapter auto-detection was removed entirely recently, in favor of only supporting explicit ones by directly importing them and providing them via `solidOptions.adapter`.

While this PR still fixes other cases of unsafe imports, I'll follow-up with another PR in case auto-detection is still something we want.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,56 +40,6 @@ npm install -g pnpm
 
 Run `pnpm install` to install all the dependencies for the packages and examples in your monorepo.
 
-<details>
-<summary><h4>Monorepo & <code>project.json</code> <code>"workspace"</code> support</h4></summary>
-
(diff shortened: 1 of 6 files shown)


amitdahan:
Hit some windows issues with the way I'm resolving stuff from CWD, back to draft

nksaraf:
The auto detection I have seen has usually led to issues where inadvertently people dont know what adapter is being used. Specially if they installed something earlier and are not using it now.

ryansolid:

Issue #682, opened 2023-01-22
createServerAction$ doesn't invalidate any query on the page


BigDog1400:
Bug description
In the docs, it says that using createServerAction$, should behave as the createRouteAction, since it is wrapper of this function, and in the createRouteAction section, it says that to refetch data after an action you have don't have to anything since the createRouteData functions will automatically refetch data after an action is performed.

But when you return a response with json() helper inside a createServerAction$, this is not happening. It only happens if you don't return an instance of the Response interface (or if you returne a Response instance with status 204, 301, 302, 303, 307 or 308, in which case it is a redirect, so it doesn't make sense), and this is bad since it doesn't allow you for example, set headers in the response.

Minimal, reproducible example
https://github.com/BigDog1400/solid-start-bug-mutation-bug-example (npm i + npm run dev)

Steps to reproduce
Go to the index page, click on "Generate", it should invalidate the current query, but it doesn't.

PR #683, opened 2023-01-22
feat: call handleRefetch() when the response is not a redirect

BigDog1400:
#682 
The createRouteAction.tsx, is already calling handleRefetch() when the response is not instance of Response interface.
```tsx
  .then(async data => {
        if (reqId === count) {
          if (data instanceof Response) {
            await handleResponse(data, navigate, options);
          } else await handleRefetch(data as any[], options);
          if (!data || isRedirectResponse(data)) setInput(undefined);
          else setResult({ data });
        }
        return data;
      })
      .catch(async e => {
        if (reqId === count) {
          if (e instanceof Response) {
            await handleResponse(e, navigate, options);
          }
          if (!isRedirectResponse(e)) {
            setResult({ error: e });
          } else setInput(undefined);
        }
```

When it is instance of Response, the handleResponse should call handleRefetch when it is not a redirect. I don't think it make sense to invalidate query only when is happening a redirect, should be exactly the contrary

### DIFF
--- a/packages/start/data/createRouteAction.tsx
+++ b/packages/start/data/createRouteAction.tsx
@@ -254,7 +254,7 @@ function handleResponse(response: Response, navigate: Navigator, options?: { inv
     }
   }
 
-  if (isRedirectResponse(response)) return handleRefetch(response, options);
+  if (!isRedirectResponse(response)) return handleRefetch(response, options);
 }
 
 function checkFlash<T>(fn: any) {
--- a/packages/start/data/createRouteAction.tsx
+++ b/packages/start/data/createRouteAction.tsx
@@ -254,7 +254,7 @@ function handleResponse(response: Response, navigate: Navigator, options?: { inv
     }
   }
 
-  if (!isRedirectResponse(response)) return handleRefetch(response, options);
+  return handleRefetch(response, options);
 }
 
 function checkFlash<T>(fn: any) {


BigDog1400:
I think that actually when a redirect happens, for example in a nested route and both routes share some common parent component layout, for example a parent route with path /users, and two childrens /users/:id, users/:id/details, we might want to execute query invalidation in the route data of /users. If that is the case, the handleRefetch() might always should be executed inside the handleResponse(). Not shure about this, i haven't tried it yet

BigDog1400:
I tested this with the nested layout case, and as expected if you have a parent route and go to another children path, it doesn't handle invalidations in the page. I guess that there is not need to check for `isRedirectResponse`, and just return the handleRefetch() always

nksaraf:
@ryansolid might make sense to do this?

BigDog1400:
Base on my last comment, this PR should be modify and replace this
`if (!isRedirectResponse(response)) return handleRefetch(response, options);` to this   `return handleRefetch(response, options);`, if there is no other reason to do the check in the response parameter

ryansolid:
The original code was

```js
 if (response.ok || isRedirectResponse(response)) return handleRefetch(response, options);
```
The idea was it refetched except on error. Unclear why the response.ok was removed.

ryansolid:
@BigDog1400 can you just update the PR with the condition removed.. I will merge and do a release.

BigDog1400:
> @BigDog1400 can you just update the PR with the condition removed.. I will merge and do a release.

Done!

PR #684, opened 2023-01-22
fix: CSR dev server evaluates user modules

OrJDev:
Taken from https://github.com/solidjs/solid-start/pull/595 
but this should fix the failing tests i believe?


OrJDev:
it only solved like 1-2 issues
Will close this for now and let Oren handle this

PR #685, opened 2023-01-22
feat(create-solid): prompt target if not supplied through args

juliusmarminge:
prompt user for target if they didn't supply one through argv

**Disclaimer:** I couldn't get the CLI to run locally (`pnpm build` errors, even tried `node cli/index.js` and that errored too). 

CleanShot 2023-01-22 at 23 02 18@2x

CleanShot 2023-01-22 at 23 02 45@2x

Not sure how you build or run this CLI but this should do it.

ref: 
CleanShot 2023-01-22 at 22 59 21@2x

### DIFF
--- a/packages/create-solid/cli/index.js
+++ b/packages/create-solid/cli/index.js
@@ -80,7 +80,16 @@ async function main() {
 
   let args = yargsParser(process.argv.slice(2));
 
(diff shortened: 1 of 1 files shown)


ryansolid:
Thanks

PR #693, opened 2023-01-28
fix route path for back navigation data fetching

sashoism:
Issue is quickly noticeable when performing any back navigation on https://solid-hn-islands.netlify.app/
This change fixes the problem by causing the full path (pathname + search + hash) to be fetched in `navigate()`.

While this is obviously correct for the search params, I'm not so sure about the hash. But it's also done for anchor links:
https://github.com/solidjs/solid-start/blob/7ab235584edb2bd37735614d15aa4d75f7bce62f/packages/start/islands/router.ts#L67
https://github.com/solidjs/solid-start/blob/7ab235584edb2bd37735614d15aa4d75f7bce62f/packages/start/islands/router.ts#L86
So if it's wrong, both should be changed to not include the hash.

### DIFF
--- a/packages/start/islands/router.ts
+++ b/packages/start/islands/router.ts
@@ -13,7 +13,7 @@ export default function mountRouter() {
     _$DEBUG("mounting islands router");
 
     const basePath = "/";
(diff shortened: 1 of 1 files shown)


ryansolid:

Issue #660, opened 2023-01-19
Vite 4 / Rollup 3 support


birkskyum:
The work towards this touches multiple PR's, so this is a tracking issue.

huijiewei:
Is there any plan to upgrade to vite4, because of the problem of the last / in the subdirectory, vite4 is very important

birkskyum:
I believe the last issue remaining before there is Vite 4 compatibility is the open issue wrt the spa-test referenced here

huijiewei:
Because the resolveConfig behavior of vite has changed, the defaultModeEnv parameter has been added, which needs to be set as soon as possible。 so solid-start not set this parameter. vite use default development env.

PR #694, opened 2023-01-29
Bump deps for tests

birkskyum:
Related to #660 

All deps except vite and @playwright/test are bumped to their latest.

It seems like @playwright/test breaks the spa-rendering test when bumping the minor version further (not aligning to semver), and looking into that is next step after this as it's the same unit test that vite 4 breaks.


ryansolid:
Thank you.

Issue #696, opened 2023-01-30
solid-start-aws: configurable clientAddress & method in entry.mjs


btakita:
@thdxr brought to my attention that `sst` is the target for the `solid-start-aws` package.

https://github.com/solidjs/solid-start/issues/689#issuecomment-1407727363

I'm using the raw `aws-cdk-lib`, which apparently create a different shape to `event.requestContext`.

I can add what I observe as a conditional into `entry.mjs`, however it seems like `event.requestContext` is dependent on the cdk stack configuration. Since `event.requestContext` can potentially have any shape, the `solid-start-aws` default function can include an option to configure the path to `clientAddress` & `method`.

thdxr:
in non edge mode it expects an API gateway v2 request if that helps

ryansolid:
In setting up for SolidStarts next Beta Phase built on Nitro and Vinxi we are closing all PRs/Issues that will not be merged due to the system changing. If you feel your issue was closed by mistake. Feel free to re-open it after updating/testing against 0.4.x release. Thank you for your patience.  See https://github.com/solidjs/solid-start/pull/1139 for more details.

PR #698, opened 2023-01-30
+ configurable clientAddress & method props to set a custom requestContext path for the clientAddress & method

btakita:
implements https://github.com/solidjs/solid-start/issues/696

@thdxr I'm interested to hear your take using `replace` to set the value & if there is a better approach, as this approach requires debugging on a lamda instance. I'm not sure how one would know the shape of `requestContext` while creating the cdk stack. I found the shape of `requestContext` but adding a `console.debug` on the deployed lambda.


thdxr:
The replace does seem a little weird

I'm a bit confused - can you clarify why you need to fix the client IP address and the HTTP method at build time?

btakita:
@thdxr I force pushed a better solution. It turns out that `aws-cdk-lib` has v1 & v2 which have different structures for the `event`.

I converted `entry.mjs` to `entry.ts` to express the multitude of possible types. In my situation, the lambda is receiving both v1 & v2 events so both need to be supported at this time.

btakita:
@thdxr I'm also looking into a solution to serve media types, which requires `isBase64Encoded` to be true in the `APIGatewayProxyResult`.

This would be `method === 'GET'` & the `Content-Type` header being a media type.

btakita:
I added another commit to serve files in the client directory & to set `isBase64Encoded` to binary media.

thdxr:
The PR for switching between v1 and v2 payloads is good

however, I don't think we want to support deploying in a way where the Lambda function serves the static assets - that breaks a few optimizations

in terms of cost I don't think serving out of S3 would be more expensive than serving from the function

btakita:
I think we should support `isBase64Encoded` regardless of supporting static files. In my case, I have routes that generate pdf files, which do require `isBase64Encoded: true`.

Incidentally, the pdf files in my app are too big (> 6mb) to be served by lambda, so I'm going with the app route generating the pdf, putting the pdf into a S3 Bucket, & redirecting to a route that is configured using `S3Origin`.

> however, I don't think we want to support deploying in a way where the Lambda function serves the static assets - that breaks a few optimizations

What are the optimizations? I can think of Lambda startup time being increased by including the static assets in the instance.

> in terms of cost I don't think serving out of S3 would be more expensive than serving from the function

From what I remember, lambda is orders of magnitude cheaper than s3 if static files can be served in < 10ms. Please let me know if I'm missing factors in the calculation.

https://aws.amazon.com/s3/pricing/
vs
https://docs.aws.amazon.com/whitepapers/latest/how-aws-pricing-works/aws-lambda.html
https://aws.amazon.com/lambda/pricing/

packages/start-aws/entry.ts
```diff
@@ -0,0 +1,64 @@
...
+import { default as manifest } from '../../dist/client/route-manifest.json';
+// @ts-ignore
+import { default as _server } from './entry-server.js';
+const server = _server as (event:FetchEvent)=>Promise<Response>;
+
+export async function handler(
+  event:APIGatewayProxyEvent|APIGatewayProxyEventV2
+):Promise<APIGatewayProxyResult> {
+  const clientAddress =
+    (event as APIGatewayProxyEvent).requestContext.identity
```
Daggron, line 15: Suggestion: We can use typescript `in` operator instead of as here for better ts safety.

ryansolid:

PR #702, opened 2023-01-31
feat(solid-start-vercel): use @vercel/nft to include un-bundable dependancies

arbassett:
This is a companion/alternative for Vercel pr to #644. [Discord discussion](https://discord.com/channels/722131463138705510/1063330181793714246)

Vercel functions expect all required files to be within the `*.func` folder while normally this is done automatically with the build process but some files can't be bundled notably prisma's `*engine.node` and `schema.prisma` files

in order to do this we use vercels nft (node file trace) package which is used is used internally by next. unlike next we can't just include a [[file].js.nft.json](https://vercel.com/docs/file-system-api/v2#configuration/bundling) file along side the handler due to it not being supported in their [Build Output v3](https://vercel.com/docs/build-output-api/v3). So what we have to do is recreate the folder structure of the app from the root of the filesystem within the `.func` folder. [astro does this to solve the same issue](https://github.com/withastro/astro/pull/3216#issuecomment-1124223700)

Example output
```
.vercel
├── output
│   ├── config.json
│   ├── functions
│   │   ├── api.func
│   │   │   ├── etc
│   │   │   │   └── os-release
│   │   │   └── vercel
│   │   │       └── path0
│   │   │           ├── dist
│   │   │           │   └── index.js
│   │   │           ├── node_modules
│   │   │           └── package.json
│   │   └── render.func
│   │       ├── etc
│   │       │   └── os-release
│   │       └── vercel
│   │           └── path0
│   │               ├── dist
│   │               │   └── index.js
│   │               ├── node_modules
│   │               └── package.json
│   └── static
│       ├── assets
│       │   ├── entry-client.912e70b6.css
│       │   ├── entry-client.bc62b30c.js
│       │   └── index.1502606f.js
│       ├── favicon.ico
│       ├── manifest.json
│       ├── route-manifest.json
│       └── ssr-manifest.json
└── project.json
```

why from the root? its due to a limitation of `@vercel/nft`. `@vercel/nft` will not return files that is outside of what you pass in as root which by default is the cwd. so if a package gets hoisted or files are included from a [sibling package](https://github.com/t3-oss/create-t3-turbo/tree/main/packages/db/prisma) we can't copy it

The alternative to this would be to detect if we are inside a workspace and use the workspace root as the root directory only have to create the folder structure from there. But I haven't found a consistent way to determine the package root of an unknown package manager. and would also require keeping up with changes to how workspaces are handled in the future.

since we no longer output directly to `.vercel/output` for server build we can follow what other adapters do and output the result of the build to `dist` and then copy it inside `.vercel/output/fucntions/[func].func/[workspacePath]`

- [x] Base case (not in workspace)
- [x] @vercel/nft warnings (some can just be ignored like trying to parse a .env file)
- [x] Include and exclude globs.
- [x] Symlinks 
- [x] Workspace support
- [x] Serverless deployment 
- [ ] Test edge deployment (need help testing)
- [ ] Test pre-render deployment (need help test)

### DIFF
--- a/packages/start-vercel/index.js
+++ b/packages/start-vercel/index.js
@@ -20,19 +20,29 @@ import { fileURLToPath, pathToFileURL } from "url";
  *
  */
 const copyDependencies = async ({ entry, outputDir, workingDir, cache }) => {
-  const { fileList, warnings } = await nodeFileTrace([fileURLToPath(entry)], { cache });
+  let base = entry;
+  while (fileURLToPath(base) !== fileURLToPath(new URL("../", base))) {
+    base = new URL("../", base);
+  }
+
+  const { fileList, warnings, reasons } = await nodeFileTrace([fileURLToPath(entry)], {
+    cache,
+    processCwd: process.cwd(),
+    base: fileURLToPath(base)
+    // base: fileURLToPath(new URL("/", import.meta.url))
+  });
+
   // TODO: handle warnings some of them can be ignored like .env or .md files
 
-  // remove self and package.json
-  const results = [...fileList].filter(
-    file => !entry.pathname.includes(file) && file !== "package.json" // TODO: handle user includes and excludes
-  );
+  // TODO: handle user includes and excludes
+  const results = [...fileList];
 
   for (const file of results) {
     // Create directories recursively
     mkdirSync(dirname(fileURLToPath(new URL(file, outputDir))), { recursive: true });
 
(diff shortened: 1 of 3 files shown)


arbassett:
to make working with absolute path easier i used URL objects.. for now i left in the equivalent `path.join` call in a comment for easier reading

ryansolid:
Thank you so much. Very much appreciated.

Issue #674, opened 2023-01-20
spa-test breaks in Vite 4 due to hydration (data-hk)


birkskyum:
Part of #660 

In spa-test, the combination of 'with index.html' and app.goto('/about') leads to this data-hk inconsistency in Vite 4:

<img width="713" alt="Screenshot 2023-01-20 at 14 43 00" src="https://user-images.githubusercontent.com/74932975/213709869-bc820993-b53d-4f53-94a1-627da7c4d9f8.png">

ryansolid:
Hmm.. that suggests it is getting the server rendered code instead of client rendered. Some configuration issue only present in Vite 4. Maybe the way some setting is merged? Hmm..

birkskyum:
Is there an issue made for the dev mode issue anywhere? Where are these reports?

birkskyum:
@ryansolid  It could very well be that the prod/dev issue emerged as a result of this breaking change in vite 4 where the way to set prod/dev flags was being reworked: https://github.com/vitejs/vite/pull/10996

huijiewei:
#753

PR #704, opened 2023-02-01
Bump playwright to 1.25.2

birkskyum:
@playwright/test breaks the spa-test  from version 1.26.0 ( same as Vite 4 does #674 ), so this bumps it to the latest version before that.

We bumping to 1.26 and beyond, this happens for both index.html and root.tsx:
Screenshot 2023-02-01 at 11 25 49


ryansolid:
Thank you for all your efforts here. Very appreciated.

Issue #707, opened 2023-02-02
spa-test: Node 18 required to bump @playwright/test >= 1.26.0


birkskyum:
Continuation of #704

When updating @playwright/test to 1.26.0 and beyond (same result all the way to latest 1.30.x), the "client side rendering" tests in spa-test will error. The rendered template is:

```tsx
<div id="content">
<h1>Root</h1>
<FileRoutes />
</div>
```

The `<FileRoutes />` cover two routes: index.tsx (path: "/") and about.tsx (path: "/about"). The content from these two routes is a simple page title (h2), and it is not included, breaking the test.

By manually adding said routes like below, the tests does pass, narrowing the issue down to incompatibility between **FileRoutes** and this new playwright version:

```tsx
import { Meta, FileRoutes, Scripts, Routes, Route } from "solid-start";
import { Suspense } from "solid-js";
import About from "./routes/about";
import Index from "./routes/index";

export default function Root() {
  return (
    <>
      <div id="content">
        <h1>Root</h1>
        <Routes>
          <Route path="/" component={Index} />
          <Route path="/about" component={About} />            
         </Routes>
      </div>
    </>
  );
}
```

birkskyum:
FileRoutes returns in both 1.26.0 and 1.25.2 the following:

```ts
[{"path":"/about"},{"path":"/"}]
```

ryansolid:
In setting up for SolidStarts next Beta Phase built on Nitro and Vinxi we are closing all PRs/Issues that will not be merged due to the system changing. If you feel your issue was closed by mistake. Feel free to re-open it after updating/testing against 0.4.x release. Thank you for your patience.  See https://github.com/solidjs/solid-start/pull/1139 for more details.

Issue #255, opened 2022-08-17
Typescript `tsc` can not exclude `node_modules/solid-start/`


P233:
When running the `tsc` command at a new created project root, I got the following error messages. Even I have already added `"skipLibCheck": true` and `"exclude": ["node_modules"]` to `tsconfig.json`. Is it something with `"types": ["vite/client"]` ?

```
...

Found 5 errors in 4 files.

Errors  Files
     2  node_modules/solid-start/entry-client/StartClient.tsx:3
     1  node_modules/solid-start/entry-server/StartServer.tsx:5
     1  node_modules/solid-start/islands/server-router.tsx:126
     1  node_modules/solid-start/types.ts:14
```

New project created via:

```
npm init solid@next
✔ Which template do you want to use? › bare
✔ Server Side Rendering? … yes
✔ Use TypeScript? … yes
```

version: [v0.1.0-alpha.99](https://github.com/solidjs/solid-start/releases/tag/v0.1.0-alpha.99)

viridia:
I noticed this too.

Whenever I create a new project, the first test I always write is a "compilation test" - meaning that the test passes if the code can compile without errors. However, when I attempt to compile my solid-start project, I see that:

1) The compilation takes a very long time to run (several minutes, when it would normally take one or two seconds.)
2) I get compilation errors from the solid-start package.

I suspect this is due to the way that solid-start's exports are set up in package.json.

luy19:
yes, solid-start contains a lot of ts errors, maybe the developer didn't use tsc to compile the code. 🤦‍♂️

luy19:
Good news, I found a workaround: https://github.com/microsoft/TypeScript/issues/38538#issuecomment-892555422

nirtamir2:
@luy19 Can you share your workaround for solid-start with TypeScript?

luy19:
@nirtamir2 here is my workaround, I hope this helps.
package.json
```
{
  "scripts": {
    "postinstall": "node add-ts-nocheck.cjs",
    "check": "tsc --noEmit"
  }
}
```
add-ts-nocheck.cjs
```
const fs = require("fs");

const ADDED_STR = "// @ts-nocheck\n\n";
const FILES = [
  "node_modules/solid-start/data/createRouteAction.tsx",
  "node_modules/solid-start/data/createRouteData.tsx",
  "node_modules/solid-start/data/Form.tsx",
  "node_modules/solid-start/entry-client/mount.tsx",
[...]

PR #709, opened 2023-02-03
Initial typings fs-router folder

birkskyum:
It's hard to dive in and debug the FileRouter ( #707 ), because the fs-router folder doesn't have types.

This PR converts some of the fs-router files to typescript, without breaking the tests, and adds some types as a starting point.

The "typecheck" on the start package is a bit problematic, and should maybe be temporarily disabled, as it make a gradual migration to typescript very hard - because it will break until a file is 100% typed. I did partial typings of a few more files and had to revert it to .js to make the typecheck pass.


packages/start/fs-router/fileRoutesImport.ts
```diff
@@ -1,7 +1,21 @@
-export default function fileRoutesImport({ types: t }) {
+// @ts-ignore
```
aminya, line 1: Why `@ts-ignore` has been added here?

aminya:
> The "typecheck" on the start package is a bit problematic, and should maybe be temporarily disabled, as it make a gradual migration to typescript very hard - because it will break until a file is 100% typed. I did partial typings of a few more files and had to revert it to .js to make the typecheck pass.

I guess it makes sense to ignore the errors using `@ts-ignore` until the issues are fixed. The TypeScript errors are also affecting the end-users, so the migration should be done (see #255).

packages/start/fs-router/fileRoutesImport.ts
birkskyum: Otherwise it'll warn:

Cannot find module '@babel/types/lib' or its corresponding type declarations.ts(2307)

aminya: Doesn't `[@types/babel__types](https://www.npmjs.com/package/@types/babel-types)` have the types?

birkskyum: I'll have another look at this

ryansolid:
This code runs in node.. unfortunately which means it chokes on ts as it doesn't get transformed breaking stuff. I will need to revert this unfortunately for now.

PR #710, opened 2023-02-03
add @auth/solid-start example

nksaraf:
Add an example for @auth/solid-start, I think it probably is the fastest way to do versatile auth right now with solid-start without any custom code. Also has the benefit of being a library that has wide usage amonst NextJS and SvelteKit users so familiarity for those people


nksaraf:
Sorry closing #708 already does this..

PR #714, opened 2023-02-04
Add Node 18 to matrix + disable cloudflare workers / windows workflows

birkskyum:
Prepare for Node 18, and make the test suite pass again

Until now, the practice has been to ignore failing cloudflare tests - they still spend a lot of time running though, and this PR disables that.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,7 +33,7 @@ jobs:
         adapter:
             - solid-start-node
             # - solid-start-cloudflare-workers
-            # - solid-start-cloudflare-pages
+            - solid-start-cloudflare-pages
             - solid-start-deno
         os:
             - ubuntu-latest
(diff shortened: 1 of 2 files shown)


ryansolid:
Do we need to disable windows? I think if I had disabled the cloudflare tests early I would have detected breaking windows faster.

birkskyum:
If we don't disable windows we need to figure out where they get stuck. I can't infer where from the logs unfortunately


on top of the windows/cloudflare workers causing trouble for 16&18, This PR exposes that cloudflare pages doesn't work with node 18

birkskyum:
close in favor of https://github.com/solidjs/solid-start/commit/a4c8b18c634dc79904af8d2ab46f245aea91abde

ryansolid:
I see.. yeah at the time I wrote the tests hadn't finished running.

Issue #715, opened 2023-02-04
[solid-start] missing `solid-start-node` peer dependency


rtritto:
## Summary
`solid-start-node` seems that it's a peer dependency of `solid-start`.
It should be added in peerDependencies of _package.json_ of `solid-start` (also all its transitive peer dependency should be added).

## Steps to reproduce
- `yarn init -y`
- `yarn set version canary`
- `yarn add @solidjs/meta @solidjs/router solid-js solid-start vite@3`
- `yarn solid-start dev`
- output:
```console
 solid-start dev
 version  0.2.19
C:\<PROJECT>\.pnp.cjs:11097
  return Object.defineProperties(new Error(message), {
                                 ^

Error: solid-start tried to access solid-start-node, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

Required package: solid-start-node (via "solid-start-node\package.json")
Required by: solid-start@virtual:ff47d5396d8139a19d431a35ae513d8c3ef61405b1dabaf86ed85031184a8cbb4fee78440c440f14b29d0f372620ff075e092b136136727884fbb7a33bb9d2c2#npm:0.2.19 (via C:\<PROJECT>\.yarn\__virtual__\solid-start-virtual-b5e4e18d9b\3\Users\Computer\AppData\Local\Yarn\Berry\cache\solid-start-npm-0.2.19-9596654667-9.zip\node_modules\solid-start\bin.cjs)

    at makeError (C:\<PROJECT>\.pnp.cjs:11097:34)
    at resolveToUnqualified (C:\<PROJECT>\.pnp.cjs:12815:21)
    at Object.resolveToUnqualified (C:\<PROJECT>\.pnp.cjs:12995:26)
    at resolve$1 (file:///C:/<PROJECT>/.pnp.loader.mjs:2009:31)
    at nextResolve (node:internal/modules/esm/loader:165:28)
    at ESMLoader.resolve (node:internal/modules/esm/loader:844:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:431:18)
    at ESMLoader.import (node:internal/modules/esm/loader:528:22)
    at importModuleDynamically (node:internal/modules/cjs/loader:1065:29)
    at importModuleDynamicallyWrapper (node:internal/vm/module:438:21)
```

OrJDev:
> 

it is not an actual peer dep, b
ecause you can also use ``solid-start-vercel`` or ```solid-start-netlify``` so there is no actual way to do that

[...]

PR #718, opened 2023-02-05
fix: set adapters as peer dependencies

OrJDev:
Closes https://github.com/solidjs/solid-start/issues/715


rtritto:
All peerDependencies and peerDependenciesMeta should be ascending sorted.

ryansolid:
What they finally added optional peer deps. I thought I'd never see the day. This is incredibly useful. Thanks for reporting and the fix.

Issue #690, opened 2023-01-25
`__vite_ssr_import_0__.template is not a function` when using `ag-grid-solid`


AlexErrant:
# Steps to reproduce:
1. `npm init solid@latest`, bare with no SSR
2. `pnpm i`
3. `pnpm i ag-grid-solid ag-grid-community`
4. add `import AgGridSolid from 'ag-grid-solid';` and `<AgGridSolid />` to `index.tsx`
5. `npm run dev`
6. This yields an error:
```
GET http://localhost:5173/
An unhandled error occured: ReferenceError: require is not defined in ES module scope, you can use import instead
This file is being treated as an ES module because it has a '.js' file extension and 'C:\Code\solid-start-bare\node_modules\.pnpm\ag-grid-solid@29.0.0_fjdpqd2jk35t7gako4obcr5l4e\node_modules\ag-grid-solid\package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
    at file:///C:/Code/solid-start-bare/node_modules/.pnpm/ag-grid-solid@29.0.0_fjdpqd2jk35t7gako4obcr5l4e/node_modules/ag-grid-solid/dist/cjs/index.js:3:11
    at ModuleJob.run (node:internal/modules/esm/module_job:194:25)
```
7. I add `ssr: { noExternal: ["ag-grid-solid"] }` to `vite.config.ts`, based on Ryan's answer [here](https://github.com/solidjs/solid-start/issues/13#issuecomment-890450148).
8. Restart `npm run dev`, and get a different error:
```
GET http://localhost:5173/
An unhandled error occured: TypeError: __vite_ssr_import_0__.template is not a function
    at ../../src/grid/core/solidCompWrapper.tsx:121:30
    at async instantiateModule (file:///C:/Code/solid-start-bare/node_modules/.pnpm/vite@3.2.5_@types+node@18.11.18/node_modules/vite/dist/node/chunks/dep-5605cfa4.js:53364:9)
```

[StackBlitz](https://stackblitz.com/~/github.com/AlexErrant/solid-start-bare) and [Github](https://github.com/AlexErrant/solid-start-bare) repro.

[`ag-grid-solid`](https://www.npmjs.com/package/ag-grid-solid) works just fine in vanilla solid-js.

ryansolid:
My guess is ag-grid is precompiled to be client only and breaks with SSR. I don't know it makes sense trying to server render it. What's the goal?

[...]

PR #721, opened 2023-02-05
`bump.cjs` updates example deps

AlexErrant:
Helps prevent issues like https://github.com/solidjs/solid-start/issues/690

Could also do others like `@solidjs/meta` or `@solidjs/router` - just say the word!


ryansolid:
To be fair I released that version after you reported the bug. But to be fair having the examples on the latest for new installs probably isn't the worst thing. Like it will do that automatically anyway within the range but like you only install a template once so why not.

AlexErrant:
Yeah I only realized that _after_ making the PR xD

PR #727, opened 2023-02-07
Update lockfile after solidjs 1.6.10 update

birkskyum:
1. Checked out main
2. Run `pnpm i`


ryansolid:
The new bump script actually requires we do this now on every release. I guess that's the side effect.

Issue #325, opened 2022-10-17
Error is thrown when building with solid-start-static


MrFoxPro:
```
/home/projects/vitejs-vite-dxqnej/src/routes
Error: Command failed: node /home/projects/vitejs-vite-dxqnej/node_modules/.pnpm/solid-ssr@1.5.7/node_modules/solid-ssr/static/writeToDisk.mjs /home/projects/vitejs-vite-dxqnej/.solid/server/server.js /home/projects/vitejs-vite-dxqnej/dist/public/*404.html /*404 --trace-warnings
[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "TypeError: event.env.getStaticHTML is not a function".] {
  code: 'ERR_UNHANDLED_REJECTION'
}

    at InternalError (https://vitejs-vite-dxqnej.w-corp.staticblitz.com/blitz.3e4f172269cb19a250be0dfdff1ab657cb850d0e.js:6:291731)
    at exithandler (https://vitejs-vite-dxqnej.w-corp.staticblitz.com/blitz.3e4f172269cb19a250be0dfdff1ab657cb850d0e.js:6:838089)
    at EventEmitter.prototype.emit (https://vitejs-vite-dxqnej.w-corp.staticblitz.com/blitz.3e4f172269cb19a250be0dfdff1ab657cb850d0e.js:6:155574)
    at maybeClose (https://vitejs-vite-dxqnej.w-corp.staticblitz.com/blitz.3e4f172269cb19a250be0dfdff1ab657cb850d0e.js:6:965920)
    at ChildProcess/this._handle.onexit (https://vitejs-vite-dxqnej.w-corp.staticblitz.com/blitz.3e4f172269cb19a250be0dfdff1ab657cb850d0e.js:6:956550)
    at _onChildExit/< (https://vitejs-vite-dxqnej.w-corp.staticblitz.com/blitz.3e4f172269cb19a250be0dfdff1ab657cb850d0e.js:15:380871)
    at _0x3e957c (https://vitejs-vite-dxqnej.w-corp.staticblitz.com/blitz.3e4f172269cb19a250be0dfdff1ab657cb850d0e.js:15:147507)
    at _0x35599c/< (https://vitejs-vite-dxqnej.w-corp.staticblitz.com/blitz.3e4f172269cb19a250be0dfdff1ab657cb850d0e.js:15:147247)
[...]

PR #735, opened 2023-02-09
throw an error when building with solid start static when ssr is disabled

simbas:
I got caught in the pitfall of thinking that solid start SSR option is for runtime SSR, so I disabled it when building a SSG app with the static adapter and got the error `getStaticHTML is not a function`.

This PR will throw a more descriptive error when building with static adapter and SSR disabled.

this is PR is related to the issue https://github.com/solidjs/solid-start/issues/325

### DIFF
--- a/packages/start-static/index.js
+++ b/packages/start-static/index.js
@@ -23,6 +23,7 @@ export default function () {
       return `http://localhost:${process.env.PORT}`;
     },
     async build(config, builder) {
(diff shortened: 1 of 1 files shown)


ryansolid:
Yeah this is probably a good idea for now. I think we should consider if it makes sense to support this in the future but I think it is causing more confusion than good.

Issue #397, opened 2022-11-03
Hydrated not initialized on load


atk:
The hydration scripts ([route].[hash].js) are prefetched, but only loaded on re-routing from another page. Events that are supposed to be there on load are missing.

Example:
* load https://atk.github.io/solid-cheat-sheet/ -> scroll bars show that tooltips are not positioned correctly
* load https://atk.github.io/solid-cheat-sheet/404.html, click link "Cheat Sheet" -> no scroll bars, tooltips are positioned correctly

JonahLund:
I have the same problem since `0.2.2`, should be fine if you downgrade to either 0.2.1 or 0.2.0

nksaraf:
Ohh is this a new problem.. hmm thats even weirded.. let me look at it.

JonahLund:
> Ohh is this a new problem.. hmm thats even weirded.. let me look at it.

Yeah, basically activeClass on <A> and a few more things are completely broken, not sure which specific version it occurred on, it's in the 0.2.1-0.2.3 range, but it seems that it seems like a good time to set up a few automatic e2e test runners to prevent these things from slipping through

atk:
I can confirm that it occurs on 0.2.2, but not in 0.2.1.

ryansolid:
I'm gathering this is still an issue with the latest?

atk:
It is initialized if the base is "/". It still does not work if you have a directory in front of your root path.

atk:
I have tested yesterday and the issue if you use a sub path still persists with the current version.

PR #742, opened 2023-02-12
fix: use base path in client router

atk:
- fixes https://github.com/solidjs/solid-start/issues/397
- allows us to use github pages with the static adapter.


ryansolid:
Thank you

Issue #739, opened 2023-02-11
Don't process clientOnly imports in SSR


arbassett:
[initial discord discussion](https://discord.com/channels/722131463138705510/1073754430664626176)

# Summary
Currently the way solid-start builds its apps its impossible to not process that are strictly client only .. meaning they can't be bundled for the server ex monaco-editor.. this causes the build to crash after the vite builds the server when rollup is used for the final result.

this is caused by changings [vites default behavior](https://github.com/solidjs/solid-start/blob/be97acfae1718cf2b5cc4156e747d9b64c4b292e/packages/start/bin.cjs#L310) to not inline dynamically imported files this prevents tools like rollup from doing dead code elimination. changing this to a false allows packages like monaco to be stripped out when behind a clientOnly. 

But the package will still be processed.. this is where vite's ssr.external config comes into play.. currently the [node adapter](https://github.com/solidjs/solid-start/blob/be97acfae1718cf2b5cc4156e747d9b64c4b292e/packages/start-node/index.js#L53) passes this value along to rollup and can build and not include the clientOnly boundary in the result.. but the other adapters currently don't do this.. if we patch the vercel adapter to do the same as node we can get it to build and start.

- [No patch Example](https://stackblitz.com/edit/solid-ssr-vite-becir2?file=src/components/Editor/index.tsx)
- [With Patch](https://stackblitz.com/edit/solid-ssr-vite-bkw6w8?file=vite.config.ts)

Patch files for reference
- [solid-start-vercel@0.2.20.patch](https://github.com/solidjs/solid-start/files/10712537/solid-start-vercel%400.2.20.patch)
- [solid-start@0.2.20.patch](https://github.com/solidjs/solid-start/files/10712538/solid-start%400.2.20.patch)

# Solution

While im fairly certain that allowing vite to split dynamic imports into multiple files by default is fine as they will get processed by the adpaters into a single file.. im uncertain if every adapter can currently respect the ssr.external config.. this issue relates to #725. you may want to flag something as being external to not process it, you may still want to be available at runtime. this works fine in vercel with #702 fixed but still has to be decided on how its handled for other adapters.

- [x] #744
- [ ] PRs to forward ssr.external to adapter rollup build
  - [x] #745
  - These require #725 to progress first to possibly include the external module but not process it
  - [ ] Deno? (need to research how deno handles npm packages)
  - [ ] Cloudflare-workers
  - [ ] Cloudflare-pages
  - [ ] Aws

[...]

PR #744, opened 2023-02-12
Allow vite to split dynamic imports to multiple files

arbassett:
related #739 

by not allowing vite to split dynamic imports rollup is unable to do any dead code elimination... with the current setup vite will always include dynamic imports even if they are never available like clientOnly in ssr. also vite's default output in esm so setting it to esm its redundant

How does this effect adapters?
- For adapters that use file output nothing is changed it will still output a single file
- For adapters that use dir output rollup will dump the extra files in the output folder which is just dist in all cases currently and from my testing the frameworks had no issue picking up the extra files.

to confirm that client only is removed i have a example repo here https://github.com/arbassett/solid-start-clientOnly-fix with the patch and committed dist dir to show that the clientOnly import was removed from the server build

### DIFF
--- a/packages/start/bin.cjs
+++ b/packages/start/bin.cjs
@@ -306,10 +306,6 @@ prog
             rollupOptions: {
               input: config.solidOptions.serverEntry,
               external: ssrExternal,
-              output: {
-                inlineDynamicImports: true,
-                format: "esm"
-              }
             }
           }
         });


ryansolid:
I really want to look back at why we added this. It feels like it was for a reason. But I'm glad you found the problem. I'm pretty sure routing removes dynamic imports on server anyway so it probably isn't that. It looks like the adapters pass so it isn't that either.  There are some consequences for synchronous SSR will dynamic imports but it probably doesn't matter enough.

arbassett:
I'm not familiar with how solid routers internals work but from the bundler perspective having two files and one file doesn't change the end result being a Promise based import

without patch
```javascript
const Lazy$2 = lazy(() => Promise.resolve().then(() => Lazy$1));
```

with Patch 
```javascript
const Lazy = lazy(() => import('./Lazy.b41f2dd3-bba4a033.js'));
```

hopefully this info helps your decision

ryansolid:
With the router we just remove the lazy on the server. We didn't always. The other other reason I could have seen was like quick solve for bundling for old Cloudflare stuff.. But like everything passes here so it is probably fine now. Some legacy thing long forgotten.

PR #745, opened 2023-02-12
Vercel & Netlify forward ssr.external to rollup build

arbassett:
related #739 

Its possible to have packages that can't be processed for a server bundle ex `monaco-editor`. you can tell vite to not process this with ssr.external config value but for vercel and netlify this is not forwarded to the final build.. normally we wan't to bundle everything that we can butthere should be a escape hatch for these cases which the node adapter already does.

if the file is truely needed it will be copied during the final parts of the build process to where it is needed unprocessed by rollup and vite
- #702 for vercel
- Netlify does this automatically in their ci platform with [zip-it-and-ship-it](https://github.com/netlify/zip-it-and-ship-it) 

this also works with #744 to remove client only imports with `unstable_clientOnly`

### DIFF
--- a/packages/start-netlify/index.js
+++ b/packages/start-netlify/index.js
@@ -16,6 +16,7 @@ export default function ({ edge } = {}) {
       proc.stderr.pipe(process.stderr);
     },
     async build(config, builder) {
+      const ssrExternal = config?.ssr?.external || [];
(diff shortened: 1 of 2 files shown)


ryansolid:
Awesome thanks.

PR #749, opened 2023-02-15
Allow prerenderRoutes as promise

wimbarelds:
Allow prerenderRoutes to be provided as a promise (ie: to fetch routes from CMS)

### DIFF
--- a/packages/start-static/index.js
+++ b/packages/start-static/index.js
@@ -64,7 +64,7 @@ export default function () {
           .map(a => a.path)
           .filter(a => (a.includes(":") || a.includes("/")) && !a.includes("*")),
         "/404",
-        ...(config.solidOptions.prerenderRoutes || [])
+        ...await (config.solidOptions.prerenderRoutes || [])
       ];
       renderStatic(
         routes.map(url => ({
--- a/packages/start-static/index.js
+++ b/packages/start-static/index.js
@@ -64,7 +64,7 @@ export default function () {
           .map(a => a.path)
           .filter(a => (a.includes(":") || a.includes("/")) && !a.includes("*")),
         "/404",
-        ...await (config.solidOptions.prerenderRoutes || [])
+        ...(await config.solidOptions.prerenderRoutes || [])
       ];
       renderStatic(
         routes.map(url => ({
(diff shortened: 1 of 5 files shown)


packages/start-static/index.js
```diff
@@ -64,7 +64,7 @@ export default function () {
           .map(a => a.path)
           .filter(a => (a.includes(":") || a.includes("/")) && !a.includes("*")),
         "/404",
-        ...(config.solidOptions.prerenderRoutes || [])
+        ...await (config.solidOptions.prerenderRoutes || [])
```
Daggron, line 67: The await outside of parentheses / brackets   may cause a small bug.

As the await is outside of the parenthesis / brackets, JS will evaluate ```(config.solidOptions.prerenderRoutes || [])``` first and if `config.solidOptions.prerenderRoutes` is a promise and not resolved it will be truthy and always return the pending promise. The `|| []` is of no use with the above syntax.

And if your promise returned null or undefined you will end up adding garbage values in routes

If you want to await for `config.solidOptions.prerenderRoutes` result and don't want to add anything garbage values to routes  in case promise returns falsy values then move the await inside the parenthesis 

```js
...(await config.solidOptions.prerenderRoutes || [])
```

This will make sure your promise is evaluated and JS has a value or undefined value  before comparing for fallback logic.

**NOTE** -> I am not a maintainer of this repo or contributor to this framework 😅 . I just felt like going through prs to contribute to the framework 😅 

You can correct me If my understanding is wrong.

wimbarelds: I put it outside assuming that, if it is a promise, then surely it would return an appropriate value. But I am not partial to it being in or outside.

wimbarelds: Placed await inside `()`. I don't think it matters/mattered very much, but I was adding type fixes anyway.

wimbarelds:
Added type corrections anywhere `prerenderRoutes` occured.

wimbarelds:
... Today just realized the whole Vite config can be a promise, or a function returning a promise. Therefore this PR isn't actually needed/useful.

ryansolid:
Probably a good idea for us to document. And thanks for bringing this to our attention.

PR #752, opened 2023-02-17
Check string "null" for req.headers.origin

superbull:
The value of `req.headers.origin` can be a string "null". We need to consider this situation.

See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin

### DIFF
--- a/packages/start/node/fetch.js
+++ b/packages/start/node/fetch.js
@@ -138,7 +138,9 @@ class NodeRequest extends BaseNodeRequest {
 }
 
 export function createRequest(req) {
-  let origin = req.headers.origin || `http://${req.headers.host}`;
+  let origin = req.headers.origin && 'null' !== req.headers.origin
+      ? req.headers.origin
+      : `http://${req.headers.host}`;
   let url = new URL(req.url, origin);
 
   let init = {


packages/start/node/fetch.js
```diff
@@ -138,7 +138,9 @@ class NodeRequest extends BaseNodeRequest {
 }
 
 export function createRequest(req) {
-  let origin = req.headers.origin || `http://${req.headers.host}`;
+  let origin = req.headers.origin && 'null' !== req.headers.origin
```
Daggron, line 141: **NOTE** : I am not a maintainer or contributor to this project in any way, was just using this framework and felt like I can help by reviewing prs as I don't know internal code yet 😅 

Here we are assigning origin variable a value after checking for 'null' values ?

Why do we need to do an `&&` with req.headers.origin it can be simple comparison.

```js
'null' !=== req.headers.origin
```

Am i missing something here 😅 ?

superbull: `req.headers.origin` can be `undefined` or `null`.

AlexErrant:
I'm not sure why we need `origin` anyway, since the `host` or `:authority` header _must_ exist as per [the spec](https://www.rfc-editor.org/rfc/rfc9110.html#section-7.2):

> A user agent MUST generate a Host header field in a request unless it sends that information as an ":authority" pseudo-header field.

[My related PR](https://github.com/solidjs/solid-start/pull/821).

Issue #713, opened 2023-02-04
[solid-start-node] missing `rollup` peer dependency


rtritto:
## Summary
`rollup` is a peer dependency of `@rollup/plugin-node-resolve` (dependency of _solid-start-node_).
It's a transitive peer dependency and it should be added in peerDependencies of _package.json_ of `solid-start-node`.

## Alternatives:
- copy the used/needed code of `@rollup/plugin-node-resolve` and remove it
- move `@rollup/plugin-node-resolve` to dev dependencies and add/use it during build

## Steps to reproduce
- `yarn init -y`
- `yarn set version canary`
- `yarn add @solidjs/meta @solidjs/router solid-js solid-start solid-start-node undici vite@3`
- output:
```console
➤ YN0000: ┌ Resolution step
➤ YN0060: │ solid-start-node@npm:0.2.19 [1528a] provides rollup (p92e50) with version 3.13.0, which doesn't satisfy what @rollup/plugin-node-resolve and some of its descendants request
➤ YN0000: │ Some peer dependencies are incorrectly met; run yarn explain peer-requirements <hash> for details, where <hash> is the six-letter p-prefixed code
➤ YN0000: └ Completed
➤ YN0000: ┌ Fetch step
➤ YN0000: └ Completed
➤ YN0000: ┌ Link step
➤ YN0000: └ Completed
➤ YN0000: Done with warnings
```

philipbordallo:
This is due to [`@rollup/plugin-node-resolve` at `3.13.0` wanting a peer dependency `rollup` at `^2.42.0`](https://github.com/rollup/plugins/blob/node-resolve-v13.3.0/packages/node-resolve/package.json#L54-L56) and that `solid-start-node` uses `rollup` at `^3.10.0`, which doesn't fulfill that peer requirement.

At one point `@rollup/plugin-node-resolve` was updated to `^15.0.1` (615a5f0) which would have resolved this but it looks like @ryansolid reverted it (20de367).

Ideally whatever was causing the weirdness with `v15` of `@rollup/plugin-node-resolve` can be fixed but in the mean time if that warning message is bothersome, you should be able to temporarily disable it via [yarn `packageExtensions`](https://yarnpkg.com/configuration/yarnrc#packageExtensions) and setting the `peerDependenciesMeta` to be optional.

rtritto:
Thanks for reply.
#753 should fix this.

rtritto:
[...]

PR #753, opened 2023-02-17
chore(deps): update vite and rollup to latest

huijiewei:
I update vite, rollup and other deps to latest.

test has spa error. because test vite config file use adapter: process.env.START_ADAPTER, i remove this option. test is ok.


huijiewei:
#660

birkskyum:
Can you elaborate a bit on what you did to make the spa test work? Where in the code

birkskyum:
For me this actually passes even with the spa-test left as is.

packages/start/bin.cjs
```diff
@@ -519,7 +519,7 @@ async function resolveConfig({ configFile, root, mode, command }) {
     DEBUG('config file: "%s"', configFile);
   }
 
-  let config = await vite.resolveConfig({ mode, configFile, root }, command);
```
birkskyum, line 522: This line, passing the mode to ViteJS, is what fixes the spa-test. It doesn't seem to need the removal of the adapter.

packages/start/bin.cjs
```diff
@@ -519,7 +519,7 @@ async function resolveConfig({ configFile, root, mode, command }) {
...
+  let config = await vite.resolveConfig({ mode, configFile, root }, command, mode, mode);
```
birkskyum, line 522: This line, passing the mode to ViteJS, is what fixes the spa-test. It doesn't seem to need the removal of the adapter.

birkskyum:
I've made a reduced PR here #767 , that only handles Vite 4. I was a bit worried that bumping this many dependencies at a time could introduce issues, like node-resolve 15: https://github.com/solidjs/solid-start/commit/20de367addbba2fc3876572bcb746ec8efee77e2

rtritto:
Fix #713

birkskyum:
This PR is a bit too big and has also fallen behind the main branch. @rtritto , if you make a new smaller PR that only has this small change you seek to fix 713, it would be a lot easier to review and merge it.

ryansolid:
Yeah I appreciate this but I went through and did a bit myself. It seems like node-resolve may be fine now. We are going to try it. I've updated everything else.

Issue #763, opened 2023-02-21
.gitignore is missing environment files (.env, .env.local)


colbyfayock:
The `.gitignore` file is currently missing the various `.env` files. Adding these could help newcomers (or me) avoid accidentally committing a `.env` file and subsequently exposing secrets.

https://github.com/solidjs/solid-start/blob/5668213646aab4dd0442b555f16be7619309f366/packages/create-solid/cli/index.js#L17

Per the Vite docs on naming conventions, I was thinking of the following:
```
.env
.env*.local
```

The Vite docs makes a comment on what is "ignored by git", which doesnt include `.env`, but it seems to me like an obvious one to ensure is also ignored.

https://vitejs.dev/guide/env-and-mode.html#env-files

Looking at others, Svelte currently does:

```
.env
.env.*
!.env.example
```

But that's assuming someone isn't using another pattern that they _want_ to commit, so the question is whether to be aggressive with ignoring to avoid accidental commits or require someone to add special cases on their own.

Next.js does:

```
.env*.local
```

Where I personally like the addition of `.env` as i've seen a variety of instances where people don't use the `.local` pattern, especially when the examples come from other services who are showing examples of how to integrate into an app.

Happy to put together this PR if there's an agreed upon list!

colbyfayock:
went ahead and spun up a PR to move things along in case the rec is fine "as is"

PR #766, opened 2023-02-22
Add .env to .gitignore

colbyfayock:
Adds `.env` patterns to `.gitignore` to avoid new Solid Start projects from committing their environment variables accidentally.

Thought here is `.env` in addition to `.env*.local` which will cover a variety of uses intended only for local environments.

More background behind thought on specific additions on Issue.

Fixes:  #763


ryansolid:
Thanks

PR #771, opened 2023-02-25
fix pnpm lockfile after 0.2.22

birkskyum:


ryansolid:
I need to remember to do that.. only this repo..ok.

PR #786, opened 2023-03-04
Fix vercel adapter type definition to allow optional config

HatulaPro:
#717 made it so that you have to pass the options parameter, so code like [this](https://github.com/solidjs/solid-start/blob/main/packages/start-vercel/README.md) (which used to work) now throws a TS error:

```typescript
export default defineConfig({
  plugins: [solid({ adapter: vercel() })]
});
```

This fix changes it back to the way it used to work.


ryansolid:
I ended up going with #843... which was the same fix more or less.. Thank you.

PR #792, opened 2023-03-07
server-functions consistent return between client and server

### LINKED ISSUE #768: Date type mismatch between Client and Server causing Runtime "TypeError"
### Describe the bug
When using `createServerData$(...)` for example, any date objects transferred from the server to the client will be serialized into strings and remain that way on the client, causing a type mismatch between the client and server version of the affected component.

This can cause visual differences or even errors when the client needs to rerender the component (Navigation, Hot Reload, ...) because the client expects a `Date` object to be present.

### Steps to Reproduce the Bug or Issue
Using the minimal example with SSR (and TypeScript), make the following changes to the index page:

```tsx
...
+   const date = createServerData$(() => new Date());
  
    return (
      <main>
+       <p>{date()?.toLocaleDateString()}</p>
...
```

The initial render is as expected. But when switching to the about page and switching back, an error occurs. The client fails to call `toLocaleDateString()`, as the function only exists on `Date` objects, not strings.

### Expected behavior
The client should turn the date string back into a `Date` object, preventing type issues like this.

### Screenshots or Videos

Screenshot from 2023-02-23 21-18-05

### Platform
- OS: Linux
- Browser: Chrome
- Version: v0.2.21

arbassett:
fixes #768 

server-functions are inconsistent between server renders and client renders.

When the server does a render server$ is [directly invoked ](https://github.com/solidjs/solid-start/blob/d0eb2f4a1d9a455b1211cb86eece34468e1e3313/packages/start/server/server-functions/server.ts#L242-L257)
```javascript
/**
*  currently in the context of the server render you can basically think this as const fn = () => new Data()
*/
const $$server_module0$1 = server$.createHandler(async function $$serverHandler0() {
  return new Date();
}, "/_m/0dbe216f23/routeData", true);
server$.registerHandler("/_m/0dbe216f23/routeData", $$server_module0$1);
```
and on the client server$ is a [fetch to the generate endpoint](https://github.com/solidjs/solid-start/blob/d0eb2f4a1d9a455b1211cb86eece34468e1e3313/packages/start/server/server-functions/browser.ts#L143-L154)

```javascript
/** 
* again you can think of this as const fn = () => fetch('/_m/0dbe216f23/routeData')
*/
const $$server_module0$1 = server$.createFetcher('/_m/0dbe216f23/routeData', false)
```

So when the server does a SSR render you get a raw Date object back from the server function while on the client its over a network boundary so it gets [serialized](https://github.com/solidjs/solid-start/blob/d0eb2f4a1d9a455b1211cb86eece34468e1e3313/packages/start/server/server-functions/server.ts#L192-L200) by the server function middleware. This causes issues when returning none serializable object like Date where on the server you can call `toLocaleDateString` and get a successful first render but when the client navigates to the page it throws

TRPC went though a similar issue with their types https://github.com/trpc/trpc/pull/3261 and what this PR is based on

- [ ] Make server-functions consistent between client and server
- [ ] update types
  - [x] server$
  - [ ] createServerData$
  - [ ] createServerAction$
  - [ ] createServerMultiAction$
- [ ] How does this effect customer seralizers

### DIFF
--- a/packages/start/server/server-functions/server.ts
+++ b/packages/start/server/server-functions/server.ts
@@ -4,14 +4,14 @@ import { FormError } from "../../data";
 import { ServerError } from "../../data/FormError";
 import {
   ContentTypeHeader,
+  isRedirectResponse,
   JSONResponseType,
   LocationHeader,
   ResponseError,
   XSolidStartContentTypeHeader,
   XSolidStartLocationHeader,
   XSolidStartOrigin,
-  XSolidStartResponseTypeHeader,
-  isRedirectResponse
+  XSolidStartResponseTypeHeader
 } from "../responses";
 import { PageEvent, ServerFunctionEvent } from "../types";
 import { CreateServerFunction } from "./types";
@@ -280,9 +280,8 @@ server$.hasHandler = function (route) {
 };
 
 server$.createServerResultHandler = function (handler) {
-  return (...args) =>
-    handler(...args).then((data: any) => {
-      // we need to emulate what respondWith does minus error handling.
+  return (...args) => {
+    const handleResponse = (data: any) => {
       if (data instanceof ResponseError) {
         return data.clone();
       } else if (data instanceof Response) {
@@ -295,9 +294,11 @@ server$.createServerResultHandler = function (handler) {
       ) {
         return JSON.stringify(data);
       }
-
-      return null;
-    });
+    };
+    const data = handler(...args);
+    data.then(handleResponse);
+    return data;
+  };
 };
 
 // used to fetch from an API route on the server or client, without falling into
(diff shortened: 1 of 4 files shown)


packages/start/server/server-functions/server.ts
```diff
@@ -279,6 +279,27 @@ server$.hasHandler = function (route) {
...
+        return data.clone();
+      } else if (data instanceof Response) {
+        return data;
+      } else if (
+        typeof data === "object" ||
+        typeof data === "string" ||
+        typeof data === "number" ||
+        typeof data === "boolean"
+      ) {
+        return JSON.stringify(data);
```
arbassett, line 296: This could be optimized to only do it on object but to make the check consistent with `repondWith` I left it as is

ryansolid:
This is a good point. I wonder if we should just use Solid's serializer that we use for resources. We did something simple initially.

Actually this probably more relevant for [Bling](https://github.com/TanStack/bling) (where we are currently in the process of extracting this into a reusable library) and where this will ultimately land.

PR #794, opened 2023-03-09
Fix error when use edge function in netlify adapter

progamesigner:
This fixes a bug in `entry-edge.js` that caused it to use the removed handler instead of importing `entry-server.js` directly. This bug was introduced in #770 and affected the edge function on Netlify.


ryansolid:
Thank you seems like we just missed that one.

PR #795, opened 2023-03-11
fix: vercel adapter normalize windows paths

arbassett:
[discord thread](https://discord.com/channels/722131463138705510/1083879274739675256)

trying to covert `process.cwd()` to a URL on windows would result in it thinking the drive is the protocol and get a invalid file path
`C:\` vs `file:///C:/`. vite exposes normalizePath which solves this issue


ryansolid:
I imagine this issue isn't restricted to the Vercel adapter?

arbassett:
> I imagine this issue isn't restricted to the Vercel adapter?

im not quite sure but its probably safe to assume so

PR #799, opened 2023-03-14
[start-static] Add index.d.ts

leader22:
Fixes this TS error.

```ts
// vite.config.ts
import solid from "solid-start/vite";
import adapter from "solid-start-static"; // 🔴 Could not find a declaration file for module 'solid-start-static'.

export default defineConfig({
  plugins: [
    solid({ adapter: adapter() }),
  ],
});
```


ryansolid:
Thanks

PR #804, opened 2023-03-17
Introduce resetOnSuccess prop for route action forms

knpwrs:
This is somewhat related to #640.

I have a comment form in my solid-start project that I would like to reset upon successful submission, and _only_ on successful submission. I'm having trouble figuring out how to do that cleanly. So far I have the following:

```tsx
  createEffect(() => {
    if (!submittingComment.pending && !submittingComment.error) {
      formRef.reset();
    }
  });
```

The idea being that if the submission is not pending and there is no error then it's safe to call `formRef.reset()`. One way this gets tricky though is if there are multiple instances of `submitComment.Form` on the page (for instance, submitting a new thread vs submitting a reply to a thread both using the same `submitComment.Form` component, and therefore the same handler). Now one must handle multiple refs and determine which one submitted.

This is particularly tricky for `createServerAction$` because the action handler doesn't run any code client-side.

Looking at my proposed solutions for #640, I don't think any of them would work here. I'm not entirely sold on what I have in this PR. I did `form =` in order to preserve backwards compatibility with refs being passed in.

Am I overlooking something simple? I would basically like to be able to have multiple instances of my form on a page and clear whichever one successfully submits whenever any of them successfully submits.

### DIFF
--- a/packages/start/data/createRouteAction.tsx
+++ b/packages/start/data/createRouteAction.tsx
@@ -90,19 +90,27 @@ export function createRouteAction<T, U = void>(
       }) as Promise<U>;
   }
   submit.url = (fn as any).url;
-  submit.Form = ((props: Omit<FormProps, "action" | "onSubmission">) => {
+  submit.Form = ((
+    props: { resetOnSuccess?: boolean } & Omit<FormProps, "action" | "onSubmission">
+  ) => {
     let url = (fn as any).url;
-    return (
+    let form = (
       <FormImpl
         {...props}
         action={url}
-        onSubmission={submission => {
-          submit(submission.formData as any);
+        onSubmission={async submission => {
+          await submit(submission.formData as any);
+          if (!result()?.error && props.resetOnSuccess) {
+            if (form instanceof HTMLFormElement) {
+              form.reset();
+            }
+          }
         }}
       >
         {props.children}
       </FormImpl>
     );
+    return form;
   }) as T extends FormData ? ParentComponent<FormProps> : never;
 
   return [


knpwrs:
Something that I have working now is:

```tsx
  let submittingCommentForm: HTMLFormElement | null = null;

  function setSubmittingCommentForm(e: SubmitEvent) {
    if (e.target instanceof HTMLFormElement) {
      submittingCommentForm = e.target;
    } else {
      submittingCommentForm = null;
    }
  }

  createEffect(() => {
    if (!submittingComment.pending && !submittingComment.error) {
      submittingCommentForm?.reset();
      submittingCommentForm = null;
    }
  });
```

And then I pass `setSubmittingCommentForm` to `onSubmit` on each form. This works, but something built-in or otherwise simpler would be great.

One idea that comes to mind is setting a `form` property at the same time `pending`, `error`, and `result` are set.

ryansolid:
Would having an onComplete, or onSuccess/onError hooks help here? I'm just thinking it might be better for these sort of updates to go a more event. Just thinking if the mechanism needs to be general beyond forms.

knpwrs:
Those hooks would be great, especially if within the context/arguments of the callback I could get a reference to the submitted form instance.

Are you thinking as a parameter to `createServerAction$` or as props on `Form`?

Tc-001:
I would love to see it on the form, that way a oneliner like this will do a reset:
```jsx
<Form
  onSuccess={e => e.target.reset()}
/>
```
While also being easy to expand.

(Plus there already is the `createEffect` approach for usecases where you don't need the form element)

adoublef:
Ah that's all good, I seem to be able to get the formData (I can link my sandbox to see if the behaviour is the same as you needed) inside both the `createServerAction` hook and my API route.

The only flaw is that if you wanted to use the input value for some optimistic UI I haven't actually tested thoroughly yet but looks like that isn't possible

adoublef:
Regarding this I was able to come up with this variation but as a hook

```ts
// hook
function useFormReset(submit: RouteAction<FormData, unknown>[0]) {
    let form: HTMLFormElement | null = null;
    createEffect(() => {
        if (!submit.pending && !submit.error && form) (form.reset());
    });

    return (el: HTMLFormElement) => form = el;
}

// usage
const formRef = useFormReset(submit);
return (
  <Form ref={formRef}>
    //...
  </Form>
)
```

unsure if this adds anything to the thread but felt like sharing

PR #810, opened 2023-03-22
add support for custom CSR server root

arbassett:
Apps that are using CSR only can't set custom head tags in dev.. for prod you can create your own custom `index.html` but you have to setup your client entry manually

[stackblitz](https://stackblitz.com/edit/github-1j7k1j-o4k7uq?file=patches%2Fsolid-start%400.2.23.patch,src%2Froot-csr.tsx)
[Discord Issue](https://discord.com/channels/722131463138705510/910635844119982080/1087870164319613018)

This adds another root file `root-csr.tsx`. if one is present in CSR mode it will be used. if its not present default to [CsrRoot.tsx](https://github.com/solidjs/solid-start/blob/main/packages/start/dev/CsrRoot.tsx)

some things to note
 - any children on Body will be ignored
 - you can use solid-start `Meta` & `Title` to match what the client is rendering
   - if you use a normal meta tag that is also rendered with a Meta on the client it will be duplicated
   - if you don't include the same `Meta` or `Title` on the client it will be removed on render
 - anything else under Head will not be hydrated
   
 
 ```tsx
 // root-csr.tsx
 
 import { Body, Head, Html, Meta, Title } from "solid-start";

export default function Root() {
  return (
    <Html lang="en">
       <Head>
        {/* these will be "hydrated" (removed and re-rendered) on the client */}
        <Title>SolidStart - Bare</Title>
        <Meta charset="utf-8" />
        <Meta name="viewport" content="width=device-width, initial-scale=1" />
        {/* this will be persevered */}
        <meta name="test" content="test" />
        {/* this will be duplicated since we are rendering it with Meta as well */}
        <meta name="viewport" content="width=device-width, initial-scale=1" />
      </Head>
      <Body />
    </Html>
  );
}
 ```

### DIFF
--- a/packages/start/vite/plugin.js
+++ b/packages/start/vite/plugin.js
@@ -339,10 +339,11 @@ function solidStartFileSystemRouter(options) {
         return {
           code: code.replace(
             "var fileRoutes = $FILE_ROUTES;",
-            !options.ssr && ssr ? "var fileRoutes = [];" :
-            stringifyPageRoutes(router.getNestedPageRoutes(), {
-              lazy: ssr ? false : true
-            })
+            !options.ssr && ssr
+              ? "var fileRoutes = [];"
+              : stringifyPageRoutes(router.getNestedPageRoutes(), {
+                  lazy: ssr ? false : true
+                })
           )
         };
       } else if (code.includes("var routeLayouts = $ROUTE_LAYOUTS;")) {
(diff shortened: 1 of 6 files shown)


arbassett:
looking for suggestions on the file name. I'm not super sold on the `root-csr.tsx` name to me that implies this is the root for **_client side_** routing.

arbassett:
Looks like the [lock file is out of sync due to this commit](https://github.com/solidjs/solid-start/commit/fd5eba8dbcfce2a3f439d29138d0cf0ef7607155).. for now i just added it to this PR so tests will run

packages/start/vite/plugin.js
```diff
@@ -339,10 +339,11 @@ function solidStartFileSystemRouter(options) {
         return {
           code: code.replace(
             "var fileRoutes = $FILE_ROUTES;",
-            !options.ssr && ssr ? "var fileRoutes = [];" :
-            stringifyPageRoutes(router.getNestedPageRoutes(), {
-              lazy: ssr ? false : true
-            })
+            !options.ssr && ssr
```
arbassett, line 342: Nothing changed here. just prettier formatting

packages/start/vite/plugin.js
```diff
@@ -339,10 +339,11 @@ function solidStartFileSystemRouter(options) {
...
+              ? "var fileRoutes = [];"
+              : stringifyPageRoutes(router.getNestedPageRoutes(), {
+                  lazy: ssr ? false : true
+                })
```
arbassett, line 346: Nothing changed here just prettier formatting

ryansolid:
@nksaraf I want your thoughts on the general direction. My understanding was our intention was to render index from root.tsx The fact that index.html works is incidental? We did do something perhaps odd in dev to allow for isolation but I want to get an idea of what the preferred solution is so we can build towards that. Atleast to understand what you were envisioning.

ryansolid:

PR #819, opened 2023-03-31
allow internalFetch to pass env and locals

rvlzzr:
With Cloudflare at least, `internalFetch` is mostly useless in many situations, as environment variables are only available through the event, and are reset to `{}` by `internalFetch`.

Also, `locals` is the most common place for middleware to store request context, and as middleware isn't run again when `internalFetch` is used, it's useful to be able to pass this through from the parent request.

### DIFF
--- a/packages/start/api/internalFetch.ts
+++ b/packages/start/api/internalFetch.ts
@@ -8,7 +8,7 @@ export const registerApiRoutes = (routes: MatchRoute[]) => {
   apiRoutes = routes;
 };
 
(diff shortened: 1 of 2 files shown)


rvlzzr:
An alternative would be to change the second argument to an `APIEvent` and just override `request`, `params`, `clientAddress`, `$type`, and `fetch`, and pass through the rest. This way we wouldn't need to add more arguments if more properties are added to `APIEvent` in the future. But that would be a breaking change.

PR #821, opened 2023-04-02
Fix: HTTPS + dev + API route = `http://undefined/` as the request url

AlexErrant:
Currently, if you use a self-signed cert, HTTPS, and try to use an API route on dev, the `request.url` looks like `http://undefined/`.

## Example Setup

This is best exemplified in the `with-authjs` example, whose [README I've updated](https://github.com/solidjs/solid-start/blob/92dd6e94e811801968d3ad86f9e3629b3359e5e9/examples/with-authjs/README.md) in [the first commit](https://github.com/solidjs/solid-start/pull/821/commits/92dd6e94e811801968d3ad86f9e3629b3359e5e9). Follow the instructions there, creating a `.env` file, Discord app, and self-signed cert. (Apologies in advance for how long this may take.) When you click "Sign in with Discord" using that self-signed cert, it will redirect you to `http://undefined/api/auth/signin/discord`. To demonstrate that this isn't a problem with `authjs`, go to [`with-authjs/src/routes/api/auth/[...solidauth].ts`](https://github.com/solidjs/solid-start/blob/main/examples/with-authjs/src/routes/api/auth/%5B...solidauth%5D.ts) and replace `export const { GET, POST } = SolidAuth(authOptions);` with

```ts
import { PageEvent } from "solid-start";

export const POST = (x: PageEvent) =>  {
  console.log(x.request.url)
  return SolidAuth(authOptions).POST(x);
};

export const GET = (x: PageEvent) => {
  console.log(x.request.url)
  return SolidAuth(authOptions).GET(x);
};
```

If you revisit the login screen, you'll see the following in your console:

```
GET https://127.0.0.1:3000/
GET https://127.0.0.1:3000/api/auth/csrf
http://undefined/api/auth/csrf
POST https://127.0.0.1:3000/api/auth/signin/discord?
https://localhost:3000/api/auth/signin/discord?
GET https://127.0.0.1:3000/api/auth/signin?csrf=true
http://undefined/api/auth/signin?csrf=true
```

## The Fix

The fix is implemented in [the second commit](https://github.com/solidjs/solid-start/pull/821/commits/c46ced2e755379fb720fe397347ddd3dca918bfb). (Ignore the multiple force pushes below - I strived for a clean commit history.)

Here's a somewhat relevant section from [the spec](https://www.rfc-editor.org/rfc/rfc7540#section-8.1.2.3) as to why the `host` header is missing; emphasis mine:

> Clients that generate HTTP/2 requests directly SHOULD use the ":authority" pseudo-header field _instead of_ the Host header field.

Note that this only works for HTTP/2 - I don't think it'll work for HTTP/1.1. From [the 1.1 spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html):

> A client MUST include a Host header field in all HTTP/1.1 request messages

I'm unsure if we can get the correct URL anywhere else in Node. See [the comment](https://github.com/solidjs/solid-start/pull/821#discussion_r1155373068) below for details.

`createRequest` has 3 callers, [`packages/start/dev/server.js`](https://github.com/solidjs/solid-start/blob/8ef83015d5e1af8acc3c0a520d96d3ee8fac8aad/packages/start/dev/server.js#L116), [`packages/start-node/server.js`](https://github.com/solidjs/solid-start/blob/8ef83015d5e1af8acc3c0a520d96d3ee8fac8aad/packages/start-node/server.js#L44), and [`packages/start-static/entry.js`](https://github.com/solidjs/solid-start/blob/8ef83015d5e1af8acc3c0a520d96d3ee8fac8aad/packages/start-static/entry.js#L11), so this PR has security implications for production deployments.


packages/start/node/fetch.js
```diff
@@ -137,8 +137,29 @@ class NodeRequest extends BaseNodeRequest {
...
+ * @param {import('http').IncomingMessage} req
+ */
 export function createRequest(req) {
-  let origin = req.headers.origin || `http://${req.headers.host}`;
+  let origin;
+  if (req.httpVersionMajor === 1) {
+    origin = req.headers.origin || `http://${req.headers.host}`;
+    // It's impossible for `host` to be empty since :authority doesn't exist on HTTP/1.
+    // https://www.rfc-editor.org/rfc/rfc9110.html#section-7.2 A user agent MUST generate a Host header field in a request unless it sends that information as an ":authority" pseudo-header field
+    // We initially try the origin header because it has the scheme. The host header doesn't, and so defaults to HTTP
```
AlexErrant, line 150: I would ideally like to use only the host header but I don't know how to get the scheme. `createRequest` has 3 callers, but I only know how to pass in the scheme to [`packages/start/dev/server.js`](https://github.com/solidjs/solid-start/blob/8ef83015d5e1af8acc3c0a520d96d3ee8fac8aad/packages/start/dev/server.js#L116) (use `viteServer.resolvedUrls.local[0]`). I'm not good enough at Node (or Javascript) to know how to get the scheme from [`packages/start-node/server.js`](https://github.com/solidjs/solid-start/blob/8ef83015d5e1af8acc3c0a520d96d3ee8fac8aad/packages/start-node/server.js#L44) or [`packages/start-static/entry.js`](https://github.com/solidjs/solid-start/blob/8ef83015d5e1af8acc3c0a520d96d3ee8fac8aad/packages/start-static/entry.js#L11).

AlexErrant:
Svelte solves this issue [here](https://github.com/sveltejs/kit/pull/3553) by forcing HTTP/1. Note that this solution uses the fact that Vite ["downgrades to TLS only when the `server.proxy` option is also used"](https://vitejs.dev/config/server-options.html#server-https). [We hardcode `http` though](https://github.com/solidjs/solid-start/pull/821/files#diff-1541a6df41c3ea0f14bb35807d6a4b16f4c3109e0b5768adc0717803b2ef4fd1R149), so that'll need to be switched to `https`... somehow. (In my `examples/with-authjs` reproduction, `req.headers.origin` is `undefined`.) I've still no idea how to get the scheme.

FYI this isn't blocking me, so I'm in no rush to merge. I did try to use [pnpm patch](https://pnpm.io/cli/patch) and it worked great, but then I realized that my "real" code uses Cloudflare Workers... not Node... I switched to Node when attempting to create a minimal reproduction and never looked up. 🤦‍♂️ Cloudflare has its own problem that I discuss (and resolve) [here](https://github.com/solidjs/solid-start/issues/834). (I'm a good developer and know exactly what I'm doing 😊)

ryansolid:

Issue #818, opened 2023-03-30
Pnpm version can't update button


LuizzAugusto:
I can't make button update in pnpm version (Hello World - Bare Template). An error message appear on inspect `Uncaught The browser resolved template HTML does not match JSX input:
<solid-children></solid-children>

<solid-children>. Is your HTML properly formed? `

This error not occur on npm version.

LuizzAugusto:
Actually the problem is not in create the templates with pnpm, but install the libraries with it.

dayblox:
Full error message:
```
Uncaught The browser resolved template HTML does not match JSX input:
<solid-children></solid-children>

<solid-children>. Is your HTML properly formed?
```

ryansolid:
This is the result of 1.6.x version of Solid, and 1.7.x version of our build plugins. If you update Solid to 1.7 it should be fine.

dayblox:
You are right, upgrading everything (including solid) to latest did the job
`pnpm up -L`

PR #825, opened 2023-04-03
Fix broken examples

AlexErrant:
Closes https://github.com/solidjs/solid-start/issues/818


AlexErrant:
Also requiring PNPM v7 since CICD uses that. PNPM v8, which recently came out, is not backwards compatible.

ryansolid:
I ended up doing most of the updates myself. Had to fix a few tests. Thanks for doing this though.

PR #843, opened 2023-04-15
fix vercel: make input optional

OrJDev:
This will not the force the user to pass an empty object when using the vercel adapter


AlexErrant (approved):
[The adapter defaults to `{}`](https://github.com/solidjs/solid-start/blob/main/packages/start-vercel/index.js#L123), so this looks fine to me. Ryan's eager to abandon the adapters though for Astro so we'll see if/when this gets merged lol.

PR #868, opened 2023-05-03
Fix solid-start-cloudflare-workers (durable objects & crypto)

shorkyw:
Hi, I was attempting to use solid-start-cloudflare-workers with a durable object but noticed it didn't seem to work. I saw that the issue was that the durable object was never exported in the miniflare script, I have resolved this by making it so all durable objects passed into miniflare will be exported by the script, if you can spot any issues with this implementation please let me know.

This also fixes the globally assigning crypto error.


shorkyw:
Just updated the with-websockets example to work with the updated adapter and added some info about how to utilise the adapter with durable object.

ryansolid:

PR #877, opened 2023-05-09
Update what-is-solidstart.md

DaveMadden:
Just fixing a tiny typo that doesn't *really* matter.
But I've never contributed to open source before so figured this would be a good PR to start with.


ryansolid:
Thanks

PR #893, opened 2023-05-24
Use port provided in astro config if present

rvlzzr:
The `config.server.port` from astro.config.ts  is currently ignored, which feels like a bug.


ryansolid:
Thank you

PR #895, opened 2023-05-27
fix "astro build"

rvlzzr:
This doesn't seem to be looking for the correct path in the key.


ryansolid:
Thanks.. difference between the local repo and deployed node_modules I imagine.

PR #901, opened 2023-06-05
add with-cssed example

opudalo:
[CSSED](https://github.com/okotoki/cssed) literally dumbest and simplest css-in-js implementation. Think solid ecosystem can benefit from something like this. Unlike styled-solid it does not provide any capabilities at runtime – you just write your CSS as is, including support for media queries and so on.


ryansolid:

Issue #905, opened 2023-06-13
build error node:18: Error: connect ECONNREFUSED


sirenkovladd:
https://github.com/nodejs/node/issues/40702

https://github.com/solidjs/solid-start/blob/main/packages/start/bin.cjs#L401

```
making HTTP(S) head request to  url:http://localhost:{port}/ ...
  HTTP(S) error for http://localhost:{port}/ Error: connect ECONNREFUSED ::1:{port}
```

PR #906, opened 2023-06-13
Fix econnrefused node18

sirenkovladd:
https://github.com/solidjs/solid-start/issues/905


ryansolid:
Hmm.. this change seems to do the opposite for me. It seems to breaks a number of things where it was working before. I am running Node 18 as well so the initial cause must be more involved than this. I can revert the places it breaks but then I wonder about the rest of the changes. Safer for me to revert. It's unfortunate the package lock made it difficult to run these tests in CI. I will have to look into that.

PR #919, opened 2023-06-28
Fix/docs build failing

caseybaggz:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] CI related changes
- [x] Documentation content changes
- [ ] infrastructure changes

The docs deploy task is failing due to a broken example shipped.

## What is the new behavior?
- Adds new CI tasks that validates the docs successfully build when PR's are submitted that change any file in the docs dir
- Fixes the broken example

## Other information

### DIFF
--- a/docs/api/server.md
+++ b/docs/api/server.md
@@ -8,7 +8,7 @@ active: true
 
 # server$
 
-##### `server$` takes a function that should only run on the server and compiles its usage into an RPC call to the server.
+Perform actions on the server environment only (i.e. console logging, etc.).
 
 <div class="text-lg">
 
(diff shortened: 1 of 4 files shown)


docs/api/server.md
```diff
@@ -40,13 +42,18 @@ function Component() {
 }
 ```
 
-In this code snippet regardless of whether we are server rendering this component or client rendering it, the `logHello` function generates a log on the server console only. How does it work? We use compilation to transform the `server$` function into an RPC call to the server. 
+In this example, regardless of whether we are rendering this component on the server or in the browser, the `logHello` function generates a log on the server console only. How does it work? We use compilation to transform the `server$` function into an RPC call to the server.
```
michaelessiet, line 45: Add the meaning of an RPC call somewhere, maybe as an aside note or a footnote. This will help the less savvy readers grasp the concept of server$

michaelessiet (changes requested):

PR #927, opened 2023-07-06
RouteAction returns undefined in case of error

frenzzy:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes
- [x] TypeScript typings

In case of error `createServerAction$` returns undefined as a result, but typings don't:
```tsx
import { FormError } from 'solid-start'
import { createServerAction$ } from 'solid-start/server'

  const [action, submit] = createServerAction$(async (formData: FormData) => {
    if (formData.get('hasError')) {
      throw new FormError('Validation failed')
    }
    return 'demo'
  })
  return (
    <submit.Form
      onsubmit={async (event) => {
        event.preventDefault()
        const formData = new FormData(event.currentTarget)
        const result = await submit(formData)
        console.log(action.result === result) // true
        console.log(action.result) // type: string | undefined
        /* ACTUAL:   */ console.log(result) // type: string
        /* EXPECTED: */ console.log(result) // type: string | undefined
      }}
    >
      <input type="checkbox" name="hasError" value="yes" />
      <input type="submit"/>
    </submit.Form>
  )
}
```

## What is the new behavior?
Typings reflect the actual behavior

### DIFF
--- a/packages/start/data/createRouteAction.tsx
+++ b/packages/start/data/createRouteAction.tsx
@@ -26,14 +26,14 @@ export type RouteAction<T, U> = [
     clear: () => void;
     retry: () => void;
   },
-  ((vars: T) => Promise<U>) & {
+  ((vars: T) => Promise<U | undefined>) & {
     Form: T extends FormData ? ParentComponent<FormProps> : never;
     url: string;
   }
 ];
 export type RouteMultiAction<T, U> = [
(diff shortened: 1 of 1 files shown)


ryansolid:
Thanks

PR #934, opened 2023-07-11
Promoting debug types dev dep to regular dep, enhancing create script

FoxtrotPerry:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

When attempting to run `npx tsc` in a personal project using `solid-start`,  the following error was being thrown:
```
⟩ npx tsc
node_modules/solid-start/types.ts:1:31 - error TS7016: Could not find a declaration file for module 'debug'. '/home/perry/dev/mcmanager-solid-app/node_modules/debug/src/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/debug` if it exists or add a new declaration (.d.ts) file containing `declare module 'debug';`

1 import type { Debugger } from "debug";
                                ~~~~~~~


Found 1 error in node_modules/solid-start/types.ts:1
```

## What is the new behavior?

So what this PR does is allows solid-start to download the required `@types/debug` dependency such that the end user no longer has to run `npm i --save-dev @types/debug` in order to get types for `debug` in their node_modules folder and get a successful execution out of `tsc`.

I also went ahead and expanded the list of dependencies (both dev and regular) that get removed if the user opts out of using typescript when using the `create-solid` CLI tool.

## Other information

In the future, I would be very much interested in helping with a transition towards using (possibly emitted) `*.d.ts` files [in cases like these](https://github.com/solidjs/solid-start/blob/main/packages/start/types.ts#L4) so that `skipLibCheck` will be more potent when used with solid-start. Until then though, I thought I would make this PR since it's just the one error that I ran into! Let me know what you think of the approach!


ryansolid:
Yeah I think this is fine for now. We are in the process of rebasing SolidStart on a different deployment architecture. A lot of the lower level code changes (although not the source that people import or the way they author). A lot of code goes away so I think it would be a better time then to revisit.

PR #936, opened 2023-07-13
fix(data): update Form types in createRouteAction.tsx

caseybaggz:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

When you use the `createRouteAction` controller using the `Form` it returns without an argument, the types are set to `never` which results in an TS/IDE error.

This is because the typing incorrectly forces the return value based on if someone passes `FormData` as an argument which may not always be the case.

## What is the new behavior?
- Updates the types to use the correct return for the component while utilizing the type as a Union/alternate so the IDE or TS doesn't error, allowing users to still use the `Form` magic component without needing arguments.
- Also fixes current actual type errors in the file from not using the `unknown as` pattern.

## Other information

Example use case, when using Supabase Auth, the API's have to run on the client and `signOut` doesn't need arguments and to keep the UI a11y to all users, you should use a form for the button submission.

```typescript
const [, { Form: SignOutForm }] = createRouteAction(async () => {
    return signOut()
  })
```

### DIFF
--- a/packages/start/data/createRouteAction.tsx
+++ b/packages/start/data/createRouteAction.tsx
@@ -27,14 +27,14 @@ export type RouteAction<T, U> = [
     retry: () => void;
   },
   ((vars: T) => Promise<U>) & {
-    Form: T extends FormData ? ParentComponent<FormProps> : never;
+    Form: ParentComponent<FormProps | T>;
     url: string;
   }
 ];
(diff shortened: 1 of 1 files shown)


ryansolid:
Yeah I guess we were trying to be too strict.

PR #955, opened 2023-07-27
fix setting multiple set-cookie headers on response in undici >= 5.19.0

### LINKED ISSUE #953: [Bug?]: Setting multiple Set-Cookie headers is not handled properly on undici version >=5.19.0
### Duplicates

### Latest version

### Current behavior 😯

Since `undici` version 5.19.0, setting multiple cookies in `createServerAction$` with the `Set-Cookie` header isn't possible anymore. It only returns the last(?) cookie which was appended to the header.

I hope I have submitted this issue to the correct repository. If I should open this issue on the `undici` project instead, feel free to close the issue.

### Expected behavior 🤔

The server action should return all set cookies in the response like described in #656

### Steps to reproduce 🕹

I created an example here: https://github.com/GiyoMoon/solid-start-multi-cookie

Steps to reproduce:
```
git clone https://github.com/GiyoMoon/solid-start-multi-cookie
cd solid-start-multi-cookie
pnpm install
pnpm dev
```
Navigate to http://localhost:3000 and call the server action with the `Set cookies` button. If you check the response of the http call you just made, it only contains one `Set-Cookie` header.
If you check the cookie store, it only contains `cookie-2`, but it should also contain `cookie-1`.

If you downgrade to `undici` version `5.18.0`, it works again.

### Context 🔦

I'm trying to set two cookies with different `maxAge` values, and therefore I have to use two separete stores which means two separate cookies.

### Your environment 🌎

```shell
System: Apple M1 Pro v13.4.1
Nodejs: v18.16.1
```

edivados:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

In undici >=5.19.0
Set-Cookie appears multiple times in headers. Server iterates over all set-cookie keys and only the last one will be set. Splitting does nothing here as the cookies are already splitted.

In undici 5.18.0
Set-Cookie appears only once in headers. Server splits the string and setHeader works as expected when passing in an array.

https://github.com/solidjs/solid-start/blob/52d44e2cad33faf3dc40148520d7ae41086dd6a8/packages/start/node/fetch.js#L160-L164

## What is the new behavior?

Now works in both versions.

## Other information

fixes #953

### DIFF
--- a/packages/start/node/fetch.js
+++ b/packages/start/node/fetch.js
@@ -2,7 +2,7 @@ import { once } from "events";
 import multipart from "parse-multipart-data";
 import { splitCookiesString } from "set-cookie-parser";
 import { Readable } from "stream";
-import { File, FormData, Headers, Request as BaseNodeRequest } from "undici";
+import { Request as BaseNodeRequest, File, FormData, Headers } from "undici";
 
 function nodeToWeb(nodeStream) {
   var destroyed = false;
@@ -159,7 +159,7 @@ export async function handleNodeResponse(webRes, res) {
 
   for (const [name, value] of webRes.headers) {
     if (name === "set-cookie") {
-      res.setHeader(name, splitCookiesString(value));
+      res.appendHeader(name, splitCookiesString(value));
     } else res.setHeader(name, value);
   }
 


ryansolid:
Thank you.

PR #960, opened 2023-07-31
Updates to Experimental features and bug fixes

ryansolid:
9 months ago @nksaraf and few others worked on a movies and notes demos to show off new Islands router capability and over the month made huge changes to Solid Start especially around experimental features I haven't been in the position to merge until now. This has delayed being able to merge improvements around these features against `main` since `dev` had moved so much.

Key differences here is that all experimental flags being moved under experimental key on solid-start config, and that instead of the `unstable_islands` this uses `use client` to define Islands. This is still still very experimental but in order to move forward we need to get this back into the trunk.

Hopefully nothing else has broken due to these changes and I've tried my best to vet my merge the last couple weeks but you never know.


ryansolid:
I fixed the types but tests are still broken in everything but node.. This commit is likely the culprit: https://github.com/solidjs/solid-start/commit/848ad00b267e4433aa6b328268a8d31c79ed3b4b

It looks like it only updated node and not other environments.

PR #975, opened 2023-08-09
Set node version for monorepo dev to latest stable (v20)

devinrhode2:
## PR Checklist

> Please check if your PR fulfills the following requirements:
> 
> - [ ] Tests for the changes have been added (for bug fixes / features)

Does this repo have tests that run against node 18 and node 20? I'm assuming it doesn't.

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [x] CI related changes
- [x] infrastructure changes

Not sure what node version we should use when making changes to solid-start's monorepo.

## What is the new behavior?
Node version will be automatically set, if you use asdf-vm.

## Other information

This may bring up a larger discussion:
what node version should everyone be using?

I think 99% of the time, the exact version doesn't matter.

But edge cases are a thing.

Since SolidStart has not reached 1.0 status,
I don't think it makes sense to support node LTS (v18) and latest stable
(i.e. even numbers) is best.

This change is only meant to effect developers working in the solid-start monorepo.


devinrhode2:
If PR is rejected - I do just need to know one thing: What node version should we be using when working in this monorepo?

PR #980, opened 2023-08-09
fix: https with cookies broken, due to appendHeader (#974)

### LINKED ISSUE #974: [Bug?]: https with cookies broken, due to appendHeader
### Duplicates

### Latest version

### Current behavior 😯

When adding a Set-Cookie header while running solid-start with https the following error occur: 

```sh
An unhandled error occured: TypeError: res.appendHeader is not a function
    at handleNodeResponse (file:///Users/admin/repos/app/node_modules/solid-start/node/fetch.js:168:11)
```

### Expected behavior 🤔

I expected the cookie to just be set.

### Steps to reproduce 🕹

Repro here:

https://github.com/birkskyum/solid-start-https-appendHeader-0.3.1

3 changes are made:

* I added a simple API route called test.ts that returns a response with a Set-Cookie header.
* I also added the https config to the vite.config.ts
* I installed the `vite-plugin-mkcert`

### Context 🔦

I want to keep [using https in development](https://auth0.com/blog/using-https-in-your-development-environment/), because I use Auth0. This worked with solid-start up to 0.2.29.

Related 
- https://github.com/solidjs/solid-start/issues/953#issuecomment-1670492120
- https://github.com/solidjs/solid-start/issues/971

Setting https: false make the code run with http, which is how I narrowed it down so far. ([link](https://github.com/solidjs/solid-start/issues/953#issuecomment-1670492120)).

### Your environment 🌎

```shell
macOS arm64
Node 18 and 20. Node 16 breaks even earlier
```

edivados:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

PR #955 breaks using vite with https. When using https [Http2ServerResponse](https://nodejs.org/api/http2.html#class-http2http2serverresponse) is used which does not have appendHeader function and throws.

## What is the new behavior?

Collecting all values returned by splitCookiesString and setting them all with a single setHeader call.
This should fix #974 and still allow adding multiple set-cookie entries to a response as intended in #955.

## Other information

With solid-start 0.3.1 not setting undicis Headers class in globalThis, the Headers behaviour described in #953 is now dependent on the node version. Headers can contain a single set-cookie entry that requires splitting or multiple already splitted entries depending on the node version.

I would like to add a test for adding multiple set-cookie entries to a response but I am not familiar with playwright and have to take a look at it first.


edivados:
Added a test under misc to check if multiple cookies are returned. Feel free to remove if bad.

ryansolid:
Thank you

PR #1020, opened 2023-08-20
fix: router not generating same path for renamed index as for index file

### LINKED ISSUE #1012: [Bug?]: solid-start-static build doesn't work with "(home).tsx" style filenames
### Duplicates

### Latest version

### Current behavior 😯

If I run `solid-start build` while using the `solid-start-static` Vite adapter, the route at `src/routes/index.tsx` produces an output file at `dist/public/index.html` as expected. However, if I change the `src/routes/index.tsx` filename to `src/routes/(home).tsx`, then the `dist/public/index.html` file is no longer produced.

It seems like the `solid-start-static` Vite adapter doesn't properly handle the alternative index filename style (e.g. `(home).tsx`).

As a temporary workaround, I can fix the issue by adding the missing routes to the `prerenderRoutes` option of the Vite plugin:
```
export default defineConfig({
  plugins: [solid({ adapter: staticAdapter(), prerenderRoutes: ["/"] })],
})
```
However, it seems like this shouldn't be required.

### Expected behavior 🤔

I expected the route at `src/routes/(home).tsx` to produce the `dist/public/index.html` file when building using the `solid-start-static` adapter.

### Steps to reproduce 🕹

Steps:

1. `npm init solid@latest`
    * `Where do you want to create the app? › my-app`
    * `Which template do you want to use? › bare`
    * `Server Side Rendering? › n`
    * `Use TypeScript? › y`
2. `cd my-app`
3. `npm add --save-dev solid-start-static`
4. Edit `vite.config.ts` to be:
```
import solid from "solid-start/vite"
import staticAdapter from "solid-start-static"
import { defineConfig } from "vite"

export default defineConfig({
  plugins: [solid({ adapter: staticAdapter() })],
})
```
5. Rename `src/routes/index.tsx` to `src/routes/(home).tsx`
6. `npm run build`
7. Notice that `dist/public/index.html` does not exist.

### Context 🔦

I am trying to build the site for static hosting.

### Your environment 🌎

```shell
OS: macOS Ventura 13.2.1
solid-start: 0.3.3
solid-start-static: 0.3.3
```

edivados:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

The router does not generate the same path for a renamed index as it would for an index file.

Example:

|Index|Id|Path|Index (Renamed)|Id|Path|
|------|--|------|----------|-|----|
|index.tsx|/index|/|(home).tsx|/(home)||
|test/index.tsx|/test/index|`/test/`|test/(test).tsx|/test/(test)|`/test`|

This happens because `toPath` in `path-utils.js` removes every segment looking like a route group from the id which includes renamed index.

This also causes the static adapter to not generate an index.html for routes with a renamed index as it relies on the trailing slash.

## What is the new behavior?

When processing a route file, after creating the route config, check if the route id ends with a renamed index and add a trailing slash to the path.

## Other information

fixes #1012

Tests passed and did a quick manual test. Let's see if it breaks anything else... 🤣

### DIFF
--- a/packages/start/fs-router/router.js
+++ b/packages/start/fs-router/router.js
@@ -146,7 +146,7 @@ export class Router {
       let routeConfig = this.createRouteConfig(path);
 
       // renamed index should have a trailing slash like index files
-      if (!routeConfig.path.endsWith("/") && /\([^)/]+\)$/.test(routeConfig.id)) {
+      if (!routeConfig.path.endsWith("/") && /\/\([^)/]+\)$/.test(routeConfig.id)) {
         routeConfig.path += "/";
       }
 
--- a/packages/start/fs-router/router.js
+++ b/packages/start/fs-router/router.js
@@ -145,6 +145,11 @@ export class Router {
       log("processing", path);
       let routeConfig = this.createRouteConfig(path);
 
+      // renamed index should have a trailing slash like index files
+      if (!routeConfig.path.endsWith("/") && /\([^)/]+\)$/.test(routeConfig.id)) {
+        routeConfig.path += "/";
+      }
+
       if (this.routes[routeConfig.id]) {
         // get old config, we want to compare the oldConfig with the new one to
         // detect changes and restart the vite server


edivados:
Noticed something is off in the repo and I tested this with the bare example within it. Will revisit tomorrow and test again outside the repo.

edivados:
Tested outside the repo with `0.3.3` and everything seems to work. Setting back to `ready for review`.

PR #1026, opened 2023-08-23
Allow passing in `routerProps` to `<StartClient`

### LINKED ISSUE #96: Support router integrations of solid-app-router
example usage: `src/entry-client.tsx`

```tsx
import { render } from "solid-js/web";
import { hashIntegration } from "solid-app-router";
import { StartClient } from "solid-start/entry-client";

render(() => <StartClient routerIntergration={hashIntegration}  />, document.body);
```

example change: `packages/start/entry-client/StartClient.tsx`
```tsx
import { MetaProvider } from "solid-meta";
import { Router } from "solid-app-router";
import { StartProvider } from "../entry-server/StartContext";
import Root from "~/root";

const rootData = Object.values(import.meta.globEager("/src/root.data.(js|ts)"))[0];
const dataFn = rootData ? rootData.default : undefined;

// typeof routerIntergration = RouterIntegration | undefined
// TODO: validate routerIntergration

export default ({ routerIntergration }) => {
  return (
    <StartProvider>
      <MetaProvider>
        <Router source={routerIntergration && routerIntergration()} data={dataFn}>
          <Root />
        </Router>
      </MetaProvider>
    </StartProvider>
  );
};

```
I hardcoded `hashIntegration` from the above solution, the outcome looks good to me

devinrhode2:
Closes #96

When embedding a SolidStart app into a larger javascript app, users may need to customize the router source/integration, to avoid messing with the parent/host apps url. Users could use an in-memory based router, a [search-param based router](https://github.com/solidjs/solid-router/pull/291), hash-based router, etc.

## PR Checklist

There aren't any existing tests for `<StartClient`. I think in the future it would be good to add a "solid-as-a-micro-frontend" example, and we could add a test which ensures that example app renders with [`searchParamIntegration`](https://github.com/solidjs/solid-router/pull/291)

## PR Type

- [x] Feature

Use `patch-package` to pass in routerProps

## What is the new behavior?

Users can customize router source/integration.

### DIFF
--- a/packages/start/entry-client/StartClient.tsx
+++ b/packages/start/entry-client/StartClient.tsx
@@ -16,7 +16,22 @@ function throwClientError(field: string): any {
   );
 }
 
-export default () => {
+export default ({
+  routerProps
+}: {
+  /**
+   * Override props passed to `<Router`.
+   * 
+   * Only `source` is supported today.
+   * 
+   * If you need to pass in other props, please submit a PR.
+   */
+  // Before actually submitting a PR, you can test out the new prop like so:
+  //   import type { RouterProps } from "@solidjs/router";
+  //   const routerProps = { newProp } as RouterProps;
+  //   <StartClient routerProps={routerProps as unknown as undefined} />
+  routerProps?: Pick<RouterProps, 'source'>
+}) => {
   let mockFetchEvent: PageEvent = {
     get request() {
       if (process.env.NODE_ENV === "development") {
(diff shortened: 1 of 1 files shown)


ryansolid:
I think this is fine. Just wondering if similar options need to make it to server. I will merge as is and we will find out I guess.

devinrhode2:
For my use case, my Solid backend is only ever receiving requests for 100% normal SolidStart (path-based) urls.

When the Solid code renders in a browser, it will be search-param based (via [`searchParamIntegration`](https://github.com/solidjs/solid-router/pull/291)).

Here's an example client side (no ssr) call flow:
1. Load Next.js demo w/ url `nextjs-demo.com/asdf/asdf?widgetUrl=${encodeURIComponent("solid-app.com/foo?paramA=a&paramB=b"}`
2. `<FakeFrame href="solid-app.com/foo?paramA=a&paramB=b"`
3. `fetch('solid-app.com/foo?paramA=a&paramB=b'`
4. Inject the html and scripts into the browser
5. SolidStart uses `searchParamIntegration('widgetUrl')` to correctly hydrate

I did get hydration working with this approach

PR #1033, opened 2023-08-28
Kill entire process group spawned by `build` command for `spaClient`

emdede:
## PR Checklist

## PR Type

- [x] Bugfix

Currently, when there is a long-running listener created, such as a WebsocketServer, the port of this listener will be occupied even after `build` has completed. This leads to EADDRINUSE errors on `dev` or `start` after `build`.

Afaict, this behavior is unique to `spaClient` because a shell child_process is being spawned that itself spawns the vite process. The `proc.kill()` command terminates the shell, but the vite subprocess does not receive the signal, which prevents graceful shutdown.

## What is the new behavior?
The child_process is spawned as a detached process, which makes it the leader of a process group. For a shutdown, the whole process group is being killed, which will forward `SIGTERM` to the vite process.

## Other information
I've tested this change locally and it fixes the issue. There could be [side-effects for Windows users](https://nodejs.org/api/child_process.html#optionsdetached:~:text=On%20Windows%2C%20setting%20options.detached%20to%20true%20makes%20it%20possible%20for%20the%20child%20process%20to%20continue%20running%20after%20the%20parent%20exits.%20The%20child%20will%20have%20its%20own%20console%20window.%20Once%20enabled%20for%20a%20child%20process%2C%20it%20cannot%20be%20disabled.) (I am running in WSL2) considering the nature of the `detached` option.

Without the change:
Screenshot 2023-08-28 125304

With the change:
Screenshot 2023-08-28 125446

Edit: as a side note I was also able to "fix" this by changing the `stdio` option. But in this specific case they bring their own bugs with them.
"inherit" will complete the build process but throw an error on the next user input
"pipe" will complete the build process but not exit. After terminating the hanging process, it's fine.

Seeing that both of those options come with their own bugs (and in case of "inherit" it's also a current bug when run with DEBUG flag), I think a way to properly close the process instead seems like the best way to go.


emdede:
Yeah, tests for windows are failing. 8[

emdede:
I've updated the code so the Windows build doesn't error anymore. But the way I did it is to just apply this fix conditionally for non-windows environments.

After seeing the Windows tests failing I have installed everything on my Windows host machine and tried to test the change there. The issue is exactly the same as in Linux, i.e. if merged as is, the problem will remain in Windows.

I'm not sure how to move forward from here.

edivados:
Wild suggestion I am not sure works properly... what if we don't spawn a vite server to generate the index.html and build/use server entry instead? Basically what is done on the astro branch.

- No spawn
- No waitOn

Quick example of what I mean: 
https://github.com/edivados/solid-start/tree/spa-build-mod 

Note: The generated index.html required a bit of cleanup. Not so cool...
- Removing nodes in head

Issue #1008, opened 2023-08-16
[Bug?]: Type checking with @total-typescript/ts-reset fails


lirbank:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

Type checking SolidStart with https://www.npmjs.com/package/@total-typescript/ts-reset fails.

### Expected behavior 🤔

Type check should pass.

### Steps to reproduce 🕹

Steps:

1. `npm init solid@latest`
2. `npm install`
3. `npm install @total-typescript/ts-reset`
4. Create a `reset.d.ts` file in your project with these contents (see https://www.npmjs.com/package/@total-typescript/ts-reset#user-content-get-started):

```ts
// Do not add any other lines of code to this file!
import "@total-typescript/ts-reset";
```
5. `npx tsc --noEmit`

### Context 🔦

The https://www.npmjs.com/package/@total-typescript/ts-reset package has some very useful better default types for some common functions. I use it in all my projects, it would be epic if it worked with SolidStart too.

lirbank:
Related to https://github.com/solidjs/solid-start/issues/976

lirbank:
In the meantime, here is a nice workaround:
https://github.com/solidjs/solid-start/issues/255#issuecomment-1307003993

ryansolid:
In setting up for SolidStart's next Beta Phase built on Nitro and Vinxi we are closing all PRs/Issues that will not be merged due to the system changing. If you feel your issue was closed by mistake. Feel free to re-open it after updating/testing against 0.4.x release. Thank you for your patience.  See https://github.com/solidjs/solid-start/pull/1139 for more details.

PR #1034, opened 2023-08-28
bundle `solid-start/entry-client`

AlexErrant:
## PR Checklist

## PR Type

- [x] Build related changes

Solid-Start is distributed as Typescript, leading to issues like https://github.com/solidjs/solid-start/issues/1008 and https://github.com/solidjs/solid-start/issues/255 This is [not recommended](https://github.com/microsoft/TypeScript/issues/12358#issuecomment-261676923).

## What is the new behavior?
This PR creates a new `solid-start/entry-client` "package", similar to `web`, `store`, `h`, etc. in SolidJS. (Really I just looked at SolidJS and was like ctrl-c ctrl-v.) I only did `entry-client` for now to solicit feedback - I _imagine_ the other folders of start to be more or less the same type of change. If the maintainers approve, I'll do the other dirs too.

This PR introduces breaking changes. Users on `"moduleResolution": "node",` will need to make changes like

```diff
-      import mount from "solid-start/entry-client/mount";
+      import mount from "solid-start/entry-client/dist/mount";
```

I'd argue this is a good thing as it forces Solid (and users) to think about the API, buuuuut I'd get it if that means we need to wait for 0.4 before this PR gets any traction. Also, since I'll be moving nearly every file in `/start` (to a nested `/src` folder), it may introduce conflicts with other PRs/branches. IMO this is a reason to get this done sooner rather than later, but choosing the right time may be important. Sorry if this ruins your timing, but I'm tired of getting ts errors in my node_modules 😭


packages/start/entry-client/package.json
AlexErrant, line 1: cf. https://github.com/solidjs/solid/blob/main/packages/solid/html/package.json

I'm not including `main` cause I believe we require Node18.... I don't know nearly enough about Node to know if that's good enough justification but we're not using a "real" bundler like Rollup. Maybe later?

packages/start/router.d.ts
```diff
@@ -0,0 +1,39 @@
...
+    interface StartRoutes {
+        $: {
+            params: any;
+            data: any;
+        };
+    }
+    interface Route {
+        "/notes/[note]": "/notes/[note]";
+    }
+}
```
AlexErrant, line 35: This file was generated from `Router.tsx`. It's used by the `tsconfig` via `"types": [ ..., "solid-start/router"`, which afaik only accepts `.d.ts` files ([or whole projects](https://www.typescriptlang.org/tsconfig#types)).

In actuality all `entry-client` needs is this here `declare global`, but I'm leaving it in `Router.tsx` for now for easy review. Could easily move it elsewhere if requested!

packages/start/entry-client/package.json
AlexErrant, line 1: cf. https://github.com/solidjs/solid/blob/main/packages/solid/html/package.json

I'm not including `main` because I believe we require Node18.... I don't know nearly enough about Node to know if that's good enough justification but we're not using a "real" bundler like Rollup. Maybe later?

packages/start/entry-client/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "solid-start/entry-client",
+  "module": "./dist/index.jsx",
+  "types": "./dist/index.d.ts",
+  "type": "module",
+  "sideEffects": false
+}
```
AlexErrant, line 7: cf. https://github.com/solidjs/solid/blob/main/packages/solid/html/package.json

I'm not including `main` because I believe we require Node18.... I don't know nearly enough about Node to know if that's good enough justification but we're not using a "real" bundler like Rollup. Maybe later?

ryansolid:
Truthfully.. we're sort of completely changing all of this in 0.4.. like completely changing the entry imports etc.. so I will take this under consideration but no action at the moment.

AlexErrant:
Coolbeans, more than happy to make PRs if/when you get to that point!

PR #1041, opened 2023-09-01
fix: undefined HTTP method handlers not responding with 405

edivados:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

The [docs](https://start.solidjs.com/core-concepts/api-routes#writing-an-api-route
) state the following:
```quote
 If a handler is not defined for a given HTTP method, SolidStart will return a 405 Method Not Allowed response.
```
This is not the case.
- Defined and undefined routes return 200 for undefined HTTP method handlers
- Wildcard route placed at the root of routes directory exporting a POST method handler breaks server functions (i.e. [...404])

## What is the new behavior?

- Defined routes return 405 for undefined HTTP method handlers
- Undefined routes return 200 for undefined HTTP method handlers (same as before)
- inlineServerFunctions middlware gets called before apiRoutes to prevent wildcard routes from mistakenly handle requests for server functions
- In island mode POST method handler of a route is skipped since it is used for navigation.

### 404 Page

[...404] wilcard route should look like this if it's desired to return 404 for all HTTP methods otherwise 405 will be returned.

```tsx
const notFound = () => new Response(null, { status: 404 });

export const POST = notFound;
export const PUT = notFound;
export const PATCH = notFound;
export const DELETE = notFound;

export default function NotFound() {
  ...
}
```


packages/start/fs-router/router.js
```diff
@@ -386,17 +386,20 @@ export function stringifyAPIRoutes(
    * @return {string}
    */
   function _stringifyRoutes(/** @type {(RouteConfig)[]} */ routes) {
+    const methodNotFound = jsFile.addNamedImport("methodNotFound", "solid-start/api/utils");
```
edivados, line 389: Not 100% sure about this. Could also just write `./utils` instead of `solid-start/api/utils`.

packages/start/fs-router/router.js
```diff
@@ -386,17 +386,20 @@ export function stringifyAPIRoutes(
    * @return {string}
    */
   function _stringifyRoutes(/** @type {(RouteConfig)[]} */ routes) {
+    const methodNotAllowed = jsFile.addNamedImport("methodNotAllowed", "solid-start/api/utils");
```
edivados, line 389: Could also just write `./utils` instead of `solid-start/api/utils`. Not sure if one is better than the other.

edivados:
Apperently this breaks createServerAction$

packages/start/entry-server/index.ts
```diff
@@ -21,8 +21,8 @@ export const render = (
   }
 ) =>
   composeMiddleware([
-    apiRoutes,
     inlineServerFunctions,
+    apiRoutes,
```
edivados, line 25: Without this requests for server functions will be mistakenly handeld by wildcard routes (like 404) that now have a handler for all methods.

packages/start/server/render.ts
```diff
@@ -12,8 +12,8 @@ export function renderSync(
     renderId?: string;
   }
 ) {
-  return () => apiRoutes({
-    forward: inlineServerFunctions({
+  return () => inlineServerFunctions({
+    forward: apiRoutes({
```
edivados, line 16: Using `apiRoutes` and `inlineServerFunctions` middleware again here doesn't make much sense to me right now.  Just looking at it I am expecting these two middlewares to be called twice when requesting a page route...

packages/start/fs-router/router.js
```diff
@@ -392,12 +392,15 @@ export function stringifyAPIRoutes(
...
-                v =>
-                  i.apiPath != null &&
-                  `${v}: ${jsFile.addNamedImport(v, path.posix.resolve(i.apiPath[v]))}`
-              ),
-              i.componentPath ? `GET: "skip"` : undefined,
+              ...API_METHODS.map(v => {
+                if (v === "GET" && i.componentPath)
+                  return `${v}: "skip"`;
+                if (v === "POST" && options.islands && i.componentPath)
+                  return `${v}: "skip"`;
```
edivados, line 399: Not sure if to check for `islands`, `ìslandsRouter` or both.

packages/start/server/render.ts
edivados: Fixed in 5e615768f87b2270c5c73a7aab802cbaeb33a036

packages/start/fs-router/router.js
edivados: Checking for islandsRouter as islands can be used without the router.

Dezzj (commented):
Veiwed

ryansolid:
Thank you for all your work on this. However, in setting up for SolidStarts next Beta Phase built on Nitro and Vinxi we are closing all PRs/Issues that will not be merged due to the system changing. If you feel your issue was closed in mistake. Feel free to re-open it after updating/testing against 0.4.x release. Thank you for your patience.  See https://github.com/solidjs/solid-start/pull/1139 for more details.

PR #1057, opened 2023-09-11
fix: using network host errors during logging

emdede:
## PR Checklist

## PR Type

- [x] Bugfix

Providing a network IP as Host option is ok for Vite but errors because of Solid's logging - it only accounts for `local` addresses.
See [Discord](https://discord.com/channels/722131463138705510/910635844119982080/1150662412337426482)

## What is the new behavior?
If no `local` address defined, try `network` address.

## Other information
Should have no runtime implication since these values are only collected for logging.


ryansolid:
Thank you

PR #1059, opened 2023-09-12
feat(bun): Add bun adapter

pan93412:
Experimental but works 🚀

image

## TODOs

- [x] `bun run --bun build`
- [x] `bun run --bun start`
- [x] `bun run --bun dev`
- [ ] Support clientIP (https://github.com/oven-sh/bun/pull/4559)

## Usage

The `--bun` flag is essential; otherwise, this server will be spawned by Node.

```bash
export START_ADAPTER="solid-start-bun"

bun run --bun build
bun run --bun start
```

## Benchmark

wrk command: `wrk -t8 -c500 -d10s --latency "http://localhost:3000"`
system: macOS 13.5.2, on Apple M1.

| Platform | Base RAM usage | Requests/sec | RAM usage after `wrk` |
| --- | --- | --- | --- |
| Node | 22.2MB | 3107.68 | 126.5MB |
| Bun | 16.8MB | 7057.49 | 112.2MB |

PR #1060, opened 2023-09-12
feat(start/node): Don't `Object.assign` on Bun

pan93412:
```
3 | 
4 | // FIXME: Bun does not support this assignment.
5 | // We only assign if the runtime is not Bun.
6 | //
7 | // https://bun.sh/guides/util/detect-bun
8 | Object.assign(globalThis, Streams);
   ^
TypeError: Attempted to assign to readonly property.
      at /private/tmp/solid/test/node_modules/solid-start/node/globals.js:8:0
      at processTicksAndRejections (:1:2602)
```

Ready for #1059

### DIFF
--- a/packages/start/node/globals.js
+++ b/packages/start/node/globals.js
@@ -1,7 +1,14 @@
 import crypto from "crypto";
 import Streams from "stream/web";
 
-Object.assign(globalThis, Streams);
+// Bun does not support this assignment.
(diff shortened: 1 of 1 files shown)


ryansolid:
Yeah no problem.. a lot of these old polyfills will be able to go away so soon.

PR #1061, opened 2023-09-13
fix: api and inlineServerFunctions middleware being called twice on page request

edivados:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

When requesting a page the `api` and `inlineServerFunctions` middleware gets called twice.
This happens because this two middlewares are called again inside the render middleware.

https://github.com/solidjs/solid-start/blob/043577a6400b8461baa2345e70b665f86da9692f/packages/start/entry-server/index.ts#L40

https://github.com/solidjs/solid-start/blob/043577a6400b8461baa2345e70b665f86da9692f/packages/start/server/render.ts#L15-L17

Created by the big merge. 
`server/render.ts` = original (before merge)
`entry-server/index.tsx` = new (after merge)

## What is the new behavior?

On page request the two middlewares get only called once.

## Other information

Decided to keep the new code.

### DIFF
--- a/packages/start/server/render.ts
+++ b/packages/start/server/render.ts
@@ -1,9 +1,7 @@
 import { JSX } from "solid-js";
 import { renderToStream, renderToString, renderToStringAsync } from "solid-js/web";
-import { apiRoutes } from "../api/middleware";
-import { inlineServerFunctions } from "../server/middleware";
 import { redirect } from "../server/responses";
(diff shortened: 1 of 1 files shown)


ryansolid:
Thank you.. I think I noticed this the other day. Basically Nikhil and I both fixed the same issue in a different place. And then we got double.

PR #1082, opened 2023-10-06
Allow . in Netlify edge function paths

iainmerrick:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

When deploying as a Netlify edge function, it explicitly does not handle any paths containing a "." character.

These paths fall through to the default Netlify 404 page (not the project's own 404).

## What is the new behavior?

All requests are handled by the edge function. If the project has its own 404 page, we don't ever see the default Netlify 404.

## Other information

I've manually tested that this covers the following cases:
- static asset (with or without `.` in name)
- dynamic routes (with or without `.` in name)
- internal `fetch` of static asset (from `public`)
- internal `fetch` of dynamic route

If you'd like me to put together a test, let me know. It's not really specific to Netlify so I guess it doesn't belong in this package.

### DIFF
--- a/packages/start-netlify/entry-edge.js
+++ b/packages/start-netlify/entry-edge.js
@@ -1,7 +1,7 @@
 import manifest from "../../netlify/route-manifest.json";
 import handler from "./entry-server.js";
 
-export default (request, context) => {
+export default async (request, context) => {
   const env = {
       ...context,
       manifest,
@@ -9,20 +9,15 @@ export default (request, context) => {
     },
     clientAddress = request.headers.get("x-nf-client-connection-ip");
 
-  function internalFetch(route, init = {}) {
-    if (route.startsWith("http")) {
-      return fetch(route, init);
-    }
+  // First check for static assets, to avoid shadowing them.
+  const asset = await context.next(request.clone());
+  if (asset.status !== 404) {
+    return asset;
+  }
 
-    let url = new URL(route, "http://internal");
-    const request = new Request(url.href, init);
-    return handler({
-      request,
-      clientAddress,
-      locals: {},
-      env,
-      fetch: internalFetch
-    });
+  function internalFetch(route, init = {}) {
+    const url = new URL(route, request.url);
+    return fetch(url, init);
   }
 
   return handler({
@@ -31,5 +26,5 @@ export default (request, context) => {
     locals: {},
     env,
     fetch: internalFetch
-  })
-}
+  });
+};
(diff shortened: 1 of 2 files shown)


iainmerrick:
Now I see what's going on; edge functions shadow static assets, so with this PR as it stands all assets will be shadowed and served as 404s by the edge function instead. I didn't notice this at first because all the assets were cached, doh.

Skipping the edge function for paths containing `.` handles most common cases pretty well, but it blocks a lot of useful legitimate usage (for example, I'd like to implement the GOPROXY protocol which necessarily uses `.`).

iainmerrick:
I've added `"excludedPath": "/assets/*"` to the manifest, and that fixes the Vite assets.

Static files in `public` (e.g. `favicon.ico`) still don't work because they could live anywhere.

iainmerrick:
OK! This PR now covers all the following cases:
- static asset (with or without `.` in name)
- dynamic routes (with or without `.` in name)
- internal `fetch` of static asset (from `public`)
- internal `fetch` of dynamic route

If you'd like me to put together a test, let me know. It's not really specific to Netlify so I guess the test wouldn't belong in this package.

iainmerrick:
I had a quick look at adding a test, but it looks like:
- existing tests aren't yet run against the various adapters (including `start-netlify`);
- tests are currently failing.

ryansolid:
I think the case I'd want to check is top level wildcard routes.. classically having star routes like we do in the Hackernews example break when we don't put some sort of criteria in.

iainmerrick:
> I think the case I'd want to check is top level wildcard routes.. classically having star routes like we do in the Hackernews example break when we don't put some sort of criteria in.

Makes sense... I've deployed the HN sample using my solid-start-netlify branch here: https://github.com/iainmerrick/solid-start-edge-test/pull/1

Looks like it's working correctly. I'll add a few routes to test those cases I listed.

iainmerrick:
Test routes:
- static asset: https://deploy-preview-1--solid-start-edge-test.netlify.app/favicon.ico
- redirect to asset: https://deploy-preview-1--solid-start-edge-test.netlify.app/favicon-redirect.ico
- internal fetch asset: https://deploy-preview-1--solid-start-edge-test.netlify.app/favicon-static-fetch.ico
- fetch another route: https://deploy-preview-1--solid-start-edge-test.netlify.app/favicon-dynamic-fetch.ico

ryansolid:
Seems good to me thanks for testing

PR #1086, opened 2023-10-12
add omitNestedClosingTags to types

birkskyum:
Typescript is complaining about [omitNestedClosingTags](https://github.com/solidjs/solid/blob/main/CHANGELOG.md#smaller-templates) missing in vite config.

This PR might not be needed if https://github.com/solidjs/vite-plugin-solid/pull/122 is merged.

### DIFF
--- a/packages/start/vite/plugin.d.ts
+++ b/packages/start/vite/plugin.d.ts
@@ -20,6 +20,7 @@ export type Options = Omit<import("vite-plugin-solid").Options, "ssr"> & {
   rootEntry: string;
   serverEntry: string;
   clientEntry: string;
+  omitNestedClosingTags: boolean;
   router: import("../fs-router/router").Router;
 };
 


birkskyum:
seems like this will be needed too. Tried installing vite-plufgin-solid 2.7.2 , with solid-start and it wasn't able to find omitNestedClosingTags

edivados:
Hmm it's there with 2.7.2. How are you trying to set it?

```js
import solid from "solid-start/vite";

export default defineConfig({
  plugins: [solid({
    solid: {
      omitNestedClosingTags: true
    }
  })],
});

```

birkskyum:
You're right, it's there under:
```ts
    solid({
      ssr: true,
      solid: {
        omitNestedClosingTags
      }
```

This PR would add it to:

```ts
    solid({
      ssr: true,
      omitNestedClosingTags
```

Turns out I'm a bit confused about the nested  `solid: { solid: {}}` 😅 can't be alone with that

ryansolid:
It is confusing because there is a VIte plugin Solid and a separate SolidStart plugin with a solid option that passes to the Vite Plugin Solid. This weirdness goes away in the next version. I don't know.. Do the options actually pass through. If so I guess I should merge.

birkskyum:
@edivados  Can you check if it takes effect if the omitNestedClosingTags is added to both the outer and inner solid: {} ?

edivados:
The inner one works since it gets added to presets in vite-plugin-solid. The outer one gets passed down to vite-plugin-solid but it does nothing with it so it does not end up in dom-expressions.

birkskyum:
Okay, thanks for checking it out.

ryansolid:
Thanks you both. Ok. I'm going to do the release now.

PR #1094, opened 2023-10-25
New prop serialization step

lxsmnsyc:
# 🚧 WORK IN PROGRESS 🚧

This PR utilizes the new `sharedConfig.context.serialize`. The current prop serialization process is pretty limited and brings a lot of problems, mainly the deduping problem which the new `serialize` method solves. Of course this is all for consistency.

The challenge right now is how to generate the ID for both the `serialize` and the hydration's `renderId`. Currently, hydration step hack its way through by extracting the `renderId` from the `data-hk`. ID must now be known on the server so we can map the serialized props on the client (since the props is no longer in the `solid-island` element), as well as utilize it for `renderId`

### DIFF
--- a/packages/start/islands/index.tsx
+++ b/packages/start/islands/index.tsx
@@ -9,7 +9,6 @@ declare module "solid-js" {
   namespace JSX {
     interface IntrinsicElements {
       "solid-island": {
-        "data-props": string;
         "data-component": string;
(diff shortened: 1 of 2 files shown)


lxsmnsyc:
Moving this to `vinxi` branch

PR #1096, opened 2023-10-26
feat: new serialization

lxsmnsyc:
# 🚧 WORK IN PROGRESS 🚧
- [ ] Changes prop serialization step for islands.
  -  This will utilize `sharedConfig.context.push` to serialize props and generate island ids. This unlocks reference deduping across the entire page, as well as make it so that the serialization mechanism is consistent to that of Solid.
  - Since the props are no longer inserted as a property to the island, we use `sharedConfig.load` to load the props data into the island during hydration.
- [x] Use `seroval` for server functions
  - Low priority. Intent is for server functions to utilize seroval for serialization rather than using JSON. This allows server functions to be consistent with data serialization.

### DIFF
--- a/packages/start/islands/index.tsx
+++ b/packages/start/islands/index.tsx
@@ -2,14 +2,13 @@ import { Component, ComponentProps, lazy, sharedConfig } from "solid-js";
 import { Hydration, NoHydration } from "solid-js/web";
 import { useRequest } from "../server/ServerContext";
 import { IslandManifest } from "../server/types";
-import { splitProps } from "./utils";
 export { default as clientOnly } from "./clientOnly";
 
 declare module "solid-js" {
   namespace JSX {
     interface IntrinsicElements {
       "solid-island": {
-        "data-props": string;
+        "data-id": string;
         "data-component": string;
         "data-island": string;
         "data-when": "idle" | "load";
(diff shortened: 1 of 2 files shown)


packages/start/islands/index.tsx
```diff
@@ -64,42 +60,31 @@ export function island<T extends Component<any>>(
...
-      };
-
       // @ts-expect-error
       if (!sharedConfig.context?.noHydrate) {
-        return <Component {...compProps} />;
+        return <Component {...props} />;
       }
+      
+      // TODO
+      const islandID = sharedConfig.context.push(props)
```
lxsmnsyc, line 69: Some big questions:
- What happens to JSX?

packages/start/islands/mount.tsx
```diff
@@ -18,32 +19,28 @@ export function hydrateServerRouter() {
...
-      get children() {
-        const p = el.getElementsByTagName("solid-children");
-        getHydrationKey();
-        [...p].forEach(a => {
-          (a as any).__$owner = getOwner();
-        });
-        return;
-      }
+      ...sharedConfig.load(id),
+      // get children() {
```
lxsmnsyc, line 28: honestly don't know what to do with this one lol

PR #1101, opened 2023-10-29
add support for @solidjs/meta ^0.29.0 peer dep

rpivo:
solid-start 0.3.8 does not allow for @solidjs/meta ^0.29.0 peer dep.

### DIFF
--- a/packages/start/package.json
+++ b/packages/start/package.json
@@ -142,7 +142,7 @@
     "vitest": "^0.20.3"
   },
   "peerDependencies": {
-    "@solidjs/meta": "^0.28.0",
+    "@solidjs/meta": "^0.28.0  || ^0.29.0",
     "@solidjs/router": "^0.8.2",
     "solid-js": "^1.6.2",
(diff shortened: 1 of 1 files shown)


packages/start/package.json
```diff
@@ -142,7 +142,7 @@
     "vitest": "^0.20.3"
   },
   "peerDependencies": {
-    "@solidjs/meta": "^0.28.0",
+    "@solidjs/meta": "^0.28.0  || ^0.29.0",
```
rpivo, line 145:
```suggestion
    "@solidjs/meta": "^0.29.0",
```

Dezzj (commented):
Veiwed

ryansolid:
v0.29.x actually doesn't work with SolidStart currently. I can update SolidStart but then everyone would need to 0.29

ryansolid:
I did.. see v0.3.9.. keep in mind it needs Solid 1.8.4 or later.

Issue #848, opened 2023-04-19
Document type-safe locals usage


parisholley:
An update was made to support a `locals` object that can be populated in middlewhere but it isn't clear what the best practice is to utilize it.

https://github.com/solidjs/solid-start/issues/642

For my project I have a custom types definition so that I can change the `Record` into a known type like so: 

```typescript
declare module 'solid-start' {
  interface ServerFunctionEvent {
    locals: MyCustomType;
  }
}
```

To consume locals in a typesafe way, I have to do the following:

```typescript
server$(async function (this: ServerFunctionEvent) {
  this.locals.myCustomFunction();
});

createServerAction$(async function (this: ServerFunctionEvent) {
  this.locals.myCustomFunction();
});
```

If using an API that makes `FetchEvent` available as an argument, that can be used as well but its availability is not consistent (eg: cannot access via an argument with `server$`).

Is the usage of `this` a blessed approach? If so, is there a way to update the API interfaces to enforce the type of `this` so it does not have to be explicitly defined?

austins:
+1. This is a very powerful feature. An example of how to configure a custom types file to type them, and where to assign locals would be good.

ryansolid:
Good idea.. we should document how to extend locals on FetchEvent

ryansolid:
Documented now https://start.solidjs.com/api/RequestEventLocals

PR #1114, opened 2023-11-05
feat: Add LocalsExt interface

NathanHuisman:
## PR Type

- [x] Feature

I found myself struggling to extend FetchEvent.locals globally, because typescript does not allow declaration merging of non-method members of interfaces (see https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces)

## What is the new behavior?
We can improve this by adding a LocalsExt interface that can be merged in your own app.d.ts like this:

```typescript
import type { LocalsExt } from "solid-start/server";

declare module "solid-start/server" {
  interface LocalsExt {
    auth?: AuthUser;
  }
}
```

## Other information

Issue #848 

I saw Lucia doing [something similar](https://lucia-auth.com/getting-started/solidstart/) with a namespace


ryansolid:
Ok thank you.

PR #1116, opened 2023-11-07
[vinxi] set content type header on page response

edivados:
## PR Type

- [x] Bugfix

On GitHub Codespaces page responses return plain text.

## What is the new behavior?

Set `Content-Type` header in `createPageEvent()`.

## Other information

- Did not add it to `server/spa/handler.tsx` because its never used
- Apperently when returning a string the header is set automatically


ryansolid:
Thanks... I have refactored stuff since so I'm not going to merge the PR directly, but I will apply your fix.

PR #1093, opened 2023-10-24
Update examples

birkskyum:
I updated all the examples dependencies to their latest versions, still within Node 18. This is a preparation step towards Node 20 which now is the current LTS.

I identified that the tRPC example is breaking (and using [archived](https://github.com/OrJDev/solid-trpc) packages), so I didn't touch that. Reported it here instead
- #1092 

I identified that the following examples will be challenging when moving to Node 20, because `start-cloudflare-worker` isn't compatible with Node 20 atm:
- `notes`
- `with-websocket`

For the authjs example, [@solid-auth/base](https://github.com/orjdev/solid-auth) is archived and moved to @solid-mediakit/auth, so I updated the dependencies to reflect that. Might help with: https://github.com/solidjs/solid-start/issues/1007

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [X] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] infrastructure changes
- [ ] Other... Please describe:

PR #1121, opened 2023-11-13
Update examples that only need trivial changes

birkskyum:
Remake of part of #1093 , considering examples that only need updates to deps, but not code changes.

I noticed that the Notes app both before/after this update has a bug. When clicking "Edit" on a note, the editor doesn't receive the **body** of the note, but instead has an empty text field.

## PR Type

- [ ] Bugfix
- [ ] Feature
- [X] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

### DIFF
--- a/examples/bare/package.json
+++ b/examples/bare/package.json
@@ -8,15 +8,15 @@
   "type": "module",
   "devDependencies": {
     "@types/node": "^18.17.5",
-    "esbuild": "^0.14.54",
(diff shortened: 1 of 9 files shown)


birkskyum:
makes more sense to look into this after the solidjd/start v.0.4 release

PR #1139, opened 2023-12-08
SolidStart Beta 2

ryansolid:
This PR follows the work on our next stage of beta for SolidStart. This will be a large change.

The whole current adapter ecosystem goes away as well as including libraries like `@solidjs/router` and `@solidjs/meta` as built-ins. Additionally as you may have noticed those libraries have changed their APIs over the past couple months so suffice to say patterns around data loading have also changed.

The content of this PR is largely covered by the RFC I posted a few months back: https://github.com/solidjs/solid-start/discussions/1052

I think the biggest thing to be aware of with such a change is pretty much all of the existing issues/PRs no longer apply or should be taken somewhere else. From a code standpoint we shrunk the project to ~1400 lines of code compared to its previous ~10,800. So as you can imagine it just isn't the same project anymore internally.

I will be closing out all the issues and PRs after this release. It is also unlikely given our resources to continue fixing bugs on older versions. I realize this may be a severe inconvenience to many of you, and I apologize ahead. This is beta software and we just don't have the resources to maintain earlier beta versions. Ultimately the lack of resources was what influenced this refactor in the first place as we couldn't keep on top of the current issues as they were. So going back makes even less sense.

Thank you all for your understanding and patience as we boldly move forward with the future of Solid.


packages/start/config/index.js
```diff
@@ -0,0 +1,179 @@
...
+      {
+        dir: resolve.absolute(config.dir, router.root),
+        extensions: config.extensions ?? ["js", "jsx", "ts", "tsx"]
+      },
+      router,
+      app
+    );
+}
+
+export function defineConfig(baseConfig = {}) {
```
brenelz, line 39: Thoughts on converting this file to typescript and doing something like this? This would allow the config to be typed

```
import { InlineConfig } from "vite";

type SolidStartInlineConfig = Omit<InlineConfig, "router"> & {
  start?: {
    ssr?: boolean,
    extensions?: string[],
    server?: any,
    appRoot?: string,
    middleware?: string
    islands?: boolean
  }
}

export function defineConfig(baseConfig: SolidStartInlineConfig = {}) {
```

docs/routes/api/dev.md
```diff
@@ -8,7 +8,7 @@ active: true
 
 # solid-start dev
 
-##### `solid-start dev` starts a development server backed by [Vite](https://vitejs.dev/).
+##### `solid-start dev` starts a development server backed by [Vinxi](https://vinxi.vercel.app/).
```
birkskyum, line 11: I'm a bit confused about this - all the updated examples used `vinxi dev` not `solid-start dev`. Same for `vinxi build` and `vinxi start`

ryansolid: You are right. There is no solid-start commands after thus

birkskyum: okay - i've made a small pr here to clarify that https://github.com/solidjs/solid-start/pull/1145

packages/start/config/index.js
ryansolid: If it works. I'm unclear what Vite transforms in these scenarios. Whether it is part of the build or part of the runtime. We might be better to JSDocs it etc..

birkskyum:
if possible, it would be quite nice to have the vite 5 bump be part of this shift, but would vinxi need an update first to unlock that?

ryansolid:
> if possible, it would be quite nice to have the vite 5 bump be part of this shift, but would vinxi need an update first to unlock that?

Yes. We don't control Vite version directly anymore. Updated solid plugin to the latest. But Vite is coming from Vinxi.

birkskyum:
there are some lingering references to old adapter in i.e. the create-fixture.ts

```ts
import { createServer } from "solid-start-node/server.js";
import "solid-start/node/globals.js";
import type { FetchEvent } from "solid-start/server/types.js";
```

ryansolid:
> there are some lingering references to old adapter in i.e. the test/helpers/create-fixture.ts
> 
> ```ts
> import { createServer } from "solid-start-node/server.js";
> import "solid-start/node/globals.js";
> import type { FetchEvent } from "solid-start/server/types.js";
> ```

Yeah.. tests don't work. I actually don't have a testing solution yet for the repo but I didn't want to delete this incase I could reuse it. Maybe need some better naming to indicate.

JesseVelden:
I was already trying out the new @solid/start, and saw an issue with calling a promise in a middleware function. It specifically goes wrong with `/_server` calls. The request get stuck and shows (pending) as status in Chrome's Network Inspector. When not using promises, it works fine.

I made a CodeSandbox for the issue: https://codesandbox.io/p/devbox/solid-start-middleware-for-server-with-promise-not-working-4lpw65?file=%2Fsrc%2Fmiddleware.ts%3A12%2C7 (See the behavior if you click the Generate New button, with and without the promise inside the src/middleware.ts function)

lxsmnsyc:
@JesseVelden interesting, it works for `Promise.resolve` and for `new Promise`, but not when `new Promise` is called with `setTimeout`.

JesseVelden:
> @JesseVelden interesting, it works for `Promise.resolve` and for `new Promise`, but not when `new Promise` is called with `setTimeout`.

Yea I noticed that too. In my real world example I use the `useSession` from H3/ @solidjs/start/server, but that hangs as well unfortunately. Here is another sandbox to show that: https://codesandbox.io/p/devbox/solid-start-middleware-for-server-with-promise-session-not-working-6kz3tm?file=%2Fsrc%2Fmiddleware.ts%3A31%2C57

PR #1146, opened 2023-12-19
Add config types

brenelz:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [X] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

When you use vite.config.ts you don't get typings

## What is the new behavior?
Now when using typescript you get proper typings

## Other information
image

image

### DIFF
--- a/packages/start/config/index.d.ts
+++ b/packages/start/config/index.d.ts
@@ -1,0 +1,14 @@
+import { InlineConfig } from "vite";
+
+type SolidStartInlineConfig = Omit<InlineConfig, "router"> & {
+    start?: {
+        ssr?: boolean,
+        extensions?: string[],
+        server?: any,
+        appRoot?: string,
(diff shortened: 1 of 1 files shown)


packages/start/config/index.d.ts
```diff
@@ -0,0 +1,14 @@
+import { InlineConfig } from "vite";
+
+type SolidStartInlineConfig = Omit<InlineConfig, "router"> & {
+    start?: {
+        ssr?: boolean,
+        extensions?: string[],
+        server?: any,
```
brenelz, line 7: if vinxi exposed the `NitroConfig` type this could be used instead of any

ryansolid: I see.. well I can mention this to @nksaraf

PR #1189, opened 2023-12-27
Fix for formData not being parsable in worker environments

fezproof:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

Currently form based actions fail to submit in worker based environments (such as cloudflare or vercel edge). It seems cloudflare thinks these requests will never finish and kills the worker with a 500.
This is due to an issue with h3 and how it parses form data described [here](https://github.com/unjs/nitro/issues/1721)

## What is the new behavior?

This is a short term fix to unblock solid start and allow form based actions to work on these environments. Functionally it is the same and it uses a fix that was proposed in [this comment](https://github.com/unjs/nitro/issues/1721#issuecomment-1819215122)

## Other information

Here is a worker deployed with the issue unfixed.
https://worker-solid.fezproof.workers.dev

You can see how the trigger (on click event calling an action directly) works fine, but the form based submit fails.
Code for this is [here](https://github.com/fezproof/workers-solid)


packages/start/config/server-handler.js
```diff
@@ -89,7 +89,17 @@ async function handleServerFunction(event) {
     contentType.startsWith("multipart/form-data") ||
     contentType.startsWith("application/x-www-form-urlencoded")
   ) {
-    parsed.push(await readFormData(event));
+    // Temporary workaround until https://github.com/unjs/nitro/issues/1721 is resolved
+    // parsed.push(await readFormData(event));
+
+    const request = new Request(getRequestURL(event), {
+      // @ts-ignore Undici option
+      duplex: "half",
```
lxsmnsyc, line 97: This is required?

fezproof: I believe so yeah. It is taken from the h3 code [here](https://github.com/unjs/h3/blob/53703dc860f1ff6fe7ce71d543deff1cfa810b11/src/utils/request.ts#L181C22-L181C22) and thought it best to leave it in.

fezproof: Seems to be related to this line in the [undici code](https://github.com/nodejs/undici/blob/6a04edcd5867c4de6bd53d10f6b44eaecf02755e/lib/fetch/request.js#L505)

fezproof: Actually testing it on a node server it seems fine, will remove it.

lxsmnsyc (approved):

fezproof:
Is there anything more I have to do to get this merged @lxsmnsyc ?

lxsmnsyc:
@fezproof we can wait for @nksaraf

nksaraf:
Should we special case the reading of the formData for the environment its needed in. Because this might be less performant for cases where big files are uploaded. You might not want to eagerly evaluate the whole thing. So a little careful consideration is needed. And the fix should probably be in `nitro` or `h3` for this kind of thing. @pi0 does it make sense to push this upstream?

AirBorne04:
still see this issue in the latest release. I will follow up on nitro to get it fixed though

PR #1196, opened 2023-12-29
feat: dev overlay

lxsmnsyc:
This PR introduces the new Dev Overlay which intends to replace the current ErrorBoundary being shipped.

The Dev Overlay in this PR provides better Error UI. In the future, we might be able to expand some more dev-related features:
- Sticky Dynamic Island/Tool Bar
- HMR indicator (just a good ol spinner)
- Routes Inspection (theoretically possible)
- Server Function Inspection (this is a bit tricky but theoretically should be possible)
- Server-Client Hydration Mismatch diff comparison (same possibility)
- Hydration Data Inspection

and possibly more, but for now, the Error UI would be our starting point.

image


nksaraf:
Do you think we should push this upstream to `vinxi`? Then everyone gets the error screen. We can update the default Vite Error Overlay.

lxsmnsyc:
@nksaraf The thing is at some point this is going to be personalized for SolidStart so I'm not sure. Universal error overlay is a good idea though.

nksaraf:
Yeah I can take learning's from this and apply them there while you take the SolidStart specific one further

nksaraf:
Two issues remaining I think:

-  [ ] tried to do a builld and looks like the DevOverlay is there in the build. Needs some more treeshaking magic I think
-  [ ] I think using all of shiki languages is causing some unnecessary slowdowns, so if its has a way of doing it without all the languages. maybe using https://github.com/antfu/shikiji

lxsmnsyc:
I'm actually currently using shikiji.

For the treeshaking, I'll check again

ryansolid:
Yeah not sure why but treeshaking doesn't appear to be working.

ryansolid:
Looks good now.. thanks @nksaraf.

PR #1200, opened 2023-12-30
Allow customizing Vite plugins per router

XiNiHa:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

Users only can apply the same kind of Vite plugins to all Vinxi routers

## What is the new behavior?
Users can customize which Vite plugins to apply per Vinxi router

### DIFF
--- a/packages/start/config/index.d.ts
+++ b/packages/start/config/index.d.ts
@@ -9,7 +9,14 @@ type SolidStartInlineConfig = Omit<InlineConfig, "router"> & {
         appRoot?: string,
         middleware?: string,
         islands?: boolean
+    },
+    plugins?: Plugins | {
+        client?: Plugins,
+        server?: Plugins,
+        serverFunctions?: Plugins
     }
 }
+
+type Plugins = InlineConfig["plugins"] | (() => Promise<InlineConfig["plugins"]> | InlineConfig["plugins"])
 
 export declare function defineConfig(baseConfig?: SolidStartInlineConfig)
--- a/packages/start/config/index.js
+++ b/packages/start/config/index.js
@@ -57,6 +57,19 @@ export function defineConfig(baseConfig = {}) {
     entryExtension = ".jsx";
   }
 
+  const { clientPlugins, serverPlugins, serverFunctionsPlugins } =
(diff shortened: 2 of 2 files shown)


nksaraf:
I like this and I think level of deetail about the infra is good to expose and will make people understand the system better. Because I don't think they can work well in it without understanding it. They need to know  the layers between vite and them. That's the public API that they are building against.

packages/start/config/index.d.ts
```diff
@@ -9,7 +9,14 @@ type SolidStartInlineConfig = Omit<InlineConfig, "router"> & {
...
         islands?: boolean
+    },
+    plugins?: Plugins | {
+        client?: Plugins,
+        server?: Plugins,
+        serverFunctions?: Plugins
     }
 }
 
+type Plugins = InlineConfig["plugins"] | (() => Promise<InlineConfig["plugins"]> | InlineConfig["plugins"])
```
brenelz, line 20: you have `InlineConfig["plugins"]` twice?

XiNiHa: That's interpreted as `T | (() => (Promise<T> | T))`

packages/start/config/index.js
```diff
@@ -57,6 +57,19 @@ export function defineConfig(baseConfig = {}) {
     entryExtension = ".jsx";
   }
 
+  const { clientPlugins, serverPlugins, serverFunctionsPlugins } =
+    typeof plugins === 'object' && !Array.isArray(plugins)
```
nksaraf, line 61: should check if plugins is a function, if it is treat as the same as if it was an array

XiNiHa: `typeof plugins` will return `"function"` if it is.

ryansolid:
Hmm.. this does ask a question though. I'm starting to wonder if trying to wrap this into a Vite config was too ambitious. There are certain fields we can't support and things we need to expose from underlying layer.

ryansolid:
Thanks for the PR but there is a new way to do this in 0.6.0

you can return a different vite config per router .. it is passed in as an argument to vite option.

PR #1204, opened 2024-01-01
Set HttpStatusCode 500 in ErrorBoundary

frenzzy:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

In case of errors response status code is 200 OK

## What is the new behavior?
In case of errors response status code is 500 Internal Server Error

### DIFF
--- a/packages/start/shared/ErrorBoundary.tsx
+++ b/packages/start/shared/ErrorBoundary.tsx
@@ -18,37 +18,35 @@ function ErrorMessage(props: { error: any }) {
   createEffect(() => console.error(props.error));
 
   return (
-    <div style={{ padding: '16px' }}>
+    <div style="padding:16px">
       <HttpStatusCode code={500} />
       <div
-        style={{
-          'background-color': 'rgb(252,165,165)',
-          color: 'rgb(153,27,27)',
-          'border-radius': '5px',
-          overflow: 'scroll',
-          padding: '16px',
-          'margin-bottom': '8px',
-        }}
+        style={
+          'background-color:#fca5a5;' +
+          'color:#991b1b;' +
+          'border-radius:5px;' +
+          'overflow:scroll;' +
+          'padding:16px;' +
+          'margin-bottom:8px'
+        }
       >
-        <p style={{ 'font-weight': 'bold' }} id="error-message">
+        <p style="font-weight:bold" id="error-message">
           {props.error.message}
         </p>
         <button
           id="reset-errors"
           onClick={resetErrorBoundaries}
-          style={{
-            color: 'rgb(252,165,165)',
-            'background-color': 'rgb(153,27,27)',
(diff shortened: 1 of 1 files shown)


packages/start/shared/ErrorBoundary.tsx
```diff
@@ -17,31 +18,37 @@ function ErrorMessage(props: { error: any }) {
...
-          style={`color: rgba(252, 165, 165);
-            background-color: rgb(153, 27, 27);
-            border-radius: 5px;
-            padding: 4px 8px`}
+          style={{
+            color: 'rgba(252, 165, 165)',
+            'background-color': 'rgb(153, 27, 27)',
+            'border-radius': '5px',
+            padding: '4px 8px',
+            'margin-top': '8px',
```
frenzzy, line 44: With some space between error message and retry button looks prettier

lxsmnsyc:
I cannot agree to this because some errors can be intentional, and some unintentional. Not sure what others think.

frenzzy:
Yes, but for intentional errors, you'll likely create a custom `ErrorBoundary` to suit your specific needs. For instance, you might not want to display the stack trace for these errors, and you'll want the error message to match the style of your site.

Alternatively, we could add support for customizing the status code. For example:
```diff
- <HttpStatusCode code={500} />
+ <HttpStatusCode code={props.error.statusCode ?? 500} />
```
However, I believe that defaulting to a 500 status code in error scenarios is generally better, as it's a widely accepted standard in web development.

packages/start/shared/ErrorBoundary.tsx
ryansolid: I actually changed all these styles to strings intentionally to reduce compiled size output.

frenzzy: Changed back to string.
I just thought styles are compiled to string during the build time...

nksaraf:
I think it's reasonable the the default ErrorBoundary has this as the default. Its more web friendly and better overall for keeping up the semantics. I do agree that it should be trivial to override with your own ErrorBoundary (which people should be encouraged to know anyway)

ryansolid:
I think the intent here is now captured with the new dev overlay and default errorboundary. So while I won't be merging this. It is getting in.

frenzzy:
Am I correct in understanding that by default, SolidStrat will not alter response status codes in the case of errors, since the primary audience for the default ErrorBoundary is developers, and it is designed to be most visible to them? In other words, are developers always expected to create their own ErrorBoundary to override the default behavior?

Would it be useful if the default ErrorBoundary displayed a message to suggest this practice?

Also, is there a possibility to tree shake away the default ErrorBoundary when a custom one is used?

PR #1211, opened 2024-01-03
Add dot configuration option to allow glob to access the .data direct…

u1in:
Add "dot" configuration option to allow glob to access the .data directories in example/todomvc and example/with-auth projects. Otherwise, example/todomvc will throw an error during startup because there is no .data document.

Or 

Change fsLiteDriver.base in example/todomvc and example/with-auth to "./data"


nksaraf:
Actually we should be picking up dotfiles anyway. What if there's something important there.

u1in:
> Actually we should be picking up dotfiles anyway. What if there's something important there.

I agree. I prefer to change ./.data to ./data.

ryansolid:
Although this CLI will be deprecated pretty soon. I can't actually build the CLI locally anymore.

u1in:
> Although this CLI will be deprecated pretty soon. We will be replacing it with the Solid Community CLI (under the same command). I can't actually build the CLI locally anymore.

🎉The new CLI? That's Awesome!

PR #1218, opened 2024-01-04
fix(docs): wrong highlight width in code blocks

femincan:
## PR Type

- [ x ] Other... Please describe: A style change in the documentation website.

## Changes
In PR #1183 I fixed the problem with highlighted lines that are longer than the code block and don't fill the entire line. But mistakenly, I broke the highlighted lines with a smaller width than the code block. Now these lines don't fill the code block. This PR fixes this and now both look okay.

image

## What is the new behavior?

image
image


femincan:
**Additional Note:** Currently docs can't be built because of an error from `vinxi`. Updating `vinxi` to `0.1.0` version fixed the errors in my local `solid-start` fork.

ryansolid:
Thank you.

PR #1241, opened 2024-01-11
encode path components

rvlzzr:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

- Create a file or directory within your routes containing characters that browsers encode, for example src/routes/漢字.tsx
- Browse /漢字
- Not found

Need to rename the file to %25E6%25BC%25A2%25E5%25AD%2597.tsx. This isn't documented anywhere and shouldn't be necessary. File and directory names like this are too unwieldy to use.

## What is the new behavior?
- Create a src/routes/漢字.tsx
- Browse /漢字
- Page loads as expected

### DIFF
--- a/packages/start/config/fs-router.js
+++ b/packages/start/config/fs-router.js
@@ -14,7 +14,15 @@ export class SolidStartClientFileRouter extends BaseFileSystemRouter {
           return `:${m.slice(1, -1)}?`;
         }
         return `:${m}`;
-      });
+      })
+      .split('/')
+      .map(s => {
+        if (s.startsWith(':') || s.startsWith('*')) {
+          return s.charAt(0) + encodeURIComponent(s.slice(1));
+        }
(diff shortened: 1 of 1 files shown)


ryansolid:
This seems reasonable. I was thinking maybe we should be handling this at the router level rather than the file system routing level, as those paths break the router even without Start. I'd love to know if other routers handle it there.

rvlzzr:
https://github.com/solidjs/solid-router/pull/354 fixes it on the router end if this is preferred.

ryansolid:
Yeah I'm certain the router is the better place to handle this. I'm not even sure we should be handling this in the config in but rather better handling this when processing the input from the URL or the matching. I've updated the router and even if that fix isn't perfect I think that is where the fixes of this nature belong.

PR #1244, opened 2024-01-11
Support large values in deserializeStream

### LINKED ISSUE #1243: [Bug?]: Seroval deserializeStream fails for large values
### Duplicates

### Latest version

### Current behavior 😯

I have a solidstart app with two routes which each load 1-3 MB of nested JSON data in a server function. This data used to be serialized with `JSON.stringify` but since 0.4.0 has been replaced with seroval.

When I quickly navigate repeatedly between these two routes (and sometimes when preloading data onmousehover), sometimes I will encounter the following error
image
browsing up the stacktrace
image
the error originates inside code [introduced by a recent bugfix](https://github.com/solidjs/solid-start/pull/1131) to `deserializeStream`

 * When I print the current `split` that is deserialized (https://github.com/solidjs/solid-start/blob/main/packages/start/config/server-runtime.jsx#L46), I notice that it is an incomplete JS expression. This suggests the large value being serialized is somehow truncated by the `ReadableStream` before a newline is inserted.
 * Where this truncation occurs changes across trials.
 * Looking at network traffic, it appears the entire payload makes it across the wire. The bug is probably in solidstart's `deserializeStream`

### Expected behavior 🤔

I expect to be able to navigate between these routes without issue. This was the case prior to 0.4.x, and expected behavior is restored when I `pnpm link` a fork which removes seroval deserialization on server functions (https://github.com/solidjs/solid-start/compare/main...feynmanliang:solid-start:seroval-error-repro)

### Steps to reproduce 🕹

Steps: Issue occurs in proprietary app, I will create a stackblitz with minimal repro

1. Create two routes which have a `load` that serdes a very long quoted JS string from a server function
2. Navigate between the two routes rapidly
3. Sometimes, the `split` processed by `deserializeStream` is truncated and the closing quote is missing, leading to an error

### Context 🔦

_No response_

### Your environment 🌎

_No response_

feynmanliang:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

deserializeStream may fail due to truncated large values that do not fit in a single ReadableStream chunk

## What is the new behavior?

The stream is exhausted first prior to attempting deserialization

## Other information

Fixes #1243

### DIFF
--- a/packages/start/config/server-runtime.jsx
+++ b/packages/start/config/server-runtime.jsx
@@ -33,11 +33,12 @@ async function deserializeStream(id, response) {
     }
   }
 
-  const result = await reader.read();
-  if (result.done) {
-    throw new Error("Unexpected end of body");
+  let serialized = ""
+  while (true) {
+    const line = await reader.read();
+    if (line.done) break;
+    serialized += new TextDecoder().decode(line.value);
   }
-  const serialized = new TextDecoder().decode(result.value);
   let pending = true;
   let revived;
   const splits = serialized.split("\n");
@@ -108,7 +109,7 @@ async function fetchServerFunction(base, id, args) {
   let result;
   if (contentType && contentType.startsWith("text/plain")) {
     result = await response.text();
-  } else if(contentType && contentType.startsWith("application/json")) {
+  } else if (contentType && contentType.startsWith("application/json")) {
     result = await response.json();
   } else {
     result = deserializeStream(instance, response);


ryansolid:
Thanks

ryansolid:
Actually looking at this closer I need to flag @lxsmnsyc. His pop method looks like it was designed intentionally to do this same iteration, so this "fix" might impact his streaming approach.

lxsmnsyc:
Hi! Thanks for contributing.

Unfortunately, this seems to turn the server functions into a blocking server function. The new lines added makes sure to await the entire request before passing the initial serialized value, whereas the current design involves returning the first serialized value, and deserializing the latter streamed values.

@ryansolid we should revert this.

packages/start/config/server-runtime.jsx
```diff
@@ -33,11 +33,12 @@ async function deserializeStream(id, response) {
     }
   }
 
-  const result = await reader.read();
-  if (result.done) {
-    throw new Error("Unexpected end of body");
+  let serialized = ""
+  while (true) {
+    const line = await reader.read();
+    if (line.done) break;
```
lxsmnsyc, line 39: This is the problematic line. It's awaiting until the stream is done. Server functions are supposed to be non-blocking, and this entire mechanism is already implemented in the `pop` function declared just a few lines before this.

PR #1245, opened 2024-01-12
fix: Add `ChunkReader`

### LINKED ISSUE #1243: [Bug?]: Seroval deserializeStream fails for large values

lxsmnsyc:
- This PR adds the `ChunkReader` utility class.
  - The purpose of this class is for proper consumption of chunks that are being streamed by the server functions request. Previously, we only deserialize the first valid seroval chunk with the assumption that the first response chunk is a valid seroval chunk. However the browser interprets it differently, that is, if the chunk size is too large, the browser will split the "valid chunk" into smaller chunks, leading into a deserialization error.
  - `ChunkReader` ensures that the first chunk is a valid chunk. If it's invalid, it will attempt to read a new chunk, and check if part of the chunk is a valid seroval chunk. This process repeats until the stream is drained.
- Fixes #1243 

Test app:

```js
// @refresh reload
import { createEffect, createResource, createSignal } from "solid-js";
import "./app.css";

export default function App() {
  const [count, setCount] = createSignal(0);

  const [data] = createResource(count, async (value) => {
    "use server";

    return {
      data: new Promise((resolve) => {
        const giantData = [];
        for (let i = 0; i < 1000000; i++) {
          giantData.push(`index: ${i} value: ${value}`);
        }
        setTimeout(resolve, 1000, giantData);
      }),
    };
  });

  createEffect(() => {
    console.log(data()?.data);
  });

  return (
    <main>
      <h1>Hello world!</h1>
      <button class="increment" onClick={() => setCount(count() + 1)}>
        Clicks: {count()}
      </button>
      <p>
        Visit{" "}
        <a href="https://start.solidjs.com" target="_blank">
          start.solidjs.com
        </a>{" "}
        to learn how to build SolidStart apps.
      </p>
    </main>
  );
}
```

> [!NOTE]
> In my test, the maximum chunk size is around 2097152 bytes. Tested in latest Arc (Chromium browser) on Mac OS

### DIFF
--- a/packages/start/config/server-runtime.jsx
+++ b/packages/start/config/server-runtime.jsx
@@ -13,55 +13,80 @@ import {
 } from 'seroval-plugins/web';
 import { createIslandReference } from "../server/islands";
 
+class ChunkReader {
+  constructor(stream) {
+    this.reader = stream.getReader();
+    this.buffer = '';
+  }
+
+  async next() {
+    const [first, ...rest] = this.buffer.split('\n');
+    // Check if there's a chunk
+    if (first === '') {
+      // if there's no chunk, read again
+      const chunk = await this.reader.read();
+      console.log(chunk);
+      if (chunk.done) {
+        return { done: true, value: undefined };
+      }
+      // repopulate the buffer
+      this.buffer += new TextDecoder().decode(chunk.value);
+      // retry deserialization
+      return await this.next();
+    } else {
+      // Check if the chunk is a valid chunk
+      try {
+        const result = {
+          done: false,
+          value: deserialize(first),
+        };
+        // if it succeeds, remove the first valid chunk
+        // from the buffer
+        this.buffer = rest.join('');
+        return result;
+      } catch (error) {
+        // otherwise, retry again with a new chunk
+        const chunk = await this.reader.read();
+        console.log(chunk);
+        if (chunk.done) {
+          return null;
+        }
+        this.buffer += new TextDecoder().decode(chunk.value);
+        return await this.next();
+      }
+    }
+  }
+
+  async drain() {
+    while (true) {
+      const result = await this.next();
+      if (result.done) {
+        break;
+      }
+    }
+  }
+}
+
 async function deserializeStream(id, response) {
   if (!response.body) {
     throw new Error("missing body");
   }
-  const reader = response.body.getReader();
+  const reader = new ChunkReader(response.body);
 
-  async function pop() {
-    const result = await reader.read();
-    if (!result.done) {
-      const serialized = new TextDecoder().decode(result.value);
-      const splits = serialized.split("\n");
-      for (const split of splits) {
-        if (split !== "") {
-          deserialize(split);
-        }
+  const result = await reader.next();
+
+  if (!result.done) {
+    reader.drain().then(
+      () => {
+        delete $R[id];
+      },
+      () => {
+        // no-op
       }
-      await pop();
-    }
+    );
   }
 
-  let serialized = ""
-  while (true) {
-    const line = await reader.read();
-    if (line.done) break;
-    serialized += new TextDecoder().decode(line.value);
-  }
-  let pending = true;
-  let revived;
-  const splits = serialized.split("\n");
-  for (const split of splits) {
-    if (split !== "") {
-      const current = deserialize(split);
-      if (pending) {
-        revived = current;
-        pending = false;
-      }
-    }
-  }
-
-  pop().then(
-    () => {
-      delete $R[id];
-    },
-    () => {
-      // no-op
-    }
-  );
-
-  return revived;
+  return result.value;
 }
 
 let INSTANCE = 0;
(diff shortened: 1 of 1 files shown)


packages/start/config/server-runtime.jsx
```diff
@@ -13,55 +13,100 @@ import {
...
+    // Get the first valid seroval chunk
+    const [first, ...rest] = this.buffer.split('\n');
+    // Deserialize the seroval chunk
+    const result = {
+      done: false,
+      value: deserialize(first),
+    };
+    // if it succeeds, remove the first valid chunk
+    // from the buffer
+    this.buffer = rest.join('\n');
```
feynmanliang, line 56: `Array.shift` instead of split/destructure/join?

packages/start/config/server-runtime.jsx
```diff
@@ -13,55 +13,100 @@ import {
...
+      return await this.nextValue();
+    } catch (error) {
+      // If it happens that there's an error again
+      // and we are done reading the buffer
+      // then the whole stream is invalid.
+      if (this.done) {
+        throw new Error('Malformed server function stream.');
       }
+      // Since it's invalid (some syntax-related issue)
+      // we read a new chunk, and hope there's a valid
```
feynmanliang, line 72: I will test this tomorrow, but do we think this will fix #1245 where a single large JS expression is split across two chunks? AFAIK we need to concatenate the two partial chunks so that the entire chunk is a valid expression but I think ATM this code would drop both chunks?

packages/start/config/server-runtime.jsx
lxsmnsyc: I still needed to re-join it since buffer is a string.

packages/start/config/server-runtime.jsx
lxsmnsyc: I've already tested this with a gigantic chunk (as seen with the Test app). The idea is to keep reading until there's a valid seroval (JS) chunk.

feynmanliang: Can you help explain to me? My understanding is that on an invalid chunk https://github.com/solidjs/solid-start/pull/1245/files#diff-d5889473d46bb02f0141e0e1dfa59c3c5c118a8e660eb941117bfc25fd3d1234R52 will throw and `first` will lose all references and be GCed.

lxsmnsyc: Basically, when we call `deserialize` we are also trying to know if the `first` chunk is a valid seroval chunk. Once it deserializes successfully, we remove the `first` chunk from the `buffer`.

If `deserialize` fails then we move to the `catch` block. There we read the next Response chunk, and then attempt to deserialize again. Take note that the `buffer` only shrinks for every successful deserialize call.

We repeat this until a valid seroval chunk has been encountered and/or the response is done and the `buffer` is empty.

feynmanliang: Thanks for helping me out, I get it now. A reader of L52 must be aware of exception handling on L68 in order to reason about correct behavior

feynmanliang:
Really appreciate your help @lxsmnsyc getting this upstream :) Can confirm that this fixes #1243

packages/start/config/server-runtime.jsx
lxsmnsyc: TBF that thrown Error is unlikely to happen, I guess.

ryansolid:
> Really appreciate your help @lxsmnsyc getting this upstream :) Can confirm that this fixes #1243

Awesome thanks for letting us know.

PR #1251, opened 2024-01-13
grouped api endpoint should be lower priority than root api endpiont

jzohdi:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

In a previous PR [here](https://github.com/solidjs/solid-start/pull/1181) I implemented handling of api routes that are under 
a grouped directory. However, it seems that the correct priority is not maintained. See examples-experiments `pnpm --filter example-experiments run dev`:

```
src/
    routes/
        (group)/
            api.ts   <--- this one is being returned
            endpoint.ts
            other.tsx
        api.ts  <--- this one should be returned as it is closer to the route directory
```

`routes/(group)/api.ts`
```ts
export function GET() {
  return new Response("Should not take priority over routes/api.ts");
}
```

`routes/api.ts`
```ts
export function GET() {
  return "hello world";
}

export function DELETE() {
  console.log("Deleted");
}
```

Navigate to `http://localhost:3000/api` and returned is:
```
Should not take priority over routes/api.ts
```

## What is the new behavior?
changing the behavior of `routeToMatchRoute` so that a group segment subtracts from rather than adds to the api route's `score` allows defined routes closer to the route directory to be prioritized. With this change:

Navigate to `http://localhost:3000/api` and returned is:
```
hello world
```

If you delete `routes/api.ts`, `routes/(group)/api.ts` is returned as expected.
```
Should not take priority over routes/api.ts
```

### DIFF
--- a/packages/start/shared/routes.ts
+++ b/packages/start/shared/routes.ts
@@ -161,11 +161,9 @@ function routeToMatchRoute(route: Route): MatchRoute {
         index
       });
       wildcard = true;
-    } else {
+    } else if (!segment.match(/^\(.+\)$/)) {
       score += 4;
-      if (!segment.match(/^\(.+\)$/)) {
-        matchSegments.push(segment);
-      }
+      matchSegments.push(segment);
     }
   }
 
--- a/packages/start/shared/routes.ts
+++ b/packages/start/shared/routes.ts
@@ -162,6 +162,8 @@ function routeToMatchRoute(route: Route): MatchRoute {
       });
       wildcard = true;
     } else if (!segment.match(/^\(.+\)$/)) {
+      score -= 1;
+    } else {
       score += 4;
       matchSegments.push(segment);
     }


nksaraf:
And we are sure this doesn't cause any other changes in routing behaviour in normal pages. We don't have enough tests for this so we need to make sure its working fine somehoww.

jzohdi:
@nksaraf 
more extensive testing will be done/added.

@ryansolid 
I think that the behavior should be explicit and documented at least, would you agree. Three options I can see are: 1. the furthest from the root directory is lower priority 2. the closest to the root is lower priority. 3. An exception is thrown telling the user that the grouping cannot be made since the evaluated paths are identical. Currently the behavior is option 2, which I think is the least intuitive. I believe I have seen other frameworks handle this scenario with option 3. Ideally the behavior is that which aligns best with the user's intuition and expectations which could be gathered by something like a survey.

ryansolid:
Discussion here will probably be relevant to this: https://github.com/solidjs/solid-start/discussions/1279

ryansolid:
Ended up going with the error now in 0.5.0

Issue #1165, opened 2023-12-21
[Bug?]: /_server request gets stuck with async code inside middleware


JesseVelden:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

With the new @solid/start, the `/_server` request gets stuck and shows (pending) as status in Chrome's Network Inspector when using async code inside the middleware. When not using async code, it works fine.

### Expected behavior 🤔

The request resolves

### Steps to reproduce 🕹

Steps: 

1. Go to  https://codesandbox.io/p/devbox/solid-start-middleware-for-server-with-promise-session-not-working-6kz3tm?file=%2Fsrc%2Fmiddleware.ts%3A31%2C57
2. Run the dev server
3. Click on Generate New button in the preview window, and see the that the code actually executes the console.log but the request hangs as can be seen Network Inspector
4. If you uncomment the async code inside the `middleware.ts`, you can see it does work as expected

### Context 🔦

Middleware is needed for checking if a user is logged in. It is uses async functions to do that

### Your environment 🌎

```
Node.js v20.9.0

    "@solidjs/router": "^0.10.5",
    "@solidjs/start": "^0.4.1",
    "solid-js": "^1.8.7",
    "vinxi": "0.0.54"
```

nksaraf:
Yup very peculiar. Will have to debug this carefully.

edivados:
@nksaraf I think the issue is this.

The middleware in front of the server function creats a `FetchEvent` which calls [`toWebRequest`](https://github.com/unjs/h3/blob/53703dc860f1ff6fe7ce71d543deff1cfa810b11/src/utils/body.ts#L252C22-L252C22) creating a ReadableStream from the request body.

https://github.com/solidjs/solid-start/blob/a9b4b0532f76ce9f0c20545b4f6732a126771a7f/packages/start/server/middleware.ts#L59-L62

If I understand it correctly the unreleated await gives node a chance to read chunks and later we get stuck trying to read a request body that has already been read.

https://github.com/solidjs/solid-start/blob/a9b4b0532f76ce9f0c20545b4f6732a126771a7f/packages/start/config/server-handler.js#L102-L103

I verified this by replacing `toWebRequest` with a Proxy and not reading the body.

https://github.com/solidjs/solid-start/blob/a9b4b0532f76ce9f0c20545b4f6732a126771a7f/packages/start/server/middleware.ts#L20-L25

[...]

PR #1255, opened 2024-01-14
Fix server functions hanging on edge functions

oscartbeaumont:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

When you call a server function when Solid Start is deployed on Deno or Vercel Edge Functions the entire function hangs.

Example page at `routes/index.ts` to demonstrate the issue:
```ts
const demoAction = async (name: string) => {
  "use server";

  return `Hello, ${name}`;
};

  return (
    <main class="text-center mx-auto text-gray-700 p-4 flex flex-col">
      <button onClick={() => demoAction("Oscar").then(console.log)}>
        Demo
      </button>
    </main>
  );
}
```

To test locally you can install Deno and use the following:
```
NITRO_PRESET=deno_deploy pnpm build && cd ./.output && deno run -A ./server/index.ts
```

The issue is the call to [`await request.json()` within handleServerFunction](https://github.com/solidjs/solid-start/blob/a670ffa3c498af6d16f44a4156edaca9d1fbc155/packages/start/config/server-handler.js#L94) freezes forever causing the request to hang until the function runtime (Eg. Vercel) times out.

It's worth noting nothing is special about this call to `request.json()`, it's just the place where the body is used. Any usages of the `request.body` will hang (including `.arrayBuffer()`, `.text()`, etc).

SolidStart calls Vinix [`toWebRequest`](https://github.com/nksaraf/vinxi/blob/bc0c2273cc2d39c45ff8725c1c06d74a93eb3ee3/packages/vinxi/runtime/server.js#L136) which in-turn calls H3's [`getRequestWebStream`](https://github.com/unjs/h3/blob/845eecaaaeffdb415321f76744e31d55074f37b8/src/utils/body.ts#L253).

Following along with the implementation of the`getRequestWebStream` function:

`event.web?.request?.body` is `undefined` so we go to the next condition,
`event._requestBody` is `undefined` so we go to the next condition,
We reach the "finally" condition and create a `ReadableStream` and call `event.node.req.on` to set up listeners to the body.

However, none of the `.on(...)` callbacks fire in these Edge environments so the promise hangs and the entire function freezes.

I noticed that `event.node.req.body` is an `ArrayBuffer` so we can bypass the `ReadableStream` and just use it as `body` property of the `Request` which is what this PR does.

This issue is related to:
 - #1165
 - nksaraf/vinxi/pull/118

## What is the new behavior?

The request returns correctly like it does when run using Node.js.

### DIFF
--- a/packages/start/server/middleware.ts
+++ b/packages/start/server/middleware.ts
@@ -1,11 +1,11 @@
 import {
-  defineMiddleware,
   EventHandlerRequest,
+  H3Event,
+  defineMiddleware,
   getRequestIP,
   getRequestURL,
-  H3Event,
-  sendWebResponse,
-  toWebRequest
+  getRequestWebStream,
+  sendWebResponse
 } from "vinxi/server";
 import { FetchEvent } from "./types";
 
@@ -17,6 +17,40 @@ const eventTraps = {
     return target[prop] ?? target[h3EventSymbol][prop];
   }
 };
+
+function toWebRequest(/** @type {H3Event} */ event) {
+  /**
+   * @type {ReadableStream | undefined}
+   */
+  let readableStream;
+
+  const url = getRequestURL(event);
+  const base = {
+    // @ts-ignore Undici option
+    duplex: "half",
+    method: event.method,
+    headers: event.headers
+  };
+
+  if (event.node.req.body instanceof ArrayBuffer) {
+    return new Request(url, {
+      ...base,
+      body: event.node.req.body
+    });
+  }
+
+  return new Request(getRequestURL(event), {
+    ...base,
+    get body() {
+      if (readableStream) {
+        return readableStream;
+      }
+      readableStream = getRequestWebStream(event);
+      console.log(readableStream);
+      return readableStream;
+    }
+  });
+}
 
 export function createFetchEvent(event: H3Event<EventHandlerRequest>): FetchEvent {
   event.web ||
--- a/packages/start/server/middleware.ts
+++ b/packages/start/server/middleware.ts
@@ -1,11 +1,11 @@
 import {
-  EventHandlerRequest,
-  H3Event,
   defineMiddleware,
+  EventHandlerRequest,
   getRequestIP,
   getRequestURL,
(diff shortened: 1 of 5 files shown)


packages/start/server/middleware.ts
```diff
@@ -18,6 +18,40 @@ const eventTraps = {
...
+  }
+
+  return new Request(getRequestURL(event), {
+    ...base,
+    get body() {
+      if (readableStream) {
+        return readableStream;
+      }
+      readableStream = getRequestWebStream(event);
+      console.log(readableStream);
```
lxsmnsyc, line 49: Forgot to remove?

oscartbeaumont: 😳  Whoops.

lxsmnsyc:
@oscartbeaumont also would be helpful if you can point out the location of the file where `handleServerFunction` is declared (just for better context)

nksaraf:
This is awesome! Thanks for finding this. We should try to get this also upstream in `h3` and before that in `vinxi`. We want these fixes to help all the vinxi users. I want to take this whole function upstream first, @edivados has done one bit with the request caching in nksaraf/vinxi#118

I want to get this ArrayBuffer check in there too.

nksaraf:
Upstreaming this in nksaraf/vinxi#123

oscartbeaumont:
Sorry for the lack of detail on the issue, was rushing a bit. I have updated the original issue with a full rundown of what's going on.

@nksaraf I was just about to ping you. You raise a good point this should probably be fixed upstream.

I am working on a Nitro/h3 PR, I'm honestly not sure who is at fault.

Thanks both, am really loving Solid Start!

oscartbeaumont:
I have opened an upstream issue/PR in H3:
 - https://github.com/unjs/h3/issues/615
 - https://github.com/unjs/h3/pull/616

I am closing this issue as a `pnpm update --latest --recursive` updates Vinxi and patches the issue.

PR #1268, opened 2024-01-19
fix: dev overlay stuff

lxsmnsyc:
- Fix source to point to the exact file, which allows the code view to actually show the preview of the code rather than showing a fallback UI.
Start Screenshot (12)

- Fix screenshot generation


ryansolid:
Thanks

PR #1301, opened 2024-02-03
fix unocss example

birkskyum:
I couldn't build the unocss example - it was because it wasn't bumped to vinxi 0.2.x like the rest of the example. Also set vite consistently to 4.5.x (like vinxi), where currently both 4.4.6 and 4.4.9 are used.


ryansolid:
Thanks

Issue #140, opened 2022-06-18
Enable typescript strict mode


modderme123:
There are a couple of types that have an optional undefined, enabling strict mode will help with handling them (also it will make sure that users of solid-start with strict mode enabled don't get those errors in their project)

nksaraf:
Could you point out the specific ones so we could deal with them? Or make a PR?

modderme123:
Sure, I'll make a PR to use Typescript's `?.` operator

ryansolid:
I guess we are in good versions of node to do that. I have been avoiding `?.` and `??` as it tends to break some tooling.

modderme123:
Update: when I enable strict mode [here](https://github.com/solidjs/solid-start/blob/main/packages/start/tsconfig.json], I get ~300 errors from `tsc --noEmit`. I've cut that down to about 56 that I'm manually working through

davycoe:
the following also produces errors in my tsconfig:
```
"strictNullChecks": true,
"noImplicitAny": true,
"noImplicitReturns": true,
"noUnusedLocals": true,
"noUnusedParameters": true,
"strict": true
```

nksaraf:
@modderme123 could you make a PR for what you have! I can work on top to get it to completion for strict mode

modderme123:
Just got back to this and after rebasing, it looks like you've got most of the initial errors. Only 2 files had any changes that still were valid for the current codebase and they'd be trivial to recreate

PR #1303, opened 2024-02-03
fix tailwind typing in vite config

birkskyum:
there was a bug in the ts typings of tailwindcss v3.3.3, leading to a lot of red lines in vite.config

related to
- #140


ryansolid:
Thanks

PR #1313, opened 2024-02-06
Redirect helper component

marvin-j97:
## Scenario

While implementing a catch-all 404 redirect I noticed there's no built-in way for redirecting, so one has to be aware of the `location` header name and the correct HTTP status codes. This Redirect component is supposed to simplify that, and simply wraps the existing `HttpHeader` and `HttpStatusCode` components.

```tsx
// [...404].tsx

import { HttpHeader, HttpStatusCode } from "@solidjs/start";

export default function NotFound() {
  return (
    <>
      <HttpHeader name="location" value="/" />
      <HttpStatusCode code={307} />
    </>
  );
}
```

becomes

```tsx
// [...404].tsx

import { Redirect } from "@solidjs/start";

export default function NotFound() {
  return <Redirect to="/" />;
}
```

### Maybe allow setting the status code to 301 | 302 | 303 | 307 | 308 like Next.JS does.

The component could default to 307, allow `permanent` for 308 (these two should cover 99% of use cases), but the user could instead use `<Redirect to="/" statusCode={301} />` for specific status codes if required.

### DIFF
--- a/docs/routes/api/Redirect.md
+++ b/docs/routes/api/Redirect.md
@@ -26,13 +26,13 @@ import { Redirect } from "@solidjs/start";
 
 ### Using permanent redirects
 
-`Redirect` defaults to a temporary redirect (HTTP status code 307), but can be made permanent, sending a 308 status instead:
-
 ```tsx twoslash
 import { Redirect } from "@solidjs/start";
 // ---cut---
 <Redirect to="/some-other-page" permanent />;
 ```
+
+`Redirect` defaults to a temporary redirect (HTTP status code 307), but can be made permanent, sending a 308 status instead.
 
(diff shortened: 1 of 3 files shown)


ryansolid:
I think <Redirect /> belongs in the router.. And I think it already has one: https://github.com/solidjs/solid-router?tab=readme-ov-file#navigate-. Although I suppose we don't set a status other than 302. So that is worth consideration I suppose.

marvin-j97:
> I think <Redirect /> belongs in the router.. And I think it already has one: https://github.com/solidjs/solid-router?tab=readme-ov-file#navigate-. 

What if one doesn't use solid-router specifically?

PR #1341, opened 2024-02-19
fix auto import proposal

AirBorne04:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes
- [x] Other... Please describe:
module auto import fix (loosing the extra `./`)
Screenshot 2024-02-19 at 17 11 29

Currently the proposed import is showing e.g. `@solidjs/start/./server`

## What is the new behavior?
as in screenshot is showing the proper import `@solidjs/start/server`

### DIFF
--- a/.changeset/hip-insects-ring.md
+++ b/.changeset/hip-insects-ring.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/start": patch
+---
+
+fix: resolve the import path correctly (@solidjs/start/server vs @solidjs/start/./server)
--- a/packages/start/package.json
+++ b/packages/start/package.json
@@ -33,13 +33,13 @@
       ".": [
         "./shared/index.tsx"
       ],
-      "./middleware": [
+      "middleware": [
         "./middleware/index.ts"
       ],
-      "./server": [
+      "server": [
         "./server/index.tsx"
       ],
-      "./client": [
+      "client": [
         "./client/index.tsx"
       ]
     }


ryansolid:
My understanding is that paths needed to prefixed as relative otherwise they could be mistaken for export conditions. I suppose top level that wouldn't be the case. But I haven't seen that suggestion before. And every package I've seen has had the `./` in front.

Here are docs from a couple sources:
node: https://nodejs.org/api/packages.html#subpath-exports
webpack: https://webpack.js.org/guides/package-exports/

Which IDE is this? I haven't seen that suggestion before. Although I have seen it complete to `@solid-js/start/.` at one point but I assumed that was the IDEs issue. Is this tied to TS resolution I wonder. Love to know a bit more before making this change.

AirBorne04:
You are correct with the subpath exports, VSCode (i am using it on Mac) has even told me when i was changing the exports. But the change is made on the `typeVersions` field, that seems to have a different syntax, the [typescript docs](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#version-selection-with-typesversions) do not really state something specific though.

I saw this weird resolving on your stream, thats why I though I might take a shot at fixing it. 😄 
Cross check with astrojs `package.json` shows they also have the proposed change. 
Screenshot 2024-02-20 at 20 15 33

PR #1348, opened 2024-02-28
Fix refresh pragma

lxsmnsyc:
- Fixes the examples to use the `@refresh reload` pragma at the correct place (the entry files)
- Fixes tsx files in the core package to use `@refresh skip`.
- Bonus: fixes components that use isServer to use the variable at module-level so that the bundler can perform dead-code elimination properly.


ryansolid:
Thanks @lxsmnsyc

PR #1378, opened 2024-03-11
Set title required and add placeholder

birkskyum:
I found it a bit annoying that the title was "Untitled", not as a placeholder but as the actual value. Now there is a "Title" placeholder instead, and it's a required field.

Before:

Screenshot 2024-03-11 at 21 30 01

After (if you click save without typing)
Screenshot 2024-03-11 at 21 29 36


ryansolid:
Thanks.. I just copied it across this makes way more sense.

PR #1379, opened 2024-03-11
Remove the archived notes example

birkskyum:
I can't run the archived example, and I remember it was somewhat buggy on editing as well, so this PR removes it in favor of the updated one.

I get `ReferenceError: React is not defined` errors

unplugin-icons was in the vite.config.ts in the archived version, but didn't appear to be in use.


ryansolid:
Yeah I left those 2 around when we get back to looking at HTML partial solutions.

PR #1388, opened 2024-03-14
Improve cli docs

birkskyum:
@atilafassina  This removed the partial duplication of the vinxi cli docs from the Start API (it technically never belonged there). Instead there is a now a getting started section, and links to the better documented [vinxi cli](https://vinxi.vercel.app/api/cli.html) docs.

Before:
Screenshot 2024-03-14 at 13 01 14

After:
Screenshot 2024-03-14 at 13 00 47


ryansolid:
Thanks

Issue #1408, opened 2024-03-22
[Bug?]: Page Nav randomly into Blank, with "Uncaught (in promise) SyntaxError: missing ) after argument list"


hardywu:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

Strange error and hard to reproduce. Nav around the site, when nav to a page previously visited, the page has chance to go blank. It's supposed to load data and update the page.

```js console
Uncaught (in promise) SyntaxError: missing ) after argument list
    Le index.mjs:17
    next server-runtime.js:60
    next server-runtime.js:36
    D server-runtime.js:77
    B server-runtime.js:143
    v server-runtime.js:165
    n cache.js:100
    load implies.tsx:42
    P routing.js:348
    R solid.js:844
    He solid.js:543
    P routing.js:348
    ke/</g/n[_.pathname]< events.js:71
    setTimeout handler*g events.js:70
    ke events.js:113
    _e components.jsx:14
    $e Router.js:40
    s solid.js:1324
    C solid.js:477
    Qe solid.js:1324
    as app.tsx:12
    s solid.js:1324
    C solid.js:482
    Qe solid.js:1324
    get children StartClient.jsx:12
    get children ErrorBoundary.jsx:12
    On solid.js:1656
    gt solid.js:524
    On solid.js:1656
    ke solid.js:735
    z solid.js:714

```

### Expected behavior 🤔

load data and update page on the nav click.

### Steps to reproduce 🕹

Steps:

1. nav (by anchor click) to the posts list page
2. nav (by click) to other pages
3. nav (by click) to the posts list page again

### Context 🔦

The page has data loading logic like this.
```
const getPosts = cache(async (q: string, page: number) => {
  "use server";
  const userId = await currentUserIdCache();
  if (!userId) throw new Error("Not authed");
  // await new Promise((res) => setTimeout(res, 3000));
  return await postSvc.query(
    { authorId: userId },
    {
      size: limit,
      page, 
      withAuth: true,
    }
  );
}, "dash/posts");

export const route = {
  load({ location }) {
    void getPosts(location.query.q || "", +location.query.page || 1);
  },
} satisfies RouteDefinition;
```

### Your environment 🌎

```shell
System:
  OS: Ubuntu 22.04 LTS (jammy)
  CPU: Intel(R) Xeon(R) CPU E5-2697 v2 @ 2.70GHz
Binaries:
[...]

PR #1418, opened 2024-03-25
fix(SerovalChunk): incorrect buffer length on special chars

hardywu:
node.js returns incorrect string length on occuring of special charaters. Use encoded data length instead.

fixing #1408 

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

data.length in `createChunk` is less than the `(new TextEncoder().encode(data)).length`

## What is the new behavior?
not use data.length for buffer process

### DIFF
--- a/packages/start/src/runtime/server-runtime.ts
+++ b/packages/start/src/runtime/server-runtime.ts
@@ -17,10 +17,12 @@ import { createIslandReference } from "../server/islands/index";
 class SerovalChunkReader {
   reader: ReadableStreamDefaultReader<Uint8Array>;
   buffer: string;
+  bufferSize: number;
   done: boolean;
   constructor(stream: ReadableStream<Uint8Array>) {
     this.reader = stream.getReader();
     this.buffer = "";
+    this.bufferSize = 0;
     this.done = false;
   }
 
@@ -30,6 +32,7 @@ class SerovalChunkReader {
     if (!chunk.done) {
       // repopulate the buffer
       this.buffer += new TextDecoder().decode(chunk.value);
+      this.bufferSize += chunk.value.length;
     } else {
       this.done = true;
     }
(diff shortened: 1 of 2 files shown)


lxsmnsyc:
@hardywu do you have an example data that I can check and test?

hardywu:
@lxsmnsyc 
```js
> const str = 'à\u0301';
undefined
> str.length
2
> (new TextEncoder().encode(str)).length
4
>
···

**Comment by lxsmnsyc** [comment]
@hardywu Thanks! This makes sense.

File: packages/start/src/runtime/server-runtime.ts [hunk 1]
```diff
@@ -55,7 +58,7 @@ class SerovalChunkReader {
     // deserialize the data
     const bytes = Number.parseInt(this.buffer.substring(1, 11), 16); // ;0x00000000;
```
lxsmnsyc, line 59: @hardywu does the substring still work for special characters? Would like to know.

hardywu: substring fn works by charater. Special characters are treated the same as others.  So yes. It still works.

lxsmnsyc:
Only part that needs fix is when we consume a piece of chunk from the stream: https://github.com/solidjs/solid-start/blob/5c34ca7fedd040d6a133b0991bbb65667acd1e1a/packages/start/src/runtime/server-runtime.ts#L73

The commit for `bufferSize` doesn't correctly reflect the change when we consume from the buffer, we might want to resize `bufferSize`.

Other than that, the rest looks good to me.

hardywu:
> Only part that needs fix is when we consume a piece of chunk from the stream:
> 
> https://github.com/solidjs/solid-start/blob/5c34ca7fedd040d6a133b0991bbb65667acd1e1a/packages/start/src/runtime/server-runtime.ts#L73
> 
> The commit for `bufferSize` doesn't correctly reflect the change when we consume from the buffer, we might want to resize `bufferSize`.
> 
> Other than that, the rest looks good to me.

How about keeping the buffer of type Unit8Array? Less worries about the length difference.

lxsmnsyc:
@hardywu whichever works 😄

lxsmnsyc:
LGTM, thanks!

PR #1423, opened 2024-03-27
fix: add app.tsx assets to page event in dev env

katywings:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

During development `app.css` is only loaded in the browser via javascript, resulting in a flash of unstyled content.

## What is the new behavior?
`app.tsx` is registered for asset collection in the pageEvent on the server. This way `app.css` and other potential css assets of `app.tsx` are added to the document head, during server side rendering.

## Open Question
I am not 100% happy with the inlined `src/app.tsx`, do we have a proper way to access the path to `app.tsx`?

## Other information
How it currently looks in a browser with disabled js:
IS

How it should look:
SHOULD
owser with disabled js:

### Credits

@nksaraf helped me to find the proper place to debug the issue 🙏.

### DIFF
--- a/.changeset/wise-avocados-lay.md
+++ b/.changeset/wise-avocados-lay.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/start": patch
+---
+
+fix: add app.tsx assets to page event in dev env
--- a/packages/start/src/server/pageEvent.ts
+++ b/packages/start/src/server/pageEvent.ts
@@ -30,6 +30,7 @@ export async function createPageEvent(ctx: FetchEvent) {
     manifest: await clientManifest.json(),
     assets: [
       ...(await clientManifest.inputs[clientManifest.handler]!.assets()),
+      ...(import.meta.env.DEV ? await clientManifest.inputs['src/app.tsx']!.assets(): []),
       ...(import.meta.env.START_ISLANDS
         ? (await serverManifest.inputs[serverManifest.handler]!.assets()).filter(
             s => (s as any).attrs.rel !== "modulepreload"


katywings:
I gonna add the missing changeset tomorrow 😅.

edivados:
Can't reproduce with `vinxi 0.3.10` but with `vinxi 0.3.11` the issue is there. 

With 0.3.10 app.css and dev-overlay/styles.css are present here.
https://github.com/solidjs/solid-start/blob/2d75d5fedfd11f739b03ca34decf23865868ac09/packages/start/src/server/pageEvent.ts#L32

Seems like this is because of https://github.com/nksaraf/vinxi/commit/04b3ff77c5e8a15c9b8c93ccf1c5ae96850230b8 causing start's client/index.jsx to not be processed anymore. I lack insight here... may be intended by vinxi and should now be solved in start.

<hr>

>I am not 100% happy with the inlined src/app.tsx, do we have a proper way to access the path to app.tsx?

src directory (appRoot) is user configurable and jsx extension is also possible.

Suggestions:

#### Define global constant replacements

https://github.com/solidjs/solid-start/blob/2d75d5fedfd11f739b03ca34decf23865868ac09/packages/start/config/index.js#L135-L139
```js
...(import.meta.env.DEV ? await clientManifest.inputs[import.meta.env.START_SOME_NAME]!.assets(): []),
```

#### Throw it on the ssr router's vite config
```js
config("user", {
     ...userConfig,
     start: {
         someName: ""
     }
```
```js
...(import.meta.env.DEV ? await clientManifest.inputs[serverManifest.dev.server.config.start.someName]!.assets(): []),
```

katywings:
@nksaraf What do you think about @edivados suggestions?

nksaraf:
Yeah this suggestion makes sense. The compile time/build time can provide the necessary paths and the runtime can use that.

PR #1440, opened 2024-04-08
fix: ensure nonce is passed to all assets

paularmstrong:
## PR Checklist

## PR Type

- [x] Bugfix

Assets rendered to the head of the document do not include the `nonce` if set, eg:

```js
export default createHandler((context) => (
  <StartServer document={{ assets, children, scripts }}>
    <html>
      <head>
        {assets}
      </head>
      <body>
        {children}
        {scripts}
      </body>
    </html>
  </StartServer>
), (context) => ({ nonce: 'some-random-nonce' }));
```

Results in:

```html
<html>
  <head>
    <!-- These assets are missing the nonce! -->
    <link rel="stylesheet" href="…" />
    <script type="module" src="/_build/@vite/client" id="vite-client" />
  </head>
  <body>
    <!-- children -->
    <script type="module" nonce="some-random-nonce" src="...entry-cleint.tsx" />
  </body>
</html>
```

## What is the new behavior?

```html
<html>
  <head>
    <!-- These assets now have the nonce! -->
    <link rel="stylesheet" nonce="some-random-nonce" href="…" />
    <script type="module" nonce="some-random-nonce" src="/_build/@vite/client" id="vite-client" />
  </head>
  <body>
    <!-- children -->
    <script type="module" nonce="some-random-nonce" src="...entry-cleint.tsx" />
  </body>
</html>
```

## Other information

It's possible to provide nonce's for most types of assets, including but not limited to scripts, as well as stylesheets. This ensures that _all_ assets, including those coming from the solid-meta receive the nonce, not just the scripts at the end.

### DIFF
--- a/.changeset/fifty-nails-repair.md
+++ b/.changeset/fifty-nails-repair.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/start": patch
+---
+
+Ensure `nonce` is passed to all assets when rendering from the server.
--- a/packages/start/src/server/StartServer.tsx
+++ b/packages/start/src/server/StartServer.tsx
@@ -3,12 +3,12 @@
 import App from "#start/app";
 import type { Component, JSX } from "solid-js";
 import {
-    Hydration,
-    HydrationScript,
-    NoHydration,
-    getRequestEvent,
-    ssr,
-    useAssets
+  Hydration,
+  HydrationScript,
+  NoHydration,
+  getRequestEvent,
+  ssr,
+  useAssets
 } from "solid-js/web";
 import { ErrorBoundary } from "../shared/ErrorBoundary";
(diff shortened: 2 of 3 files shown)


paularmstrong:
> [!IMPORTANT]
> Hydration key scripts still don't seem to be receiving the nonce. I'm still tracking down why this is the case.

atilafassina:
I ran into this issue recently.
When a `nonce` is found in a directive (such as `script-src`), it is expected to exist in all others (e.g.: `style-src`). And make `unsafe-inline` also forbidden.

This makes it impossible to use `nonce` in SolidStart without raising issues because `renderAssets` is not passing down the `nonce` value to `<link>` and `<style>` tags.

From what I could see in the code, this PR is the right fix for it.

Issue #1466, opened 2024-05-02
[Bug?]: The example in the docs doesn't seam to work.


RonenEizen:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

The example in the [documentation](https://start.solidjs.com/api/RequestEvent) doesn't seam to work.

```
declare module "@solidjs/start/server" {
  interface RequestEventLocals {
    myNumber: number;
    someString: string;
  }
}
```

`locals` property within the RequestEvent doesn't have any types declared in global.d.ts.  For example in  one of the routes:

```
import { getRequestEvent } from 'solid-js/web'

const getMessage = cache(async () => {
  'use server'
  const event = getRequestEvent()
  // event?.locals is not typed
  return
}, 'message')

export const route = {
  load: () => getMessage(),
}
```

and my global.d.ts has

```
declare module '@solidjs/start/server' {
  interface RequestEventLocals {
    myNumber: number
    someString: string
  }
}
```

### Steps to reproduce 🕹

Steps:

1. Create an solid-start app using one of the templates
2. Declare  '@solidjs/start/server'  within the global.d.ts file
3. Try accessing `event.locals.<SOMETHING>` within a route function

ryansolid:
I am going to ask for a reproduction. We have an example template `experiments` that has this usage in it and it appears to be working.

https://github.com/solidjs/solid-start/blob/main/examples/experiments/src/entry-server.tsx#L4-L9
https://github.com/solidjs/solid-start/blob/main/examples/experiments/src/routes/index.tsx#L12-L13

(the example uses an undeclared value, but switch it to n/s shows the right types).

RonenEizen:
@ryansolid It does work if declared within entry-server.tsx. I created a [repository](https://github.com/RonenEizen/solid-typescript-bug-example/blob/main/src/routes/index.tsx) based on a basic template. The types still don't work when declared within global.d.ts file. Look at the line 9 within the `routes/index.tsx` file.

ryansolid:
Oh.. hmm so this is a weird typescript-ism.. Lovely. Not sure what to do with this then.

RonenEizen:
I am personally fine with the workaround. Should I keep this issue? Or would you rather have it resolved?

AirBorne04:
[...]

PR #1469, opened 2024-05-05
Move RequestEventLocals type definition into a namespace to make the retyping easier for users

AirBorne04:
## PR Checklist

## PR Type

- [X] Other... Type changes that require minor adjustments

When defining a `RequestEventLocals` type for a custom app typescript does not pickup the types until it the file also contains a top level import/export.

## What is the new behavior?
Now instead of declare module we declare a namespace that does not confuse typescript and makes sure the `global.d.ts` is interpreted as an augmentation module (not ambient as before)

## Other information

Related to #1466 unfortunately this would require current users to re-config their typings. (from module to namespace)


ryansolid:
I haven't touched this one since it is marked as draft. Are there more changes, or things that need to be verified?

AirBorne04:
hi @ryansolid mainly looking for approval, and adjustment on the docs side? Any thoughts?

ryansolid:
Ok I like it. We probably have to wait for minor given how this could break types. I've broken types on patch releases before but I feel like we should be cautious here.

AirBorne04:
> Ok I like it. We probably have to wait for minor given how this could break types. I've broken types on patch releases before but I feel like we should be cautious here.

Perfect, I feel the same, I have updated the changeset accordingly.

atilafassina (approved):

apatrida:
This broke the typing in TypeScript in 1.0.0, nothing can be done to use this module and override it.

PR #1481, opened 2024-05-18
Omit route components from non-SSR builds when `ssr: false`

Brendonovich:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

When building with `ssr: false`, the ssr + server-fn builds still transform + bundle all of the route components in filesystem routes.
This doesn't really make sense, since the ssr build will only ever be used during prerendering, which just renders `entry-server.tsx` and `app.tsx`, and `server-fn` only invokes `App` with `dataOnly` set on the router so none of the route components are actually used.

## What is the new behavior?

This PR adds a `dataOnly` config to `SolidStartServerFileRouter` which makes it omit the default exported component from all routes. Non-default exported components work the same and `export const route` is still included since it's needed on the server for SFM.

### DIFF
--- a/.changeset/twenty-monkeys-marry.md
+++ b/.changeset/twenty-monkeys-marry.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/start": patch
+---
+
+Omit route components from non-SSR builds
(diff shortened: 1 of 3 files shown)


ryansolid:
That's awesome.. Thank you. I was meaning to look at this and I'm pretty stoked that you applied this optimization.

ryansolid:
Hmm.. it breaks the notes demo. It is no longer single flight in SSR false. Which suggests something isn't working the way it should. I will see if I can figure it out in the next 30 mins otherwise moving forward with the release I might have to temporarily revert this.

PR #1482, opened 2024-05-18
Reintroduce flat routing

Brendonovich:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

Changes to nested route escaping in RC0 broke what was essentially flat routing, where files/folders named along the lines of `a.b.c` would have their path become `a/b/c`. This is useful for reducing folder nested. Whether it was an intentional behaviour previously I'm not sure, but I'd like it to be there.

## What is the new behavior?

Adds back the `. -> /` regex that was removed in RC0, bringing back flat route support.

## Other information

Having flat routing alongside nested routing does add the potential for multiple definitions of the same route, but IMO the convenience outweighs that risk

### DIFF
--- a/.changeset/fifty-ways-think.md
+++ b/.changeset/fifty-ways-think.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/start": patch
+---
+
+Reintroduce flat routing
--- a/packages/start/config/fs-router.js
+++ b/packages/start/config/fs-router.js
@@ -1,22 +1,31 @@
 import { analyzeModule, BaseFileSystemRouter, cleanPath } from "vinxi/fs-router";
 
+function toPathBase(src, config) {
+  const routePath = cleanPath(src, config)
+    // remove the initial slash
+    .slice(1)
+    .replace(/index$/, "")
+    // replace . with / for flat routes - e.g. foo.bar -> foo/bar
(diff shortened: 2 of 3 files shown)


ryansolid:
One thing I wasn't sure about here was what this did for people who wanted `.` in the path. I added it to solve unnesting very briefly and almost immediately regretted it. That's why I went back against this because I could already picture the issues being opened when someone made `route.json.ts` and it wasn't at `route.json`. There are cases where people make extension specific routes at minimum. Especially for known things like... `manifest.json` etc. I know Remix does flat routing. Did they ever solve this?

Brendonovich:
[Looks like](https://remix.run/docs/en/main/file-conventions/routes#dot-delimiters) Remix use `[]` to escape characters, whether that's individual periods with `manifest[.]json.ts` or entire file names like `[manifest.json].ts`. They use `$` as their dynamic route prefix similar to how SS uses `:` so there's no conflict with dynamic files, but they also use `$.ts` on its own (which i quite like tbh) to do splat routes, whereas for us doing something like `[...rest].tsx` would conflict with our splat routes.

brenelz:
Just going to close this for now until we revisit

PR #1497, opened 2024-05-22
fix: add error boundary to catch invalid code within the document code

LukasGerm:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

https://github.com/nksaraf/vinxi/issues/119

If you add gibberish code to your `entry-server.tsx` within the `document` property, the server will crash and you need to manually start it again.

## What is the new behavior?

The server will error but not crash. You can fix the issue with your code and just reload your browser.

## Other information
nothing to add


ryansolid:
I think this is the best we have right now. Funny thing is it adds comments around the HTML document which I don't think has any issues but it isn't how I ever expected this to work.

ryansolid:
No it isn't easy to do. I have an idea of a solution but I basically need to remake a limited version of the errorboundary from scratch..

ryansolid:
Looking closer I see this solution doesn't actually show the error but makes the dev server not crash. Let me see if we can do more here.

PR #1521, opened 2024-05-31
Safeguard of H3Event body stream locking in edge runtime workaround

supersoniko:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

https://github.com/solidjs/solid-start/blob/main/packages/start/src/runtime/server-handler.ts#L136
There's currently a workaround for edge runtimes to create a new Request object instead of using the current one.

## What is the new behavior?
Add a safeguard for the edge runtime workaround when a stream is already locked to not create the new Request object.

## Other information
My goal is to make Solid Start work 100% on the Azure Functions runtime. I currently have a [PR](https://github.com/unjs/nitro/pull/2479) open at Nitro to support the latest Azure Function runtime. 

However there's currently an issue that the H3 event body stream is already locked when it gets to the part of the edge runtime workaround when running on Azure Functions, which causes issues with server actions.

Im happy to say that all Solid Start functionality seems to work on Azure Functions once we get around this locking issue.


ryansolid:
Yeah I see hacks on hacks in there. I'm glad you figured something out. That being said. I'm going to probably need to test this fix against Vercel/Netlify edge as I thought the creation of the new Request was to get around stream locking so this fix may impact them.

PR #1532, opened 2024-06-08
Pass event to getResponseStatus in set statusCode

Brendonovich:
Makes behaviour of HttpStatusCode and fetchEvent a bit more consistent

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

Setting `event.response.statusCode` in non-AsyncLocalStorage environments currently throws an error from deep inside FetchEvent.

## What is the new behavior?

HTTPStatusCode will now throw when calling `getRequestEvent` rather than while setting `statusCode`, which is a bit more understandable imo.

### DIFF
--- a/.changeset/fifty-shrimps-cross.md
+++ b/.changeset/fifty-shrimps-cross.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/start": patch
+---
+
+Pass event to getResponseStatus in set statusCode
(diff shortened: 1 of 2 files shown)


ryansolid:
Thanks

PR #1553, opened 2024-06-17
Return result verbatim if `raw=true` is provided in the query string

### LINKED ISSUE #1550: [Bug?]: Cannot get response returned by a server function via a GET request
### Duplicates

### Latest version

### Current behavior 😯

Making a GET request to the correct endpoint with `serverFunction.url`, as well as the `args` query param results in an error:
```
{
  "statusCode": 500,
  "stack": []
}
```

### Expected behavior 🤔

The response from the server function should be sent verbatim to the client.

### Steps to reproduce 🕹

Steps:

1. Load https://stackblitz.com/edit/github-5jq8hh?file=src%2Froutes%2Findex.tsx
2. Observe the logged server error

### Context 🔦

I have attempted to fix this myself by patching Start:
I rewrote this
image
To this
image

This fixes the issue, but overwrites behaviour that I honestly don't fully understand.

### Your environment 🌎

_No response_

Tommypop2:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

Described in #1550 

## What is the new behavior?
Can now specify that the response is returned verbatim by adding `?raw=true` to the query string

Closes #1550


OrJDev (approved):
LGTM

ryansolid:
So if i get this right.. this is for something like an image tag.. so it falls into the no js case but our setting of flash headers messes things.. or I guess more so us messing with the body encoding with seroval would be the issue. I don't like having to rely on query param.. but I also don't see a particularly great way to handle this in this case.

Our options are to use the query param as you have if the caller makes that decision. Or to do something on the response, if the response makes that decision. It sort of feels like the response knows what it wants sent back. We could use the content-type headers of the response object maybe to make this decision. I guess that is a bit weird.

I'm just unclear all the scenarios here and before adding a convention I'd like to consider them.

Brendonovich:
IMO we should use the `Accept` request header here instead, just to account for the case where seroval is wanted.
If `Accept` has `text/javascript` then seroval is used, otherwise it's returned verbatim (unless it's a redirect ofc)

OrJDev:
> So if i get this right.. this is for something like an image tag.. so it falls into the no js case but our setting of flash headers messes things.. or I guess more so us messing with the body encoding with seroval would be the issue. I don't like having to rely on query param.. but I also don't see a particularly great way to handle this in this case.
> 
> 
> 
> Our options are to use the query param as you have if the caller makes that decision. Or to do something on the response, if the response makes that decision. It sort of feels like the response knows what it wants sent back. We could use the content-type headers of the response object maybe to make this decision. I guess that is a bit weird.
> 
> 
> 
> I'm just unclear all the scenarios here and before adding a convention I'd like to consider them.

How about a response header
```
X-Solid-Start-Raw: 1
```

ryansolid:
I went with going with "X-Content-Raw". So not going with this PR. Thanks for bring this up to my attention.

PR #1586, opened 2024-07-21
fix: store fetchEvent in h3 context

katywings:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

`getFetchEvent` is flaky, creating & storing multiple `fetchEvents` per `h3Event`. The reason for this is vinxi duplicating the code including the `fetchEventSymbol` per router (https://github.com/nksaraf/vinxi/issues/51), e.g. if you console.log the symbol https://github.com/solidjs/solid-start/blob/23b01da59b2e6a0e0139686605a7cbf8d723be97/packages/start/src/server/fetchEvent.ts#L16 you will notice that the console.log is executed twice during startup. Also if you log `Object.getOwnPropertySymbols(h3Event)` in `getFetchEvent` you get `[ Symbol(fetchEvent), Symbol(fetchEvent) ]` on requests that affect multiple vinxi routers.

## What is the new behavior?
`getFetchEvent` will maximally create one `fetchEvent` per `h3Event` by storing the `fetchEvent` in `h3Event.context`. `h3Event` is stable across multiple routers and `h3Event.context` is specifically made for situations like this as @pi0 already confirmed in an other thread (https://github.com/solidjs/solid-start/pull/1203#issuecomment-1889036169).

### DIFF
--- a/.changeset/quiet-mirrors-judge.md
+++ b/.changeset/quiet-mirrors-judge.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/start": patch
+---
+
+Fixed fetchEvent flakyness by storing the event in h3 context.
--- a/packages/start/src/server/fetchEvent.ts
+++ b/packages/start/src/server/fetchEvent.ts
@@ -13,7 +13,7 @@ import {
 } from "vinxi/http";
 import type { FetchEvent, ResponseStub } from "./types";
 
-const fetchEventSymbol = Symbol("fetchEvent");
+const fetchEventContext = "solidFetchEvent";
 
 export function createFetchEvent(event: H3Event): FetchEvent {
 	return {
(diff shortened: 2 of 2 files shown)


boehs:
lgtm, but why are symbols no longer being used? I assume the context supports them?

katywings:
@boehs You can think of the symbol like an object, when you create one its always a new one. And there exist multiple ones because vinxi duplicates the code.

Maybe this helps to explain:
```
Symbol("foo") === Symbol("foo"); // false
"foo" === "foo" // true
```

ryansolid:
Interesting. Well good catch.

PR #1609, opened 2024-08-22
extract ErrorBoundary style

AlexErrant:
## PR Checklist

## PR Type

- [x] Other... Please describe: Security improvement

Inline styles are an attack vector [0](https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#output-encoding-for-css-contexts), [1](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src#unsafe_inline_styles), [2](https://dontkry.com/posts/code/disable-inline-styles.html). They may be prevented with the Content Security Policy `style-src 'self';` [as recommended by OWASP](https://cheatsheetseries.owasp.org/cheatsheets/Content_Security_Policy_Cheat_Sheet.html#basic-non-strict-csp-policy).

## What is the new behavior?

This PR moves the inline style out to a stylesheet so users can use `style-src 'self';`.


ryansolid:
If I understanding the issue: With this setting turned on inline styles are completely disabled. What we are doing here is not a security vulnerability because we aren't using CSS variables. But if one would want to protect against a potential security vulnerability they would want to enable this setting to disable inline styles.

Did you check that it works? I'm not seeing the CSS file being included anywhere.

AlexErrant:
> Did you check that it works?

...nope 🤦‍♂️

Upon checking, I realized that the dev-overlay, specifically `shikiji`, also uses inline styles, so this PR would not ultimately fix the problem.

A future possibility could be to put the dev-overlay in an iframe... I *think* that might isolate it from a `Content-Security-Policy` of `style-src 'self';`. However I'm not sure how to load Solid into something an iframe can host, so I'm closing.

> we aren't using CSS variables

Ah, one of my OP's links was misleading; I removed it. CSS variables are not relevant to this issue.

PR #1641, opened 2024-09-29
Make dev overlay lazy

brenelz:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes
- [X] Other... Please describe:

The current behavior is that the syntax highligher always loads with the pages even when there is no error

image

## What is the new behavior?
The new behavior makes the dev overlay lazy so it only loads the syntax highlighting when their is an actual error.

I may be misunderstanding something, and this won't work but seems like it would.

image

### DIFF
--- a/packages/start/src/shared/dev-overlay/index.tsx
+++ b/packages/start/src/shared/dev-overlay/index.tsx
@@ -4,12 +4,12 @@ import {
   Show,
   createEffect,
   createSignal,
+  lazy,
   onCleanup,
   resetErrorBoundaries,
   type JSX
 } from "solid-js";
 import { HttpStatusCode } from "../HttpStatusCode";
-import clientOnly from "../clientOnly";
 
 export interface DevOverlayProps {
   children?: JSX.Element;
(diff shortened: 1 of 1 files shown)


lxsmnsyc:
`clientOnly` is already `lazy`. The reason we need `clientOnly` is because part of the component loads libraries that are meant to be on the client-side.

brenelz:
Yeah maybe I don't understand clientOnly vs lazy. I just noticed it loads the `shiki` library even when the overlay isn't present in dev mode. I'd think it would only load that javascript if the `<DevOverlay />` is shown?

ryansolid:
It isn't actually lazy.. we preload clientOnly even if it isn't rendered yet. So the desire here is legit. I think we need to both. Maybe a prop. Let me think of something.

lxsmnsyc:
@ryansolid yeah I just realized. We could probably rework `clientOnly` to be lazy by default, there isn't exactly a pro and con for it not being lazy right?

ryansolid:
I did preload intentionally so that it would load slightly faster even if not rendered right away. I was coming from the perspective that you wanted the code to only run in the client not to be lazy. It is easy to add it as an options argument. I think I will just do that.

ryansolid:
I fixed it a different way by adding lazy option to clientOnly.

PR #1647, opened 2024-10-05
fix: pass errorless ErrorEvent to dev overlay

katywings:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

Certain `ErrorEvent`s triggered on `window` do not have an `.error` property. One of such is the `ResizeObserver loop completed with undelivered notifications.` error, caused by `ResizeObserver` ([Details](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver#observation_errors)). These errors are not handled properly by the dev overlay, as can be seen in the attachments. They also are console logged as `null`.

## What is the new behavior?
`ErrorEvent` with no `.error` property are passed directly to the dev overlay and handled there as good as possible (sadly they do not include a stacktrace). Also instead of console logging `null`, such `ErrorEvents` are logged directly.

## Other information
Before fix:
Bildschirmfoto vom 2024-10-05 16-32-59

After fix:
Bildschirmfoto vom 2024-10-05 16-32-00

### DIFF
--- a/.changeset/long-pumpkins-behave.md
+++ b/.changeset/long-pumpkins-behave.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/start": patch
+---
+
+Show proper dev overlay for ErrorEvent with no error property.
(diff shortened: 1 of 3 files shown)


ryansolid:
Thanks

PR #1650, opened 2024-10-08
Remove map and createStore from mount.ts

thetarnav:
I don't know if I'm missing something or is this a leftover, but the weak map is only being written to and the props object doesn't get changed, so it doesn't need to be a store.
I don't like when stores are added to the bundle for no reason.


atilafassina:
@thetarnav are we ok to close this PR? Or are you currently able to find it in a client bundle?

thetarnav:
I'm fine
I just noticed this during my unfinished hackathon project and decided to make a pr
If the fix is incorrect feel free to close this

PR #1665, opened 2024-10-28
Update with-vitest example to vitest  2

birkskyum:
This bumps all the deps in the with-vitest example.
It doesn't require any changes to the code apart from package.json. I check the vitest ui still run as well.


atk (approved):

ryansolid:
Thanks

PR #1683, opened 2024-12-05
Add OPTIONS to HTTP_METHODS Array

christhornham:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

API routes often receive preflight OPTIONS requests. Currently, SolidStart does not have an OPTIONS method to deal with these request, making it very hard to manage CORS.

## What is the new behavior?
Adding and OPTIONS method will allow developers to build an OPTIONS() method to specifically handle preflight requests inside API routes.

## Other information
SolidStart is great :)


atilafassina:
Hey, @christhornham 
sorry it took so long to get back to you.

I tried the PR and I don't think that's enough to have an `export const OPTIONS = () =>` in an api route. Were you able to get this working?

---

Moreover, I wanted to talk to you about the use-case for this. I was under the expectation `OPTIONS` request would happen from the middleware, like:

```ts
export async function CORS(event) {
   if (event.req.method === "OPTIONS) {
      // add headers!
      return new Response()
   }
}
```

this could be in an `onRequest` handler. From the event you could also check the appropriate paths you'd be ok with receiving an OPTIONS request to, if that's the case. But, imho, these preflight request are not determined granularly per endpoint, they're general for the whole API.

Please let me know your thoughts. I'm happy to help you push this one through if there's a use-case I'm not seeing.

christhornham:
@atilafassina ,

Thank you for reviewing my PR.

You're correct that placing the `OPTIONS` handler in middleware would handle preflight requests globally. However, that approach has a downside in my specific use case.

### My Use Case:

**1. Web App-Centric Design:**

My primary projects are web applications built with SolidStart that rely heavily on server functions. This design pattern eliminates the need to create REST API endpoints for most server interactions, which works great for web apps.

**2. Mobile Integration:**

I've expanded my boilerplate to support mobile apps built with CapacitorJS. These apps, deployed as CSR frontends, share the same backend as the web app.
- While the web app uses server functions, the mobile app must interact with the backend via API endpoints.
- Typically, I only need a couple dozen API endpoints for mobile-specific functionality.

### Why Granular `OPTIONS` Handling Matters:

**1. When the `OPTIONS` handler is placed in middleware:**
- Every Request is Checked: Even requests unrelated to the few API endpoints exposed for mobile use will unnecessarily run the `(event.request.method === "OPTIONS")` check.
- Performance: For a web app where server functions dominate, this adds avoidable overhead to every request. This overhead might be negligible, but it's unnecessary.

**By enabling `OPTIONS` handlers directly on API routes, I can:**
- Limit preflight request handling to the few API endpoints I expose.
- Optimize performance for the web app.

I understand my use case may not be typical, but for developers like myself using SolidStart for diverse projects, this change would be helpful.

If you have further questions, let me know.

Thank you!

atilafassina:
thanks for the nice response @christhornham . 
I tried testing your changes and I could not trigger the OPTIONS handler.

We're looking into before merging it.
There may be something to adjust upstream before this can work...

atilafassina (approved):

Issue #1679, opened 2024-11-27
[Bug?]: Error: Client-only API called on the server side. When using npm run dev on a brand new solidstart project


Garrett-Floyd:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

On running npm run dev for new solid start project (using basic template) throws a bunch of errors about trying to use client side only features on server side. Old projects run fine.

### Expected behavior 🤔

SolidStart project should start on local host 3000.

### Steps to reproduce 🕹

Steps:
1. npm init solid@latest
2. Project Name: datePicker
3.  Is this a SolidStart project? Yes
4. Which template would you like to use? basic
5. Use Typescript? No
6. Project successfully created! 
7. cd datePicker
8. npm install
9. npm run dev
10. > dev
> vinxi dev

vinxi v0.4.3
vinxi starting dev server

 WARN  No valid compatibility date is specified.                                                      nitro 9:59:42 PM

ℹ Using 2024-04-03 as fallback.                                                                      nitro 9:59:42 PM
  Please specify compatibility date to avoid unwanted behavior changes:
     - Add compatibilityDate: '2024-11-26' to the config file.
     - Or set COMPATIBILITY_DATE=2024-11-26 environment variable.

  ➜ Local:    http://localhost:3000/
  ➜ Network:  use --host to expose
11. ctrl+click on http://localhost:3000/
12.  command line updates to:

> dev
> vinxi dev

vinxi v0.4.3
vinxi starting dev server

 WARN  No valid compatibility date is specified.                                                      nitro 9:59:42 PM

ℹ Using 2024-04-03 as fallback.                                                                      nitro 9:59:42 PM
  Please specify compatibility date to avoid unwanted behavior changes:
     - Add compatibilityDate: '2024-11-26' to the config file.
     - Or set COMPATIBILITY_DATE=2024-11-26 environment variable.

  ➜ Local:    http://localhost:3000/
  ➜ Network:  use --host to expose

9:59:52 PM [vite] page reload vinxi/routes
9:59:52 PM [vite] page reload vinxi/routes (x2)
9:59:52 PM [vite] page reload vinxi/routes (x3)
9:59:54 PM [vite] Error when evaluating SSR module /node_modules/@solidjs/router/dist/index.js:
|- Error: Client-only API called on the server side. Run client-only code in onMount, or conditionally run client-only component with <Show>.
[...]

Issue #1670, opened 2024-11-02
[Bug?]: No valid compatibility date is specified.  


OrJDev:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

Seems like some commit broke something, using

```
    "@solidjs/router": "^0.15.0",
    "@solidjs/start": "^1.0.10",
    "autoprefixer": "^10.4.19",
    "postcss": "^8.4.38",
    "solid-js": "^1.9.2",
    "tailwindcss": "^3.4.3",
    "vinxi": "^0.4.3"
```

when running `pnpm dev` you get this warning:

```
 WARN  No valid compatibility date is specified.                                                                             nitro 1:12:01 AM

ℹ Using 2024-04-03 as fallback.                                                                                             nitro 1:12:01 AM
  Please specify compatibility date to avoid unwanted behavior changes:
     - Add compatibilityDate: '2024-11-03' to the config file.
     - Or set COMPATIBILITY_DATE=2024-11-03 environment variable.
```

this is a warning i got in a brand new SolidStart app + Tailwind using the Solid cli to create this app.

<img width="823" alt="Screenshot 2024-11-03 at 1 13 34" src="https://github.com/user-attachments/assets/b38b84df-b477-4c18-8561-074458b7a916">

### Steps to reproduce 🕹

Steps:

1. Create New SolidStart + Tailwind App
2. Install Deps
3. Run the dev server

notKamui:
This seems like a vinxi issue as it is happening to me too with tanstack-start, which uses vinxi

karimmaassen:
I have the same issue. 

```
    "@solidjs/start": "^1.0.10",
    "firebase-admin": "^12.7.0",
    "firebase-functions": "^6.1.0",
    "firebase-functions-test": "^3.3.0",
    "solid-js": "^1.9.2",
    "vinxi": "^0.4.3"
```

OrJDev:
> I have the same issue. 
> 
> 
> 
> ```
> 
>     "@solidjs/start": "^1.0.10",
> 
>     "firebase-admin": "^12.7.0",
> 
>     "firebase-functions": "^6.1.0",
> 
>     "firebase-functions-test": "^3.3.0",
> 
>     "solid-js": "^1.9.2",
> 
>     "vinxi": "^0.4.3"
> 
> ```

Yes its just a vinxi (nitro) issues @nksaraf

ocombe:
[...]

PR #1685, opened 2024-12-06
Move docs from root package to /docs package

birkskyum:
The docs are currently mixed into the root package, and it creates a number of challenges. This moves the docs into it's own package, using the latest published version of start/vinxi, instead of the main branch, similar to how the examples work.


atilafassina (approved):
Great work!!

another thing that we can probably adjust is hoisting some of the dependency version to the `pnpm-workspace.yaml` with a catalog, so packages like `vinxi`, `solidjs`, and `typescript` have the same version throughout the monorepo.

I think it's better suited for a follow-up PR though to avoid scope creeping. Great initiative!

birkskyum:
@atilafassina , does the wrangler.toml look like you'd expect it to?

I think it's not possible to pin the packages in the workspace, because we use different versions. For instance, we have to bump and publish both vinxi, then solid-start, and only then mediakit/auth, and only then we can update all three packages in the with-authjs example

birkskyum:
This PR is makes makes it possible to resolve the two [most reacted-to issues](https://github.com/solidjs/solid-start/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc) of solid-start:
- https://github.com/solidjs/solid-start/issues/1679
- https://github.com/solidjs/solid-start/issues/1670

@atilafassina  Is it a Ryan-only task to land this docs-refactor?

atilafassina:
> It would be convenient, but I'm not sure we can pin the versions in root, because we might need to use different versions at the same time with our current setup

the catalog wouldn't forces to pin, it's not an override. We just define the "default" version.
But let's leave it to another PR... I thought this was merged already.

---

I don't have access to the Cloudflare deploy, I'll check that with @davedbase.
And maybe (later) we move it to Netlify so I have access to debug the deploy on my own if needed

PR #1686, opened 2024-12-06
Add clean:all command

birkskyum:
There are many node_modules folders, and the current clean:root only takes out one.

The folders are located in:
16 examples
2 packages
1 root

Potentially more will come, such as a ./docs/node_modules, or additional examples

This cleans all of them, root and nested, and makes that the default behavior.


ryansolid:
Thanks

PR #1695, opened 2024-12-19
Granular clean commands, build commands, util for using workspace package in examples

birkskyum:
This split the clean command to examples, packages, root - to more granularly clean up `.vinxi/`, `.output/`, `node_modules/`, `dist/` 

Also, since the clean:root command removes the `./node_modules/` where rimraf is currently installed, I found it was better to use `pnpx rimraf` to always allow running the clean commands

### DIFF
--- a/util/use-workspace-package-in-examples.js
+++ b/util/use-workspace-package-in-examples.js
@@ -1,0 +1,34 @@
+import { existsSync, readdirSync, readFileSync, statSync, writeFileSync } from 'fs';
+import { dirname, join } from 'path';
+import { fileURLToPath } from 'url';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+
+const examplesDir = join(__dirname, '../examples');
+
+// Get all directories in examples folder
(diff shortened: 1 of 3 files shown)


util/use-workspace-package-in-examples.js
```diff
@@ -0,0 +1,34 @@
...
+const __dirname = dirname(__filename);
+
+const examplesDir = join(__dirname, '../examples');
+
+// Get all directories in examples folder
+const examples = readdirSync(examplesDir)
+  .filter(file => statSync(join(examplesDir, file)).isDirectory());
+
+// Process each example's package.json
+examples.forEach(example => {
```
brenelz, line 15: what exactly does this file do?

brenelz:
Some conflicts to resolve

util/use-workspace-package-in-examples.js
birkskyum: it sets the `@solidjs/start` package to `workspace:*` in all the examples, so they can be used for local testing.

atilafassina (approved):

PR #1699, opened 2024-12-20
Skip typecheck when changing ignored files

atilafassina:
PRs to docs are unable to merge because typecheck CI is not skipping and not running when the paths are ignored.


ryansolid:
Thanks

Issue #1673, opened 2024-11-07
[Bug]: Actions with latest `netlify` preset return 500 due to `Response body object should not be disturbed or locked`


serhalp:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

The page renders "⚠️ Oops, something went wrong".

This is because the Action's `POST /_server` returns a 500.

This error is logged on the server:

```
[nitro] [request error] [unhandled] Response body object should not be disturbed or locked
```

### Expected behavior 🤔

The page [should render "✅ Correct!"](https://github.com/serhalp/repro-solidstart-nitro-response-body-disturbed/blob/d5e2c1f59a5c096e751ccef698222cd017374e21/src/routes/index.tsx#L34) because [the Action returns `true`](https://github.com/serhalp/repro-solidstart-nitro-response-body-disturbed/blob/d5e2c1f59a5c096e751ccef698222cd017374e21/src/lib/api.ts#L9)

This works as expected with:
- `vinxi dev` (doesn't exercise the `netlify` preset)
- `ntl dev` (doesn't exercise the `netlify` preset)
- using the `netlify-edge` preset instead of `netlify`
- removing [this config](https://github.com/serhalp/repro-solidstart-nitro-response-body-disturbed/blob/d5e2c1f59a5c096e751ccef698222cd017374e21/app.config.ts#L5-L8) or otherwise reverting to the `netlify-legacy` preset

### Steps to reproduce 🕹

Steps:

with https://github.com/serhalp/repro-solidstart-nitro-response-body-disturbed (`pnpm i`) and netlify-cli (`npm i -g netlify-cli`)

1. `ntl serve`
2. click "click me"

or

1. create a Netlify site and link it (e.g. `ntl init`)
2. `ntl deploy --build`
3. click "click me"

I've also deployed it here: https://672c1abc039334ad67383d7e--repro-solidstart-response-disturbed.netlify.app/.

### Context 🔦

Seems to be similar to https://github.com/solidjs/solid-start/pull/1577...?

[...]

PR #1705, opened 2025-01-03
fix: handle request body streaming with latest netlify preset

serhalp:
See https://github.com/solidjs/solid-start/issues/1673.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

As described in https://github.com/solidjs/solid-start/issues/1673, submitting a Solid Action with the latest `netlify` nitro preset results in a 500 response from the `POST /server` with this error:

```
[nitro] [request error] [unhandled] Response body object should not be disturbed or locked
```

You can see this in the repro repo I shared in the issue or by clicking the button at https://672c1abc039334ad67383d7e--repro-solidstart-response-disturbed.netlify.app/.

## What is the new behavior?

Solid Actions work with the latest `netlify` nitro preset (i.e. a dummy handler results in a 200).

You can see this by linking this change into the repro repo I shared in the issue.

## Other information

I... admittedly don't really know what I'm doing here. This solution seems iffy, but it does fix Actions on the latest Netlify preset, so it may be worth it for now?

### DIFF
--- a/packages/start/src/runtime/server-handler.ts
+++ b/packages/start/src/runtime/server-handler.ts
@@ -128,7 +128,9 @@ async function handleServerFunction(h3Event: HTTPEvent) {
     // This should never be the case in "proper" Nitro presets since node.req has to be IncomingMessage,
     // But the new azure-functions preset for some reason uses a ReadableStream in node.req (#1521)
     const isReadableStream = h3Request instanceof ReadableStream;
-    const isH3EventBodyStreamLocked = isReadableStream && h3Request.locked;
+    const hasReadableStream = h3Request.body instanceof ReadableStream;
+    const isH3EventBodyStreamLocked =
+      (isReadableStream && h3Request.locked) || (hasReadableStream && h3Request.body.locked);
     const requestBody = isReadableStream ? h3Request : h3Request.body;
 
     if (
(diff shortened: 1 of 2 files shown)


packages/start/src/runtime/server-handler.ts
```diff
@@ -128,15 +128,17 @@ async function handleServerFunction(h3Event: HTTPEvent) {
     // This should never be the case in "proper" Nitro presets since node.req has to be IncomingMessage,
     // But the new azure-functions preset for some reason uses a ReadableStream in node.req (#1521)
     const isReadableStream = h3Request instanceof ReadableStream;
-    const isH3EventBodyStreamLocked = isReadableStream && h3Request.locked;
+    const hasReadableStream = h3Request.body instanceof ReadableStream;
+    const isH3EventBodyStreamLocked =
+      (isReadableStream && h3Request.locked) || (hasReadableStream && h3Request.body.locked);
```
serhalp, line 133: in other words, this was handling a case where `h3Request` is a locked `ReadableStream` but in the netlify case `h3Request.body` was a locked `ReadableStream... so I just added handling for that too

ryansolid: Nice seems simple and not that impacting. Thank you for taking the time to look at this.

PR #1715, opened 2025-01-16
refactor: adopt @tanstack/server-functions-plugin

### LINKED ISSUE #1735: Refactor: Replace @vinxi/server-functions with @tanstack/server-functions-plugin

brenelz:
Resolves #1735 

Replace @vinxi/server-functions with @tanstack/server-functions-plugin


tannerlinsley:
This is insanely cool.

birkskyum:
@brenelz , to make the CI green for the docs build, simply move the `external` from:

From: server.rollupConfig.external
To: vite.build.rollupOptions.external


Or leave it both places:

```ts
// docs/app.config.ts

import { defineConfig } from "@solidjs/start/config";
import tailwindcss from "tailwindcss";
import { config } from "vinxi/plugins/config";

export default defineConfig({
  // experimental: { islands: true },
  server: {
    preset: "cloudflare_module",
    alias: {
      "_mime": "mime/index.js"
    },
    rollupConfig: {
      external: ["__STATIC_CONTENT_MANIFEST", "node:async_hooks"]
    }
  },
  vite: {
    build: {
      rollupOptions: {
        external: ["__STATIC_CONTENT_MANIFEST", "node:async_hooks"]
      }
    },
    plugins: [
      config("tailwind", {
        css: {
          postcss: {
            plugins: [tailwindcss]
          }
        }
      } as any)
    ]
  }
});
```

atilafassina:
That's awesome!!
Thanks a lot for this, @brenelz 🏆

Meanwhile I'm working on improving the test story around here so we can merge this more confidently.
Once the CI is green, let's make sure we add a Changeset and then we should be in good shape for reviews / merge. :)

birkskyum:
Since we don't have a test suite, the examples we can run manually that does use `"use server";` that we for a start can verify still works are:
- [ ] hackernews
- [ ] todomvc
- [ ] notes
- [ ] with-auth
- [ ] with-drizzle
- [ ] with-prisma

katywings:
I am wondering how easy it would be to test this out in one of my projects via npm link 🙈, or is there a chance we could publish this on some special name/tag to give it a try?

birkskyum:
@katywings , there is the pkg.pr.new link [here](https://github.com/solidjs/solid-start/pull/1715#issuecomment-2594398804) - can you use that?

brenelz:
@birkskyum I was just going to say the same thing. I haven't actually used it before but seems cool

atilafassina:
Just merged #1719 
if we solve the conflicts on this PR we can pull `main` and it would be more straightforward to add tests.

birkskyum (dismissed):
E2E test passing. LGTM

.changeset/violet-steaks-mate.md
```diff
@@ -0,0 +1,5 @@
+---
+"@solidjs/start": minor
```
atilafassina, line 2: I was kinda expecting this one to be a **patch** since it's not changing any API nor is it adding a new feature.

wdyt?

katywings:
Testing this out right now, so far it doesn't work at all for me, most likely because it seems to have a completely different approach to decide which dependencies to externalize. Still figuring it out the details though. You gonna get a second analysis soon 😅

katywings:
Here is a very basic reproduction of the missing externalization, this works in `dev` but fails to `npm run build`.
https://stackblitz.com/edit/github-vts2k1cp?file=src%2Froutes%2Findex.tsx

katywings:
I wonder if this is a problem in our implementation or a general thing with the Tanstack Server Functions plugin 🤔 -  afaik @tannerlinsley mentioned that the Tanstack plugin actually should externalize more heavily than Vinxi and not less - so I assume this must be a problem on our side 🤔.

packages/start/config/index.js
```diff
@@ -37,6 +36,35 @@ function solidStartServerFsRouter(config) {
...
     );
 }
+const SolidStartServerFnsPlugin = createTanStackServerFnPlugin({
+  // This is the ID that will be available to look up and import
+  // our server function manifest and resolve its module
+  manifestVirtualImportId: "solidstart:server-fn-manifest",
+  client: {
+    getRuntimeCode: () =>
+      `import { createServerReference } from "${normalize(
+        fileURLToPath(new URL("../dist/runtime/server-runtime.js", import.meta.url))
```
katywings, line 46: Is it intended to import a `server-runtime` in the client plugin 🤔? Or is this just a suboptimal naming 😅?

katywings: (Nevermind, seems to be the naming 😅)

birkskyum: tried to track through it and came to the same confusing conclusion

birkskyum:
There are externalization tests added to the test suite, so this PR should be rebased, and passing those.

birkskyum:
@katywings , next time you have a chance to check it, let us know if it fares better for you now with the externalization fixed.

katywings:
@brenelz 45f0b7b indeed fixed the build in my project and I was able to build the project without any changes in the vite config 🥳. My time today is limited but I will go through the project tomorrow and test out the different features related to server functions and give you another report.

packages/start/src/runtime/server-handler.ts
```diff
@@ -78,19 +80,28 @@ async function handleServerFunction(h3Event: HTTPEvent) {
...
+    functionId = url.searchParams.get("id");
     name = url.searchParams.get("name");
-    if (!filepath || !name) throw new Error("Invalid request");
+    if (!functionId || !name) throw new Error("Invalid request");
+  }
+
+  const serverFnInfo = serverFnManifest[functionId];
+  let fnModule: undefined | { [key: string]: any };
+
+  if (process.env.NODE_ENV === "development") {
```
katywings, line 96: Why do we have to load the server function via vinxi during development 🤔, does `serverFnInfo.importer` not work in dev?

birkskyum (approved):

packages/start/src/runtime/server-handler.ts
brenelz: I'm not entirely sure. I kinda copied this over from Tanstack. If you want I can see if its not needed

brenelz: It seems like its needed cause otherwise it errors

katywings: Oh I guess this really must be some limitation then, there is a comment about this in the tanstack sources:
https://github.com/TanStack/router/blob/266f5cc863cd1a99809d1af2669e58b6b6db9a67/packages/start-server-functions-handler/src/index.tsx#L81

Since this is not mentioned in the README of the tanstack server functions plugin, I recommend adding a comment referencing the original tanstack code, so that we quickly can compare the integration at a later time 🤔

birkskyum:
@atilafassina , I tried to follow the guide at [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/), but the PR title check doesn't pass... what am I doing wrong?

katywings (approved):

adamstringer:
I'm sure you guys are already aware but this is a breaking change when declaring server functions from within components returning the error "Server Functions cannot be nested in other blocks or functions". Might want to mention this in the release notes?

joeygrable94:
> I'm sure you guys are already aware but this is a breaking change when declaring server functions from within components returning the error "Server Functions cannot be nested in other blocks or functions". Might want to mention this in the release notes? 

Why are you declaring a server function from within a component scope? Ideally, you are writing server function code independently from component rendering so keep your code easier to maintain and test.

theturtle32:
This also appears to have broken support for async generators as server functions #1886

birkskyum:
@theturtle32 can you add a test for that in our server function test suite?

PR #1719, opened 2025-01-22
Add cypress server-function test

birkskyum:
This adds cypress, and a small test that simply prints to the DOM the result of isServer when run in a client component, and within a server function.

It's executed manually in two steps right now:

1. Go to `/tests/server-function` and run `pnpm dev` to launch the server of this barebones project

2. Run in the repo root `pnpm run cy:run` (or for UI run `cy:open`)

I've added a github workflow too

cc @atilafassina , @brenelz


atilafassina (approved):
Tests!!!! 🎉

PR #1720, opened 2025-01-23
Workspace versions

atilafassina:
## PR Type

**What kind of change does this PR introduce?**
infrastructure changes

**What is the new behavior?**

as we add more packages and consume the same ecosystem dependencies, we can lock the versions from the tests to the ones used in the actual package. 

--

@birkskyum


birkskyum (approved):
Nice! We should probably either also use these for the docs, or move the docs off workspace packages.

PR #1722, opened 2025-01-24
Add externalization test to server function suite 

birkskyum:
This makes sure the server function e2e tests also cover the ability to externalize modules.

As mentioned here: 
- https://github.com/solidjs/solid-start/pull/1715#issuecomment-2612592366

I've also split the suite into 4 tests:

- it("should have isServer false in the client")
- it("should have isServer true in the server function")
- it("should externalize node builtin in server function")
- it("should externalize npm module in server function")
  

## PR Checklist

- [X] Tests for the changes have been added (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [ ] Feature
- [X] Tests
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

### DIFF
--- a/tests/server-function/src/app.tsx
+++ b/tests/server-function/src/app.tsx
@@ -1,40 +1,24 @@
-import { join } from 'path';
-import { createEffect, createSignal } from "solid-js";
-import { isServer } from "solid-js/web";
-
+import { MetaProvider, Title } from "@solidjs/meta";
+import { Router } from "@solidjs/router";
+import { FileRoutes } from "@solidjs/start/router";
+import { Suspense } from "solid-js";
 import "./app.css";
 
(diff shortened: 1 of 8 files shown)


tests/server-function/src/app.tsx
```diff
@@ -1,19 +1,34 @@
+import { join } from 'path';
```
katywings, line 1: I wonder if it would make sense to also include a similar test, but with an npm library instead of a built-in 🤔. Bundlers often have different logic/assumptions when it come to built-ins vs npm libraries. By testing both import types we could probably catch more potential issues.

birkskyum: I'll add a lodash example

Issue #1693, opened 2024-12-17
[Bug?]: @tailwindcss/vite v4 breaking after dev server restart


birkskyum:
### Current behavior 😯

In tailwind 4, which is still only in beta (but it would be good to fix this before they hit stable), there are now two ways to install tailwind.
- Traditional: Using a Postcss plugin - works as expected
- New: Using a Vite plugin (works in astro, but dev server struggle in solid-start)

The vite plugin works as expected when making a prod build, and for the dev server on a clean reload, but if the dev server is restart in solid-start, the style will be broken. To get it working again, it will be necessary to do either of the following:
- delete `node_modules/.vinxi` and then restart dev server
- modify/re-save `./src/app.css` - triggering this file reload is enough to make it work again

With Astro (v5), there are no issues with restarting the dev server using @tailwindcss/vite with Solid.
With TanStack Start (vinxi 0.4.3/0.5.1), there are no issues either.

### Expected behavior 🤔

I expected the tailwind plugin to load correctly on dev server restart.

### Steps to reproduce 🕹

There's a repo here with multiple implementations:
https://github.com/birkskyum/tailwind-vite-repro

- Astro with Vite plugin (and solid) - working 
- TanStack Start with Vite plugin (and vinxi) - working
- Solid Start with Postcss plugin - working
- Solid Start with Vite plugin - breaking after dev server restart

I can reproduce this issue with vinxi 0.4.3 / 0.5.0 / 0.5.1 and overrides with vite 5/6

The working output looks like this:
<img width="284" alt="Screenshot 2024-12-17 at 04 18 59" src="https://github.com/user-attachments/assets/baa78bd5-009d-46df-811f-0114b57ccee2" />

The broken output will look like this:

<img width="184" alt="Screenshot 2024-12-17 at 04 19 15" src="https://github.com/user-attachments/assets/0694c7ec-b5a5-4a07-8644-84e90d8754bb" />

atilafassina:
I'm not sure I followed all the way.

So Tanstack Start is on Vinxi 0.5.1 and Vite 5 -- and that works, right?
And we're on Vinxi 0.5.1 and Vite 6 -- which doesn't work.

Is that the takeaway? Should we pin down Vite to 5 would the issue goes away for us as well?

birkskyum:
Because now Tailwind 4 is has been stable released:
https://tailwindcss.com/blog/tailwindcss-v4

So we better make sure it works with SolidStart

[...]

PR #1724, opened 2025-01-25
Move landing page out of monorepo root

atilafassina:
this PR is the first of 2 moving the landing page into its own workspace and out of Cloudflare.
When this hits `main`, I can setup the deployment in the Netlify dashboard and we can point the domain there with no downtime.


birkskyum:
Why aren't these assets added to the /docs folder

atilafassina:
> Why aren't these assets added to the /docs (landing page) folder
> 
> The landing page is already moved out of the repo root? https://github.com/solidjs/solid-start/tree/main/docs

it isn't. It's still using the `tsconfig.json` from the root and `wrangler.toml` is at the root.
That's why there's some issues with the types going into the project and there its dependencies get a bit intertwined with the monorepo dependencies.

I can't move the `/docs` project inside an isolated workspace because I don't have access to the Cloudflare deploy, and since we want to move it to the Netlify anyway, I figured making the move like this would be easier.

birkskyum:
> I can't move the /docs project inside an isolated workspace because I don't have access to the Cloudflare deploy

sigh... @ryansolid can @atilafassina get access to Cloudflare deploy, so we don't need to jump hoops?

atilafassina:
@birkskyum  We still need the concurrent deploys to reroute the main domain without downtime.

I also feel moving co-locating the workspaces is better. So instead of `/docs` or `/landing-page` they'd within a `/packages/landing-page` or `/workspaces/landing-page` to make a clear distinction from workspaces and generic/shared files.

(probably same for /tests, now that we're talking)

also, @davedbase is the Cloudflare master.

birkskyum:
> @birkskyum We still need the concurrent deploys to reroute the main domain without downtime.

Yes, but we can just disable the auto-deploy in cloudflare now, move the package to /packages/landing-page update the preset, and deploy to netlify, and only then update the DNS to point to netlify.

atilafassina:
sure. I'll remove the CI for deployment and we do that

birkskyum:
There is some cleanup in the 

- pnpm-workspace of the "docs" package
- package.json of the docs:dev etc. commands, if the package name is changed

packages/landing-page/ui.preset.ts
```diff
@@ -1,7 +1,6 @@
-import { fontFamily } from "tailwindcss/defaultTheme";
+import { type Config } from "tailwindcss";
```
birkskyum, line 1: ah, it's moved to .ts. Nice!

A side note - we have a bit of a challenge with this
- https://github.com/solidjs/solid-start/issues/1693

Because now Tailwind 4 is has been stable released:
https://npmjs.com/package/@tailwindcss/vite

So we better make sure it works with SolidStart

birkskyum: I'd really like us to resolve this asap:
- https://github.com/solidjs/solid-start/issues/1693

birkskyum (approved):
Few [nits](https://github.com/solidjs/solid-start/pull/1724#issuecomment-2613944186), but otherwise this looks ready.

PR #1729, opened 2025-01-27
chore: Contributing Compliance Workflow

### LINKED ISSUE #1730: Improve release flow and changelog generation to leverage automation
### Duplicates

### Latest version

### Summary 💡

This is an umbrella issue to automate some aspects of our CI

Generate changelog and release tasks
1. Comment on PR which version that PR has landed
2. Enforce every PR has a correlated issue

### Examples 🌈

_No response_

### Motivation 🔦

this will make sure contributions are streamlined and we can have a well-defined process to receive them.

atilafassina:
This sets up an automatic workflow to check if a PR is according to the contributing guidelines

- PR title is formatted according to conventional commits: in preparation for a release workflow that will generate releases in GH
- PR body refers to an issue: to make sure proper discussion and considerations were had before writing the code
- PR does not originate from protected branches (in our case, only `main`)

This workflow comes from Create-TypeScript-App by @joshuakgoldberg. Following this standard we can then also use the `post-release` workflow and the `release` workflow which will automate changes and feedback to PR authors about which release carries their code.

closes #1730

---

you can see in the comments below how the comment would work :)

### DIFF
--- a/.github/workflows/compliance.yml
+++ b/.github/workflows/compliance.yml
@@ -1,0 +1,28 @@
+name: Contributing Compliance
+
+on:
+  pull_request:
+    branches:
+      - main
+    types:
+      - edited
+      - opened
+      - reopened
+      - synchronize
+
+permissions:
+  pull-requests: write
+
+jobs:
+  compliance:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: mtfoley/pr-compliance-action@main
+        with:
+          body-auto-close: false
+          ignore-authors: |-
+            allcontributors
+            allcontributors[bot]
+            renovate
+            renovate[bot]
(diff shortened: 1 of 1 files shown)


birkskyum (approved):

.github/workflows/compliance.yml
```diff
@@ -0,0 +1,28 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: mtfoley/pr-compliance-action@main
+        with:
+          body-auto-close: false
+          ignore-authors: |-
+            allcontributors
+            allcontributors[bot]
+            renovate
+            renovate[bot]
```
birkskyum, line 27: should dependabot be added here?

PR #1748, opened 2025-01-31
Tailwind CSS bug repro

DaniGuardiola:
**DO NOT MERGE**, this is a bug repro.

Related issue: #1749

This branch reproduces a bug that causes the css output of Tailwind CSS classes that contain the `&` character to transform it into `&amp;`, which breaks it. For example, a class using a "dark theme" variant like this:

```css
.dark\:bg-red-500 {
  &:where(.dark, .dark *) {
    background-color: var(--color-red-500);
  }
}
```

This is actually outputted like this:

```css
.dark\:bg-red-500 {
  &amp;:where(.dark, .dark *) {
    background-color: var(--color-red-500);
  }
}
```

The effect is that the browser cannot parse the contents of the selector, so that class doesn't work and the color is not applied.

## How to run repro

1. Clone branch
2. `pnpm i`
3. Go to `examples/with-tailwindcss`
4. `pnpm dev`
5. See bottom section of the page.
6. See `src/routes/index.tsx` for the code.
7. Dark variant is specified in `src/app.css`

### DIFF
--- a/examples/with-tailwindcss/README.md
+++ b/examples/with-tailwindcss/README.md
@@ -28,3 +28,29 @@ npm run dev -- --open
 Solid apps are built with _presets_, which optimise your project for deployment to different environments.
 
 By default, `npm run build` will generate a Node app that you can run with `npm start`. To use a different preset, add it to the `devDependencies` in `package.json` and specify in your `app.config.js`.
+
+## Tailwind CSS bug
+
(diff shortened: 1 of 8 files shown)


atilafassina:
@DaniGuardiola I believe this issue is solved by https://github.com/solidjs/solid-start/pull/1787

Let us know if you think it isn't. For now I'm closing this :)

DaniGuardiola:
@atilafassina it is not fixed. This happens independently of whether vite or postcss is used.

PR #1750, opened 2025-01-31
test: add treeshaking, improve CI

### LINKED ISSUE #1751: Add Vitest to `/tests` workspace

atilafassina:
- **test: add treeshaking test**
- **infra: move tests inside workspaces and add unit to CI**

closes #1751


birkskyum (commented):
are all the screenshots needed?

packages/start/package.json
```diff
@@ -65,7 +65,8 @@
   "dependencies": {
     "@vinxi/plugin-directives": "^0.5.0",
     "@vinxi/server-components": "^0.5.0",
-    "@tanstack/server-functions-plugin": "^1.97.2",
+    "@solidjs/start": "^1.0.11",
```
birkskyum, line 68: It looks like the start package depend on itself

.github/workflows/server-function.yml
```diff
@@ -1,49 +0,0 @@
-name: ci
-
-on:
-  pull_request:
-    branches: [main]
```
birkskyum, line 5: why is this file removed?

atilafassina: renamed so it's not only about server-functions only. Git tracked as a removal/addition because I changed too much inside to run unit and e2e from the same built artifact instead of building things only once.

PR #1756, opened 2025-02-03
Vite 6 bump

birkskyum:
This is a version of solid-start that's built with the latest vinxi packages / vite 6.

It's not compatible with vinxi 0.4.3, thus I've added a peerDependency on 0.5.1 (and since there won't be more vinxi it's pinned).

We can use the CR flow in this PR to validate it works in our projects.


katywings (approved):
Looking good to me ❤️! P.S. I tested it with Vite 6.1 though 😅

PR #1760, opened 2025-02-03
remove conventional commit

birkskyum:
I find this check to be really annoying. I spend time restarting it a lot, and the motivation behind it appear somewhat vague due to the lack of monorepo compatible with conventional commits.

It'll have to be removed from the branch protection to let the CI pass.


birkskyum:
closed in favor of
- #1758

PR #1769, opened 2025-02-04
chore: update changeset cli

birkskyum:
minor bump of the cli package


birkskyum:
@atilafassina , this is merely a minor bump of `@changesets/cli`. I don't understand why it would break the netlify deployments

PR #1773, opened 2025-02-06
fix: move getServerFunctionMeta export to @solidjs/start

### LINKED ISSUE #1772: [Bug]: getServerFunctionMeta creates circular import
### Duplicates

### Latest version

### Current behavior 😯

If you import `getServerFunctionMeta` from `@solidjs/start/server` in a file related to a route, this creates a circular import, because `StartServer` also resides in `@solidjs/start/server`. Most of the time vite seems to be able to prevent this from breaking the app. But in some cases (especially during `vite dev`) it fails and the circular import will lead to `getServerFunctionMeta` being undefined.

### Expected behavior 🤔

You should be able to import `getServerFunctionMeta` without introducing a circular import.

katywings:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1772

`getServerFunctionMeta` is exported in `@solidjs/start/server`. Importing it can lead to circular import issues, because `StartServer` also resides in that entry. During such a circular import issue, `getServerFunctionMeta` will be undefined, resulting in an `getServerFunctionMeta is not a function` error.

## What is the new behavior?

- The old `getServerFunctionMeta` export in `@solidjs/start/server` is deprecated.
- A new `getServerFunctionMeta` export is added to `@solidjs/start`, which fixes the circular import.

## Other information

PR for changing the `getServerFunctionMeta` docs entry: solidjs/solid-docs#1064

A similar change had to be made in the past related to `FileRoutes` (#1365).

This PR includes a new test for `getServerFunctionMeta`. If you run this test with the old `getServerFunctionMeta` from `@solidjs/start/server` the test fails - because of the circular import issue. The resulting cypress screenshots:

server-function -- should have an id of type string in the server function meta - nested (failed)
server-function -- should have an id of type string in the server function meta - toplevel (failed)

### DIFF
--- a/.changeset/metal-ladybugs-check.md
+++ b/.changeset/metal-ladybugs-check.md
@@ -1,0 +1,7 @@
+---
+"@solidjs/start": patch
+---
+
+Move `getServerFunctionMeta` from `@solidjs/start/server` to `@solidjs/start` to fix circular import issues.
+
+The old export at `@solidjs/start/server` still exists, but is **deprecated** and will be removed in a future release.
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ node_modules
 
 .tmp
 test/playwright-report
+tests/server-function/cypress/screenshots
 
 # Temp
 gitignore
--- a/packages/start/src/index.tsx
+++ b/packages/start/src/index.tsx
@@ -1,6 +1,6 @@
 // @refresh skip
+export { default as clientOnly } from "./shared/clientOnly";
 export { GET } from "./shared/GET";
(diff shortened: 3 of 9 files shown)


ryansolid:
I mean you probably use this the most so I trust you. But the whole reason `/server` exists is to stop circular deps. I'm a little bit surprised there would be one here. But then again this has nothing to do with entry or middleware so the other places we hit circular deps probably don't apply here.

ryansolid (approved):
Ok.. sounds good to me. Keeping it there and deprecated makes sense for now as well.

katywings:
@ryansolid Thank you for your responsive feedback regarding this!

PR #1775, opened 2025-02-09
fix: return 404 when server function not found

### LINKED ISSUE #1779: [Feature?]: return 404 when server function not found
### Duplicates

### Latest version

### What is the current behavior?
Currently server functions calls are expected to always provide the correct id and name. However in our use case we have had bots that try to access older id and name combinations which results in a 500 error.

### Examples 🌈

_No response_

### What is the new behavior?
This change treats not found server functions with a 404 response so that it provides better logging results in case an older or wrong id is being accessed directly.

loetjvr:
fixes #1779

Currently server functions calls are expected to always provide the correct id and name. However in our use case we have had bots that try to access older id and name combinations which results in a 500 error.

## What is the new behavior?

This change treats not found server functions with a 404 response so that it provides better logging results in case an older or wrong id is being accessed directly.

### DIFF
--- a/packages/start/src/runtime/server-handler.ts
+++ b/packages/start/src/runtime/server-handler.ts
@@ -87,12 +87,13 @@ async function handleServerFunction(h3Event: HTTPEvent) {
   } else {
     functionId = url.searchParams.get("id");
     name = url.searchParams.get("name");
-    if (!functionId || !name) throw new Error("Invalid request");
+    if (!functionId || !name) return new Response("Server function not found", { status: 404 });
   }
 
   const serverFnInfo = serverFnManifest[functionId];
   let fnModule: undefined | { [key: string]: any };
 
+  if (!serverFnInfo) return new Response("Server function not found", { status: 404 });
   
   if (process.env.NODE_ENV === "development") {
     // In dev, we use Vinxi to get the "server" server-side router


atilafassina:
Thanks a lot for this PR, @loetjvr 
I think it's a great fix to throw a `404` instead of letting a `500` when the server-function name can't be found.

Looking at the code and our compliance tasks, I'd like to ask for a few additional things if you're ok with it:

1. Could you please create an issue describing why this is a problem (you can copy/paste from your PR description where it makes sense) and feel free to `@` me if you want.
2. Add to the PR description `fixes ###` with the issue so our automation can pick up on that
3. I think in DEV mode your error code is perfect, but how about we leave the "server function not found" message out in production? I think it's better for security that we don't offer additional hints on how the system work in this case.

Let me know your thoughts if you don't agree.

Cheers 🏆

loetjvr:
More than happy to make the changes @atilafassina
I did not see any of these requests in the contributions md so was a bit unsure what was required

birkskyum (approved):

atilafassina:
> More than happy to make the changes @atilafassina I did not see any of these requests in the contributions md so was a bit unsure what was required
> 
> Also on point number 4, what does this entail? I updated the pr title, should I update somewhere else as well?

That's entirely on us, @loetjvr 
we're still improving the process (just merged some important CI tests as requirement)

Now that your PR reached `main`, I'll make the changes needed on another one.

Thanks a lot for your work!! 🏆

PR #1777, opened 2025-02-09
chore: bump pnpm 10

birkskyum:


birkskyum:
pnpm v10 can't be found on remove servers for some reason.

atilafassina (changes requested):
more importantly, our monorepo doesn't seem to work with pnpm 10 at the moment 
https://github.com/solidjs/solid-start/pull/1741

PR #1785, opened 2025-02-10
fix: add e2e and return `404` when a server-function is not found

atilafassina:


birkskyum (approved):

birkskyum:
Okay, if I understand correctly, the change here is that it's now just a 404, but doesn't write that it's because the server function isn't found

atilafassina:
> If I understand correctly, the change here is that it's now just a 404, but doesn't write that it's because the server function isn't found (unless in development)

Exactly. This way we don't offer any hints about the system if a resource doesn't exist.
I also add a small test to prevent any regressions in the future

PR #1786, opened 2025-02-10
chore: remove duplicate changeset

birkskyum:


birkskyum:
@atilafassina , i think these fixes you make on the release pr get remove whenever it updates

birkskyum:
I think it would be a bit cleaner to land these first
- #1791
- #1786

PR #1747, opened 2025-01-31
Upgrade Tailwind CSS example to v4

DaniGuardiola:
Upgraded to v4 with the vite plugin.

**DO NOT MERGE YET**, using a nightly version until [this fix](https://github.com/tailwindlabs/tailwindcss/pull/16052) is released.

PR #1787, opened 2025-02-10
chore: update tailwind example to v4

birkskyum:
This PR adopts tailwind 4, with the tried and tested postcss config that we know work.

This is a safe way of doing things, in comparison to using the new tailwind vite plugin:
- #1748 
- #1747


atilafassina (approved):

DaniGuardiola:
Hi @birkskyum thank you.

A few observations:

- The bug originally preventing the vite plugin from working has been fixed.
- The other bug causing & to be wrongly processed is NOT related to vite, it happens independently of whether vite or postcss is used.
- The postcss method is outdated and no longer the main recommended approach (vite is instead). It's probably slower too.
- I think the example should be updated to the vite setup since the only actual vite-related bug was solved. The other issue should be fixed independently.
- In fact, I think the Tailwind example should be reverted to v3 for now, as it is currently broken and new users will get a broken starting boilerplate. Then, once fixed, it should be updated to the recommended vite setup.

PR #1791, opened 2025-02-10
chore: use workspace package for landing page

birkskyum:
Can this work, or will the manual overrides still be needed?

https://github.com/solidjs/solid-start/pull/1765/commits/824155990764a8ed212b4e12b8b137bd4a203a09


birkskyum:
@atilafassina , if we set it to workspace, will it still change it to 1.1.0?

atilafassina (approved):

PR #1802, opened 2025-02-12
patch: add function name to _server paths

### LINKED ISSUE #1795: [Feature?]: show function name in _server routes
### Duplicates

### Latest version

### Summary 💡

Currently all server function calls use _server route. It would be easier to log and debug errors if the route has more details eg _server/function-name.

As I understand anything going to _server will be processed in the same function and is just looking for id
and name search prop. So I propose we just add the function name to the url and dont change anything on the logic side.

I am not sure if function extraction is handled by Solid start or Vinxi and also if function name is the name of the function it extracts from or just random cause I am not always able to get a readable function name.

### Examples 🌈

Image

### Motivation 🔦

We find that with our logging and debugging that having everything go to the same url makes it difficult to know which function is being called and thus difficult to resolve issues or to setup proper observability.

loetjvr:
fixes #1795

Currently we dont add any path in the _server route

## What is the new behavior?

The extracted function name is now added eg _server/function-name

## Other information

This approach however seems flaky because it depends on the way that the Tanstack server function plugin is extracting and then constructing the ids which is "filepath--function-name_query"
This looks like it is prone to change and does not provide guarantees.

I did see that Tankstack plugin does have the filename but seems like its not in the returned options so it might be stripped off somewhere or I missed it.

### DIFF
--- a/packages/start/src/runtime/server-fns-runtime.ts
+++ b/packages/start/src/runtime/server-fns-runtime.ts
@@ -5,10 +5,17 @@ import { cloneEvent } from "../server/fetchEvent";
 export function createServerReference(fn: Function, id: string, name: string) {
   if (typeof fn !== "function") throw new Error("Export from a 'use server' module must be a function");
   const baseURL = import.meta.env.SERVER_BASE_URL;
+  // @tanstack/server-functions-plugin contructs the id from the filename + function name eg src_lib_api_ts--getStory_query
+  // So we extract the name by splitting on --
+  // This feels flaky and we should rather try and get the function name directly from the plugin
+  // but this requires a change in the plugin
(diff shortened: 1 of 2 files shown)


brenelz:
Yeah this could have changed when we moved to the tanstack server functions. I think the id and name changed a bit.

brenelz:
Would this solve the issue or is this different?
https://github.com/solidjs/solid-start/pull/1872

brenelz:
Do we still want to go ahead with this @loetjvr ?

loetjvr:
@brenelz I haven't been following along as to where you are in removing Vinxi, cause that is where the issue is I believe. So might be best just to close this PR as I think it will be resolved with the current roadmap.

PR #1820, opened 2025-02-20
Update types.ts

huseeiin:
add types for routes & assets in PageEvent

## PR Checklist

### DIFF
--- a/packages/start/src/server/types.ts
+++ b/packages/start/src/server/types.ts
@@ -1,81 +1,92 @@
 import type { JSX } from "solid-js";
 import { RequestEvent } from "solid-js/web";
 import { HTTPEvent } from "vinxi/http";
+import { RouteDefinition } from "@solidjs/router";
 
 // export const FETCH_EVENT = "$FETCH";
 
 export type DocumentComponentProps = {
-  assets: JSX.Element;
-  scripts: JSX.Element;
-  children?: JSX.Element;
+    assets: JSX.Element;
+    scripts: JSX.Element;
+    children?: JSX.Element;
 };
 
 export type Asset = {
-  tag: "style";
-  attrs: JSX.StyleHTMLAttributes<HTMLStyleElement> & { key?: string };
-  children?: JSX.Element;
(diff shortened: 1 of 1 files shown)


packages/start/src/server/types.ts
```diff
@@ -1,81 +1,92 @@
 import type { JSX } from "solid-js";
 import { RequestEvent } from "solid-js/web";
 import { HTTPEvent } from "vinxi/http";
+import { RouteDefinition } from "@solidjs/router";
```
birkskyum, line 4: Hmm. This challenge the router-agnostic principle of solid-start

huseeiin: if that's why it was previously `any`, then i'm ok with it remaining this way.

AlexErrant: You could make the type extendable like how Tanstack Table does with its [`meta`](https://tanstack.com/table/v8/docs/api/core/table#meta). [Relevant code.](https://github.com/TanStack/table/blob/db745afdb867511ddd04a79150f89d3894a03532/packages/table-core/src/types.ts#L118)

```ts
declare module '@tanstack/table-core' {
  interface TableMeta<TData extends RowData> {
    foo: string
  }
}
```

So a solid-start user's code could be

```ts
import { RouteDefinition } from "@solidjs/router";

export interface Route extends RouteDefinition {
    page: boolean;
    filePath: string;
    id: string;
    $component: {
        src: string;
        import: () => Promise<any>;
    };
}
```

And solid-start's types.ts would need to do something like

```ts
export interface PageEventRoutes {}

export interface PageEvent extends RequestEvent {
    manifest: any;
    assets: Asset[];
    routes: PageEventRoutes[];
   // remainder of def
}
```

Giving it an explicit named type would make it easier to extend, I think? idk haven't tested any of this, just theorycrafting.

brenelz:
Do we just want to close this one?

huseeiin:
> Do we just want to close this one?

yeah

PR #1827, opened 2025-02-24
tailwind v4 support

### LINKED ISSUE #1749: [Bug?]: Tailwind CSS v4 outputs `&` as `&amp;` which breaks the styles
### Duplicates

### Latest version

### Current behavior 😯

Tailwind CSS v4 (recently fixed to work with Solid Start) outputs `&` as `&amp;` which breaks the styles.

For example, a class using a "dark theme" variant like this:

```css
.dark\:bg-red-500 {
  &:where(.dark, .dark *) {
    background-color: var(--color-red-500);
  }
}
```

This is actually outputted like this:

```css
.dark\:bg-red-500 {
  &amp;:where(.dark, .dark *) {
    background-color: var(--color-red-500);
  }
}
```

### Expected behavior 🤔

Outputs `&` directly.

### Steps to reproduce 🕹

See repro: https://github.com/solidjs/solid-start/pull/1748

### Context 🔦

It doesn't depend on the vite setup because it also happened with the alternative PostCSS-based setup. Unsure if the issue is on Solid Start's side or Tailwind CSS's side.

Related Tailwind CSS issue: https://github.com/tailwindlabs/tailwindcss/issues/16133

atilafassina:
fixes #1749

the actual bug was fixed in DOM-Expressions, landed in newest version of `solid-js` and now we're merely bringing it to the templates.

### DIFF
--- a/.changeset/blue-geese-scream.md
+++ b/.changeset/blue-geese-scream.md
@@ -1,5 +1,0 @@
----
-"@solidjs/start": patch
----
-
-Tailwindv4 Support
--- a/.changeset/blue-geese-scream.md
+++ b/.changeset/blue-geese-scream.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/start": patch
+---
+
+Tailwindv4 Support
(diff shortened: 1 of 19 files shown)


birkskyum:
Looking at the diff, it doesn't seem like this bumps solid

PR #1830, opened 2025-02-26
Supabase Auth Example

kjrocker:
## PR Checklist

There is no complete example for how to use Supabase in SolidStart.

## What is the new behavior?

Create a "complete" example of Supabase Auth with the following features

1. A mixture of server-side and browser-side behavior
2. The Auth basics: Sign in, make a new account, sign out, forget password, reset password.
3. A way to protect authenticated pages.
4. A way to conditionally render UI based on authentication (**Still WIP**)

## Other information

This is actually my first non-trivial work with SolidStart, so I'm sure there are some non-idiomatic things I can improve. 

Some things still don't work properly, especially accessing the user (in this example, changing the navigation to show Sign In/Up when unauthenticated, and Sign Out when authenticated).

One of the biggest problems I encountered was the inability to create an isomorphic Supabase client. I had a lot of unpredictable behavior until I strictly separated the server and client code. I blame this one line in the docs:

> Note that Vinxi HTTP helpers do not treeshake so you can only import them in files that do not contain client or isomorphic code. (https://docs.solidjs.com/solid-start/advanced/request-events#nativeevent)

Since I use Vinxi functions when instantiating a Server Supabase client, the normal isomorphic pattern of `isServer ? doServerThing() : doClientThing()` didn't work. Maybe dynamic imports are the answer. This is just my theory on why I had to strictly seperate "use client" actions from "use server" actions.


atilafassina:
Hello.

First of all, sorry it took us so long to look at this.

Part of this is the issue is that we've been looking at how to better server the community with our templates and ongoing repo work. Because of that we have decided to move templates to \[solidjs/templates]\([https://github.com/solidjs/templates)](https://github.com/solidjs/templates\)). 

You're much welcome to recreate this template there, and read more on the new architecture in the discussion https://github.com/solidjs/solid-start/issues/1984

You're feedback is definitely much appreciated.

:blue\_heart:

PR #1840, opened 2025-03-01
feat: add ssr for tanstack router

brenelz:
## PR Checklist

Enables SSR for tanstack router app

### DIFF
--- a/packages/start/src/client/StartClient.tsx
+++ b/packages/start/src/client/StartClient.tsx
@@ -24,3 +24,13 @@ export function StartClient() {
     </Dummy>
   );
 }
+
+export function StartClientTanstack() {
+  return (
+    <Dummy>
+      <ErrorBoundary>
+        <App />
+      </ErrorBoundary>
+    </Dummy>
+  );
+}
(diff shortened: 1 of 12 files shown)


brenelz:
Closing as this is buggier than i thought

packages/start/src/server/handler.ts
```diff
@@ -31,18 +31,24 @@ export function getExpectedRedirectStatus(response: ResponseStub): number {
 export function createBaseHandler(
   fn: (context: PageEvent) => unknown,
   createPageEvent: (event: FetchEvent) => Promise<PageEvent>,
+  routerLoad?: (event: FetchEvent) => Promise<void>,
```
birkskyum, line 34: this new argument, it's probably best to have it last, because it's the only optional, and then the order doesn't change for `options`

brenelz: good call not sure why exactly i put it there

packages/start/src/client/StartClient.tsx
```diff
@@ -24,3 +24,13 @@ export function StartClient() {
     </Dummy>
   );
 }
+
+export function StartClientTanstack() {
```
brenelz, line 28: not sure we want a tanstack specific component? But it needs one less dummy component it seems

birkskyum: Hmm.. can we for now just move this definition of StartClientTanstack to the entry-client in the tanstack example?

brenelz: having some trouble inlining this

birkskyum (approved):

PR #1867, opened 2025-03-21
feat: allow action function to disable singleFlight mode

mtt-artis:

in action, i do not need to revalidate all preload function if I return an Error.

## What is the new behavior?

disable singleFlight mode by removing the header 'X-Single-Flight' in the action function

```ts
const update = action(async (formData: FormData) => {
  "use server";
  try {
    
  } catch (error) {
    getRequestEvent()?.request.headers.delete("X-Single-Flight");
    return new Error("oups");
  }
})
```

### DIFF
--- a/packages/start/src/runtime/server-handler.ts
+++ b/packages/start/src/runtime/server-handler.ts
@@ -84,7 +84,6 @@ async function handleServerFunction(h3Event: HTTPEvent) {
 
   const serverReference = request.headers.get("X-Server-Id");
   const instance = request.headers.get("X-Server-Instance");
-  const singleFlight = request.headers.has("X-Single-Flight");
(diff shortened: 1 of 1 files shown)


brenelz:
You can use something like `return redirect('/', {revaidate: ['nothing']})` instead

mtt-artis:
Hi

> You can use something like `return redirect('/', {revaidate: ['nothing']})` instead
I still want to return an error message 
I found returning a Error in action quite pleasant for type and easy discrimination between results and errors on the front side

mtt-artis:
My bad, this remove the singleflight but retriger flights to revelidate cause the response miss a 'X-Revalidate' header
https://github.com/solidjs/solid-router/blob/main/src/data/action.ts#L208

PR #1872, opened 2025-03-27
feat:encode sfn > functionId & extractedFilename

### LINKED ISSUE #1859: [Bug?]: `'use server'` will cause an error when there is Chinese in the file path
### Duplicates

### Latest version

### Current behavior 😯

```
TypeError: Failed to execute 'fetch' on 'Window': Failed to read the 'headers' property from 'RequestInit': String contains non ISO-8859-1 code point.
    at createRequest (server-runtime.js:94:12)
    at fetchServerFunction (server-runtime.js:119:11)
    at fn (server-runtime.js:175:32)
```

### Expected behavior 🤔

no error

### Steps to reproduce 🕹

Steps:

1. create file `src/routes/中文.tsx`
```ts
import { Title } from '@solidjs/meta';
import { createAsync, query } from '@solidjs/router';
import { createSignal, Suspense } from 'solid-js';

const getInfo = query(async (id: number) => {
  'use server';

  return { id };
}, 'getInfo');

export default function zh() {
  const [uid, setUid] = createSignal(0);
  const info = createAsync(() => getInfo(uid()));

  return (
    <main>
      <Title>中文</Title>
      <h1>id = {uid()}</h1>
      <div class="flex gap-lg ">
        <button onClick={() => setUid((n) => n + 1)}>id+1</button>
        <button onClick={() => setUid((n) => n - 1)}>id-1</button>
      </div>
      <Suspense fallback="loading...">
        <pre>data = {JSON.stringify(info(), null, 1)}</pre>
      </Suspense>
    </main>
  );
}
```

2.  Visit the corresponding page
3. click  btn `id+1`

### Context 🔦

Because `headers` does not allow special characters such as Chinese

But in `x-server-id` there is a full path, Chinese characters will appear

### Your environment 🌎

```shell

```

thinke5:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1859

image

## What is the new behavior?

image

## Other information

The encoded characters are guaranteed to be ASCII characters, and the path before `src` is removed to improve security and reduce string size

Currently encoded as base64, the size increases by 25%, but I don't know if there is a better way

### DIFF
--- a/packages/start/config/index.js
+++ b/packages/start/config/index.js
@@ -37,6 +37,23 @@ function solidStartServerFsRouter(config) {
     );
 }
 
+/**
+ * @param {string} filepath
+ * @returns {string}
+ */
+function convertToRelativePath(filepath) {
+  const arr = filepath.split("/src/");
+  return arr[1] ? `~/${arr[1]}` : filepath;
+}
+
+/** encode string, here to base64
+ * @param {string} str
+ * @returns {string}
+ */
+function encodeString(str) {
+  return btoa(encodeURIComponent(str));
+}
+
 const SolidStartServerFnsPlugin = createTanStackServerFnPlugin({
   // This is the ID that will be available to look up and import
   // our server function manifest and resolve its module
@@ -47,7 +64,7 @@ const SolidStartServerFnsPlugin = createTanStackServerFnPlugin({
         fileURLToPath(new URL("../dist/runtime/server-runtime.js", import.meta.url))
       )}"`,
     replacer: opts =>
-      `createServerReference(${() => {}}, '${opts.functionId}', '${opts.extractedFilename}')`
+      `createServerReference(${() => {}}, '${encodeString(opts.functionId)}', '${encodeString(convertToRelativePath(opts.extractedFilename))}')`
   },
   ssr: {
     getRuntimeCode: () =>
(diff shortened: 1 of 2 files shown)


lxsmnsyc:
Could probably use a simple hash such as a xxhash32(filename)

thinke5:
> Could probably use a simple hash such as a xxhash32(filename), but otherwise this is fine.

Hash functions are usually irreversible. 

If use encryption algorithms (such as AES) or compression algorithms (such as lz-string), need to decode it every time call `handleServerFunction`, which may cause high CPU consumption and network latency. 

So I choose to use base64 for now.

thinke5:
I have come up with a solution that can use hash algorithm，` HandleServerFunction does not require decoding. I need some time to verify

thinke5:
image

Now:
Hash `functionId` with sha256.
Encode `extractedFilename` with `encodeURIComponent`.

Process `serverFnManifest` in `server-handler` file and copy a hashed `key`. No need to decode it during request processing.

In addition, I found that `extractedFilename` is not used in this process, 🤔 maybe it can be removed?

brenelz:
@lxsmnsyc any other comments on this pr?

lxsmnsyc:
@brenelz this PR is fine, but we'll need a solution that outputs a shorter output like xxhash32

brenelz:
@thinke5 did you want to look at this following Alexis' comment?

brenelz:
@katywings can we close this pr now? Looks like you referenced some of this stuff in the new version?

Issue #1886, opened 2025-05-02
[Bug?]: "use server" in async generator is broken


theturtle32:
### Duplicates

- [x] I have searched the existing issues

### Latest version

- [x] I have tested the latest version

### Current behavior 😯

Adding the `"use server"` directive to an async generator fails with the error `"use server" in generator function not supported`.

The error is raised from the plugin `tanstack-start-directive-vite-plugin`

Error location is @tanstack/directive-functions-plugin/dist/esm/compilers.js:327:10
Version of `@tanstack/directive-functions-plugin` is `1.115.0`

### Expected behavior 🤔

`"use server"` in async generator functions should work. And it did work before adopting `@tanstack/server-functions-plugin` in #1715 

And indeed, if you hoist "use server" to the module level, the functionality DOES work even today, but then you get warnings about module-level "use server" being unsupported when you run `pnpm run build`, which I think is an otherwise unrelated issue (#1799).

### Steps to reproduce 🕹

Steps:

1. `pnpm create solid`
2. Create a SolidStart project using the `basic` template.
3. Add the following files:

```typescript
/* src/server-functions.tsx */

// Say hello in a different language every second, forever.
export async function* sayHello() {
  "use server";
  const greetings = [
    "Hello, World!",      // English
    "¡Hola, Mundo!",      // Spanish
    "Bonjour le monde !", // French
    "Hallo Welt!",        // German
    "Ciao mondo!",        // Italian
    "Olá Mundo!",         // Portuguese
    "こんにちは世界",        // Japanese
    "你好，世界"            // Chinese (Mandarin)
  ];

  for (const greeting of greetings) {
    yield greeting;
    await new Promise(resolve => setTimeout(resolve, 1000)); // Wait for 1 second
  }
}
```

```typescript
/* src/routes/generator-server-fn.tsx */

import { Title } from "@solidjs/meta";
import { For, onMount } from "solid-js";
import { createStore } from "solid-js/store";
import { sayHello } from '~/server-functions';

export default function GeneratorServerFn() {
  const [store, setStore] = createStore<Array<string>>([]);

  onMount(async () => {
    const greetings = await sayHello();
    for await (const greeting of greetings) {
      setStore(store.length, greeting);
    }
  });

  return (
    <main>
      <Title>Generator Server Fn</Title>
      <h1>Generator Server Fn</h1>
      <For each={store}>
        {(message) => <div class="message">{message}</div>}
      </For>
    </main>
  );
}
```

[...]

PR #1887, opened 2025-05-02
Adding failing test for issue #1886

theturtle32:
## PR Checklist

This PR is not intended to be merged, just to add a failing e2e test that illustrates the bug which should stop failing once the bug is fixed.

See issue #1886 

## What is the new behavior?

Added failing tests that should stop failing when the underlying bug is fixed.


brenelz:
Thank you for providing this failing test. I incorporated it in my PR that should be going out in the next release

Issue #1898, opened 2025-06-01
[Bug?]: Infinite loop when a server function returns unexpected XML


yinonburgansky:
### Duplicates

- [x] I have searched the existing issues

### Latest version

- [x] I have tested the latest version

### Current behavior 😯

When running AWS serverless, sometimes we get unexpected responses from server functions like:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessDenied</Code><Message>Access Denied</Message></Error>
```
This response causes an infinite loop which freezes the browser and causes a memory leak.

The response is handled in this function:
https://github.com/solidjs/solid-start/blob/f4052b2532b54ba09b74f566f6502e0ae522a255/packages/start/src/runtime/server-runtime.ts#L143
https://github.com/solidjs/solid-start/blob/f4052b2532b54ba09b74f566f6502e0ae522a255/packages/start/src/runtime/server-runtime.ts#L187

The infinite loop in:
https://github.com/solidjs/solid-start/blob/f4052b2532b54ba09b74f566f6502e0ae522a255/packages/start/src/runtime/server-runtime.ts#L83

The error:
https://github.com/solidjs/solid-start/blob/f4052b2532b54ba09b74f566f6502e0ae522a255/packages/start/src/runtime/server-runtime.ts#L60
head = "?xml versi"
bytes = NaN

I can add a check for NaN but I think it might be worth to also handle unexpected responses higher up in the `fetchServerFunction`

### Steps to reproduce 🕹

Steps:

1.
2.
3.
4.

### Your environment 🌎

```shell

```

PR #1899, opened 2025-06-01
fix: infinite loop on unexpected response from server function (#1898)

yinonburgansky:
This fixes an infinite loop when a server function receives unexpected response like 403
```xml
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessDenied</Code><Message>Access Denied</Message></Error>
```
see issue #1898

There aren't any unit tests setup I could find to add a test for this.

This just fixes the infinite loop, but probably better error handling for response codes like 403 and response content types like should be made, it would be beneficial to get the response code and body to be able to gracefully handle errors on server functions - this requires some API design decisions?

### DIFF
--- a/.changeset/tasty-hounds-poke.md
+++ b/.changeset/tasty-hounds-poke.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/start": patch
+---
+
+fix infinite loop on unexpected response from server function (issue #1898)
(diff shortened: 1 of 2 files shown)


lxsmnsyc (approved):

lxsmnsyc:
Thanks

Issue #1869, opened 2025-03-23
[Bug?]: server assets in .output/public


huseeiin:
### Duplicates

- [x] I have searched the existing issues

### Latest version

- [x] I have tested the latest version

### Current behavior 😯

![Image](https://github.com/user-attachments/assets/7729d596-223c-414d-99f4-c716ed5d18b8)

### Steps to reproduce 🕹

Steps:

1.
2.
3.
4.

### Context 🔦

it happens with every preset it seems

### Your environment 🌎

```shell

```

arnfaldur:
You need to fill in all the entries to describe your issue. What is your "Expected behavior" for example?

It's disrespectful to ask for help from volunteers while not doing the bare minimum requested by them.

huseeiin:
i think this is fixed now. the folder exists but its empty

![Image](https://github.com/user-attachments/assets/f4d508fe-ed91-41c3-814d-b465a146f7e3)

PR #1901, opened 2025-06-05
fix leaking server functions to client

madaxen86:
tested and bumped version

fixes leaking server code to client bundle because of missing matches for compresed js assets in [vinxi](https://github.com/nksaraf/vinxi/blob/6d31572a0516d2298e51c7e3db7846efe3e5441f/packages/vinxi/lib/build.js#L334)

this is a workaround until this is fixed in vinxi. left a comment in the code.

related issues:
[1869](https://github.com/solidjs/solid-start/issues/1869 ) from 2025-03-23

### DIFF
--- a/examples/bare/package.json
+++ b/examples/bare/package.json
@@ -8,7 +8,7 @@
   },
   "dependencies": {
     "@solidjs/start": "^1.1.0",
-    "solid-js": "^1.9.5",
+    "solid-js": "^1.9.7",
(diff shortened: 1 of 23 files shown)


madaxen86:
Vinxi merged the changes and released a new version "0.5.7" which fixed the issue. Will change this PR to update the package.json to use the latest vinxi version.

madaxen86:
will file separate PR with new vinxi version

PR #1903, opened 2025-06-08
Add strict CSP example

amirhhashemi:
This PR adds an example demonstrating how to implement a strict CSP with a nonce. I created the project with the Solid CLI using the bare template and added CSP.

### DIFF
--- a/examples/with-strict-csp/src/middleware.ts
+++ b/examples/with-strict-csp/src/middleware.ts
@@ -7,19 +7,28 @@ export default createMiddleware({
   onRequest: event => {
     const nonce = randomBytes(16).toString("base64");
 
-    event.locals.nonce = nonce;
+    if (isProd) {
+      event.locals.nonce = nonce;
+    }
 
     // Notes:
-    // 1. SolidStart uses `eval` for data serialization, which may require you to add 'unsafe-eval' to your CSP.
+    // 1. SolidStart uses `eval` for data serialization, which may require you to include the 'unsafe-eval' directive in your CSP.
     //    For more information, see: https://github.com/solidjs/solid-start/issues/1825
-    // 2. In development, Vite inlines small CSS files to enhance performance, so you will need to include 'unsafe-inline' in development mode.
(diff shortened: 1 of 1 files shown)


examples/with-strict-csp/src/middleware.ts
```diff
@@ -0,0 +1,31 @@
...
+    const nonce = randomBytes(16).toString("base64");
+
+    event.locals.nonce = nonce;
+
+    // Notes:
+    // 1. SolidStart uses `eval` for data serialization, which may require you to add 'unsafe-eval' to your CSP.
+    //    For more information, see: https://github.com/solidjs/solid-start/issues/1825
+    // 2. In development, Vite inlines small CSS files to enhance performance, so you will need to include 'unsafe-inline' in development mode.
+    // 3. During the build process, Vite inlines small assets as data URLs. Therefore, it's necessary to add data: to the relevant directives (e.g., img-src, font-src, etc.)
+    //    For more details, see: https://vite.dev/config/build-options.html#build-assetsinlinelimit
```
amirhhashemi, line 17: I added this comment to clarify some of the confusing directives I used. I'm specifically unsure about point two. I couldn't find the documentation regarding Vite inlining small CSS files, but apparently, it happens during development. I encountered some errors related to this while testing.

examples/with-strict-csp/src/middleware.ts
```diff
@@ -0,0 +1,31 @@
...
+    const csp = `
+      default-src 'self';
+      script-src 'nonce-${nonce}' 'strict-dynamic' 'unsafe-eval';
+      style-src 'self' ${isProd ? "" : "'unsafe-inline'"};
```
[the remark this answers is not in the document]
amirhhashemi: Done

amirhhashemi:
I tried to address the mentioned concerns, but also made some small changes. So please review carefully.

examples/with-strict-csp/src/middleware.ts
atilafassina: yes, it does. I have found that to be the required as well.
I think it's the critical CSS path 👍

atilafassina (approved):
🏆 thank you for the work and the patience!

PR #1910, opened 2025-06-26
Update Deno quick start command

amirhhashemi:
This PR updates the "create" command for Deno. [`init --npm`](https://docs.deno.com/runtime/reference/cli/init/) is the official command for this purpose. The documentation site was also recently updated to use `init --npm`.

### DIFF
--- a/packages/landing-page/src/components/code-snippet.tsx
+++ b/packages/landing-page/src/components/code-snippet.tsx
@@ -52,7 +52,7 @@ export function CodeSnippet() {
         <TabContent manager="bun" command="create solid" />
         <TabContent manager="npm" command="create solid" />
         <TabContent manager="yarn" command="create solid" />
-        <TabContent manager="deno" command="run -A npm:create-solid" />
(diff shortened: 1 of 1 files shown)


packages/landing-page/src/components/code-snippet.tsx
```diff
@@ -52,7 +52,7 @@ export function CodeSnippet() {
         <TabContent manager="bun" command="create solid" />
         <TabContent manager="npm" command="create solid" />
         <TabContent manager="yarn" command="create solid" />
-        <TabContent manager="deno" command="run -A npm:create-solid" />
+        <TabContent manager="deno" command="init --npm solid" />
```
brenelz, line 55: are you sure this is the right command? I think we want to use the create-solid cli?

amirhhashemi: With the `--npm` flag, Deno automatically adds a `create-` to the command. I'm not a deno user, but I assume that's how it works based on their docs: https://docs.deno.com/runtime/reference/cli/init/#options-npm

PR #1919, opened 2025-07-10
fix: `fetchEvent` -> `HeaderProxy` checks for `null` instead of `undefined`

huseeiin:
## PR Checklist

`event.response.headers.has("does-not-exist")` returns `true`

## What is the new behavior?
`event.response.headers.has("does-not-exist")` returns `false`

### DIFF
--- a/packages/start/src/server/fetchEvent.ts
+++ b/packages/start/src/server/fetchEvent.ts
@@ -53,7 +53,7 @@ class HeaderProxy {
 		return Array.isArray(h) ? h.join(", ") : (h as string) || null;
 	}
 	has(key: string) {
-		return this.get(key) !== undefined;
+		return this.get(key) !== null;
 	}
 	set(key: string, value: string) {
 		return setResponseHeader(this.event, key, value);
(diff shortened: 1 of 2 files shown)


packages/start/src/server/fetchEvent.ts
```diff
@@ -53,7 +53,7 @@ class HeaderProxy {
 		return Array.isArray(h) ? h.join(", ") : (h as string) || null;
 	}
 	has(key: string) {
-		return this.get(key) !== undefined;
+		return this.get(key) !== null;
```
brenelz, line 56: is there any way we can add a test or something for this?

huseeiin: > is there any way we can add a test or something for this?

should i add a e2e test?

huseeiin: > is there any way we can add a test or something for this?

ignore this pr, i have a much better pr now: https://github.com/solidjs/solid-start/pull/1937

atilafassina (approved):

PR #1937, opened 2025-08-08
fix: overhaul headers proxy

huseeiin:
this fixes many issues with the original headers proxy, for example, now:
`getRequestEvent().response.headers.toString()` returns "[object Headers]" instead of "[object Object]"
`getRequestEvent().response.headers instanceof Headers` will return `true`
and so on, because we're now using a native `Headers` object


huseeiin:
don't merge this, its stupid. after h3 v2 we will get a proper response Headers

PR #1950, opened 2025-08-23
Autoload .env

brenelz:


brenelz:
@Brendonovich is this still a post devinxi thing?

PR #1953, opened 2025-08-24
fix css handling during ssr

brenelz:
Before this change the initial page styles wouldn't load with ssr in the prod build for the basic example. 

You can see this with the count button on initial load.

Before:
image

After
image


Brendonovich:
In the process of redoing some manifest handling I actually ended up implementing this myself 🙈

PR #1956, opened 2025-08-28
add server base url

brenelz:
Adds back an env variable thats required and was set in the vinxi version. Also now its able to be configured.

Also fixes the issue with Solid Notes


birkskyum:
> Also fixes the issue with Solid Notes

Does that issue have a ticket? Is it the dev overlay env var?

brenelz:
It's just an error in devinxi branch. Is because of the SERVER_BASE_URL

birkskyum (approved):

PR #1957, opened 2025-08-28
Multibranch release

atilafassina:
This PR prepares the monorepo for multibranch releases.
Our plan is to continue supporting start `1.x` while `2.x` isn't stable.

- `main` branch: version `2.0.0` from DeVinxi work.
- `v1.x` will still shit the Vinxi architecture and receive bugfixes or general maintenance ( timeline for EOL is not defined yet).

Changesets does not support configurable `baseBranch` so we need dual-config for that at the moment. Going forward we can keep specific configurations per branch if we prefer.


atilafassina:
Doing https://github.com/solidjs/solid-start/pull/1977 first and then we'll split the branches

PR #1958, opened 2025-08-28
cover `fetchEvent` with a few unit tests

atilafassina:
adds Vitest and a few unit tests to existing code prevent regressions and more easily cover new fixes/features


brenelz (commented):
looks good to me. Some vinxi stuff in here we might have to rework in devinxi

PR #1965, opened 2025-09-05
WIP: Devinxi css rework

katywings:
## PR Checklist

## Design Goals

1. Consistency between dev/prod
2. Prevent foucs
3. Avoid unused route css
4. Avoid duplicate css

## Issues
- [x] `?url` imported css should not be server rendered
- [x] `<link>` should be mounted before suspension-end during spa navigation
- [x] Suspension from spa navigation should only end after `<link>` has been loaded
- [ ] Assets should be deduplicated on ssr
- [ ] Public manifest and asset mount API

## Known problems:
- [ ] app.css gets removed when navigating from /index to /about in css-garden
- [ ] `<link>` is mounted too early on spa navigation - it should wait until resources/createAsync is done


packages/start/src/shared/css.ts
```diff
@@ -0,0 +1,67 @@
...
+export const mountAssets = (assets: Asset[]) => {
+  if (!assets.length) return;
+
+  if (isServer) {
+    useAssets(() => assets.map((asset: Asset) => renderAsset(asset)));
+    const a: [] = (sharedConfig.context as any).assets;
+    const index = a.length - 1;
+    onCleanup(() => {
+      // TODO: index is not properly tracked!
+      a.splice(index, assets.length);
```
brenelz, line 15: what does the onCleanup do? Removes an index from sharedConfig?

brenelz:
This is still going to be post-devinxi into main?

brenelz:
bunch of conflicts now too

katywings:
> This is still going to be post-devinxi into main?

@brenelz It's still up for discussion of this is going straight into devinxi or if it is going into main. Why 😅?

katywings:
> bunch of conflicts now too

@brenelz Just rebased it, conflicts are gone 😃. Next on my TODO list: rework the DEV part (try to render `<link>` in dev, instead of `<style>`).

apps/fixtures/css/src/entry-server.tsx
```diff
@@ -0,0 +1,24 @@
+// @refresh reload
+import { createHandler, StartServer } from "@solidjs/start/server";
+import "./styles/entryServer.css";
+import url from "./styles/entryServerUrl.css?url";
```
brenelz: all this test are great

packages/start/src/server/spa/StartServer.tsx
```diff
@@ -18,12 +18,14 @@ export function StartServer(props: { document: Component<DocumentComponentProps>
...
   const nonce = context.nonce;
+  useAssets(context.assets, nonce);
+
   return (
     <NoHydration>
       {docType as unknown as any}
       <TopErrorBoundary>
         <props.document
-          assets={<>{context.assets.map((m: any) => renderAsset(m))}</>}
+          assets={<></>}
```
brenelz: is this supposed to empty? I guess the useAssets does it

apps/fixtures/css/src/components/layout.tsx
```diff
@@ -0,0 +1,56 @@
+import { createSignal, FlowProps, onMount } from "solid-js";
```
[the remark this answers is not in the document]
atk: FlowProps is the Props variant for a FlowComponent, i.e. one that supports function children, like e.g. For.

packages/start/src/server/spa/StartServer.tsx
katywings: Yuppa :), Solids/dom-expressions `useAssets` injects the assets directly in renderToString at the end of the head (<https://github.com/ryansolid/dom-expressions/blob/2a3de6623e10da44adceb3aaa718cf4325296576/packages/dom-expressions/src/server.js#L577C10-L577C22>).

I guess we could remove the "assets" prop here, but I am a bit hesitant if its 100% the right move 🤔

atilafassina (approved):

PR #1966, opened 2025-09-07
Import from TS in examples

Brendonovich:
This allows us to not have to build the Start package before testing it in examples.
It requires us to import `src/config` directly inside examples, as Vite doesn't transpile files from packages when importing them into the config, but I'd say this is worth it for the workflow improvement - Vite will even reload automatically when modifying the Start config implementation now.
It also requires making some modifications to our publishing process, since the default `exports` field now points to `.ts` files. `pnpm publish` will automatically use the `.js` exports defined in `publishConfig`, but changesets and pkg.pr.new require a small `jq` command to swap out the `exports` field before publishing.


birkskyum:
I can see the significant convenience of this for local development, but this change won't work on the main branch for a number of reasons, so it has to be reverted before merging into that. Alternatively, I've previously use a small script that converted the import paths to be relative like here in few ms, and that was an okay development experience as well.

The main issues that makes this incompat with main are:
- Deps tied to certain start/vinxi versions like mediakit/auth, they'll cause version conflicts in the lockfile that cause various symptoms and let to the split lockfile in the first place.
- The examples are cloned directly from the solid-cli, so they can't have relative imports.

A middleground that resolves the latter could be using overrides like tanstack/router, where in local development, the examples will resolve ot local packages, but if cloned, they'll resolve to the npm packages. The package overrides doesn't resolve the version conflict of i.e. mediakit/auth though:
- https://github.com/TanStack/router/blob/main/package.json#L72-L119

birkskyum:
solid-cli has been updated to use solid-start examples from the solidjs/templates repo, so this repo can move forward with just having some fixtures that are imported from TS

atilafassina:
let's try this one once we merge https://github.com/solidjs/solid-start/pull/1977

birkskyum:
Probably easier to remake than rebase at this point.

Brendonovich:
Agreed, will close and recreate

PR #1968, opened 2025-09-07
fix: don't send server stack trace to client

### LINKED ISSUE #1967: [Bug?]: [security] Server stack trace sent to client on error in server action
### Duplicates

### Latest version

### Current behavior 😯

When throwing an error in a server action the Error is being serialized and sent to the client with the stack trace which poses a security issue leaking production code file locations and function names.

### Expected behavior 🤔

At least on Production don't send server's stack traces to the client.

### Steps to reproduce 🕹

Any error thrown in server action:
```ts
const myAction = action(async () => {
  'use server';
  throw new Error('my server error');
});
```
[See stackblitz full example](https://stackblitz.com/edit/github-sz5m8erj-uacuifk8?file=src%2Froutes%2Findex.tsx)

### Context 🔦

_No response_

### Your environment 🌎

```shell

```

yinonburgansky:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1967
- [ ] Tests for the changes have been added (for bug fixes / features) - missing infrastructure to unit test server functions.

Avoid sending Error.stack to the client. Stack traces can leak production file paths, internal function names, and other sensitive info, increasing attack surface.

### DIFF
--- a/packages/start/src/runtime/server-handler.ts
+++ b/packages/start/src/runtime/server-handler.ts
@@ -1,5 +1,5 @@
 /// <reference types="vinxi/types/server" />
-import { crossSerializeStream, fromJSON, getCrossReferenceHeader } from "seroval";
+import { crossSerializeStream, Feature, fromJSON, getCrossReferenceHeader } from "seroval";
 // @ts-ignore
 import {
   CustomEventPlugin,
@@ -62,6 +62,7 @@ function serializeToStream(id: string, value: any) {
           URLSearchParamsPlugin,
           URLPlugin
         ],
+        disabledFeatures: import.meta.env.PROD ? Feature.ErrorPrototypeStack : undefined,
         onSerialize(data, initial) {
           controller.enqueue(
             createChunk(initial ? `(${getCrossReferenceHeader(id)},${data})` : data)


yinonburgansky:
This doesn't compile currently due to `Cannot access ambient const enums when 'isolatedModules' is enabled.ts(2748)` 
see https://ncjamieson.com/dont-export-const-enums/
[Seroval export Future as a const enum opened issue](https://github.com/lxsmnsyc/seroval/issues/63)

lxsmnsyc:
I think this has been discussed multiple times before, the reason we don't actually do this is:
- A: Bundled files wouldn't actually be helpful in stack trace, as the trace itself is mangled.
- B: Stack trace is helpful in development, so if we don't consider the reason A, we should still consider reason B.

yinonburgansky:
>     * A: Bundled files wouldn't actually be helpful in stack trace, as the trace itself is mangled.

- It is exposing the full path of the source files in the production server which can be used by an attacker in another stage to get the entire production code.
- Security through obscurity shouldn't be relied upon. Mangling just makes it more difficult to understand the code, compiling to binary also makes it difficult to reverse engineer the code, but it is 10X easier than not having the code (black box).

This is considered a well known security risk:
[CWE-209: Generation of Error Message Containing Sensitive Information](https://cwe.mitre.org/data/definitions/209.html):
> An attacker may use the contents of error messages to help launch another, more focused attack. For example, an attempt to exploit a path traversal weakness ([CWE-22](https://cwe.mitre.org/data/definitions/22.html)) might yield the full pathname of the installed application. In turn, this could be used to select the proper number of ".." sequences to navigate to the targeted file. An attack using SQL injection ([CWE-89](https://cwe.mitre.org/data/definitions/89.html)) might not initially succeed, but an error message could reveal the malformed query, which would expose query logic and possibly even passwords or other sensitive information used within the query.

see also:

- https://security.stackexchange.com/questions/19130/is-a-stack-trace-of-a-server-application-a-vulnerability
- https://www.quora.com/Is-it-acceptable-to-log-server-errors-to-a-JavaScript-fronted

>     * B: Stack trace is helpful in development, so if we don't consider the reason A, we should still consider reason B.

I changed it to remove it only on production.
If some developers want server stack traces on production - it should be opt-in since it is considered an information leak, adding an API to enable it can be done separately if really needed - it shouldn't block a security fix. 

BTW I think solid-start should provide a mechanism to decide which errors should be sent to the client, e.g. sending DB errors to the client by default is a very bad idea, potentially exposing schema info, credentials location (e.g. env.DB_URL) and other sensitive information.  
Wrapping every server action with a try catch is very verbose [see my attempt](https://stackblitz.com/edit/github-keyru1dl?file=src%2Froutes%2Findex.tsx).

lxsmnsyc:
There's no way to filter out errors atm since seroval handles the entire thing. Will be discussing the rest of the details with the core team

brenelz:
Going to close this for now as won't do

PR #1977, opened 2025-09-20
Rearrange architecture according to RFC

atilafassina:
Part of the work towards the [Fullstack Primitives](https://github.com/solidjs/solid-start/discussions/1984) initiative.

> [!CAUTION]
> landing page is failing because I need to fix the routing in Netlify before merging.


birkskyum:
it looks like /tests is here nested inside of /apps/tests. In the  [design](https://github.com/solidjs/solid-start/discussions/1984) suggested it looks like it should be at /tests at root level

katywings:
One of the past inconsistencies was with code formatting. We do have a prettier config, but I think it isn't very enforced, neither README.md, nor CONTRIBUTING.md talks about it, and there is no formatting/linting script.

Normally I would have recommended to add it here, but since there are lots of files with "wrong" formatting, I suggest to start enforcing this in either devinxi or post-devinxi. If we do it here, I fear that the merge conflicts would be a huge waste of time 😅.

atilafassina:
> it looks like /tests is here nested inside of /apps/tests. In the [design](https://github.com/solidjs/solid-start/discussions/1984) suggested it looks like it should be at /tests at root level

that was a typo in the docs. Intention was to have them in `/apps` because those are indeed SolidStart apps

PR #1986, opened 2025-10-04
[Devinxi] Separate Nitro integration into a separate plugin

Brendonovich:
This PR generalises Start's build process to be compatible with more than just Nitro, opening up support for [vite-plugin-cloudflare](https://github.com/cloudflare/workers-sdk/tree/main/packages/vite-plugin-cloudflare) in future.
Nitro v2 can be added to a project using `nitroV2Plugin` from `@solidjs/start/nitro-v2-plugin`, though we may want to separate into `@solidjs/nitro-v2-vite-plugin` to avoid having Start be dependent on Nitro v2.


Brendonovich:
already merged this into devinxi

PR #1994, opened 2025-10-27
Use <link> instead of <style> in dev

Brendonovich:
This PR is mostly just for access to preview packages


Brendonovich:
CSS module HMR seems like a dead end with `<link>`

PR #1995, opened 2025-10-29
[v1] bump solid core and vite-plugin

atilafassina:


brenelz:
Some conflicts here now

PR #2004, opened 2025-11-18
Publish `@solidjs/vite-plugin-nitro-2`

atilafassina:
This PR creates a changeset for publishing the `@solidjs/vite-plugin-nitro-2` to NPM.

a deploy release can be tested using the [instructions below](https://github.com/solidjs/solid-start/pull/2004#issuecomment-3547448805)

### DIFF
--- a/.github/workflows/cr.yml
+++ b/.github/workflows/cr.yml
@@ -35,8 +35,7 @@ jobs:
       - name: Build Packages
         run: |
           pnpm run build
-          # rewrite .ts exports to .js
-          pnpm --filter='./packages/*' -c exec "echo \$(cat package.json | jq '.exports = .publishConfig.exports') > package.json"
+          pnpm rewrite-exports
 
       - name: Release Packages
         # remove the compat flag until all packages are available on npm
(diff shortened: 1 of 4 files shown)


brenelz:
I know the Nitro team pushing people pretty hard to use v3 but this probably works in the meantime

atilafassina:
> I know the Nitro team pushing people pretty hard to use v3 but this probably works in the meantime

We discussed this the other day and I already talked to the Nitro team about it. We're support Nitro 2 and 3 during alpha and only deprecate Nitro 2 support in the future.

huseeiin:
why does this exist since nitro v3 already works fine with solidstart?

brenelz:
> why does this exist since nitro v3 already works fine with solidstart?

There is still some issues with v3 that are solved using the v2 plugin. So its a temporary thing while the kinks get ironed out

atilafassina:
@huseeiin 

> why does this exist since nitro v3 already works fine with solidstart?

1. to facilitate migration for existing apps that are on Nitro 2
2. Nitro/Vite doesn't have prerender yet

PR #2001, opened 2025-11-14
fix: explicit type imports

brenelz:
## PR Checklist

Copied the bare example outside the monorepo and was having some type issues that explicty `import type` fixed

PR #2007, opened 2025-11-18
VItest + Playwright migration

DBozhinovski:
- Migrates to Vitest + Playwright from Cypress
- Updates tests.yml for GH actions
- Updated README and CONTRIBUTING to reflect changes to tests

### DIFF
--- a/packages/start-nitro-v2-vite-plugin/src/index.ts
+++ b/packages/start-nitro-v2-vite-plugin/src/index.ts
@@ -20,106 +20,120 @@ export type UserNitroConfig = Omit<
 >;
 
 export function nitroV2Plugin(nitroConfig?: UserNitroConfig): PluginOption {
-	return {
-		name: "solid-start-vite-plugin-nitro",
-		generateBundle: {
-			handler(_options, bundle) {
-				if (this.environment.name !== "ssr") {
-					return;
-				}
+	return [
+  	{
+  		name: "solid-start-vite-plugin-nitro",
+  		generateBundle: {
+  			handler(_options, bundle) {
+  				if (this.environment.name !== "ssr") {
+  					return;
+  				}
 
-				// find entry point of ssr bundle
-				let entryFile: string | undefined;
-				for (const [_name, file] of Object.entries(bundle)) {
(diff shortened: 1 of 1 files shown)


brenelz:
Some build errors that look like this:

```
import { PluginOption, ViteDevServer } from "vite";
```

was why i created https://github.com/solidjs/solid-start/pull/2001

brenelz:
Also complaining about 
>  ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with <ROOT>/apps/tests/package.json

apps/tests/src/e2e/api-call.test.ts
```diff
@@ -0,0 +1,8 @@
+import { expect, test } from "@playwright/test";
+
+test.describe("api calls", () => {
+  test("should return text/plain", async () => {
+    const response = await fetch("http://localhost:3000/api/text-plain");
+    expect(await response.text()).toBe("test");
```
brenelz: This test is failing for me. I think we need to change this file `apps/tests/src/routes/api/text-plain.tsx` to use the following instead of `e.nativeEvent.respondWith`

```
export function GET() {
  return new Response("test");
}
```

packages/start/src/config/manifest.ts
```diff
@@ -5,87 +5,80 @@ import { VIRTUAL_MODULES } from "./constants.ts";
...
-						(globalThis.START_CLIENT_BUNDLE[".vite/manifest.json"] as any)
-							.source,
-					);
-				}
-				return `export const clientViteManifest = ${JSON.stringify(clientViteManifest)};`;
-			} else if (id === `\0${VIRTUAL_MODULES.middleware}`)
-				return "export default {};";
-			else if (id.startsWith("/@manifest")) {
-				if (this.environment.mode !== "dev") throw new Error("@manifest queries are only allowed in dev");
+  let devServer: ViteDevServer = undefined!;
```
atilafassina: are changes in this file only in formatting?

PR #2011, opened 2025-11-23
Write event.response.headers to returned Responses

Brendonovich:
If a Response is returned from an API route or server function, any headers written via `@solidjs/start/http` including `useSession` are not passed on. This writes headers from `getRequestEvent().response` to the returned response, if the header doesn't already exist on the returned response.
In the case of redirects being returned, we clone the response in case it is immutable (such as those returned from `Response.redirect`)


huseeiin:
doesn't h3 v2 already do this?

Brendonovich:
@huseeiin Not in 100% of cases

PR #2013, opened 2025-11-23
make vite a peer dependency of start

Brendonovich:


packages/start/package.json
```diff
@@ -61,14 +61,17 @@
     "source-map-js": "^1.2.1",
     "srvx": "^0.9.1",
     "terracotta": "^1.0.6",
-    "vite": "7.1.10",
     "vite-plugin-solid": "^2.11.9",
     "vitest": "^4.0.10"
```
atk: vitest should be a dev dependency.

PR #2015, opened 2025-11-23
temporary clone all responses

thdxr:
ignore this PR i need to do this to test something with pkg pr


huseeiin:
if you're wondering how to use solid start v2 before release just `git clone https://github.com/solidjs/solid-start` and you'll find ready-to-use apps in `apps/fixtures` and then you can make all your changes

brenelz:
@thdxr can this pr be closed?

PR #2017, opened 2025-11-24
feat: add back HttpHeader

brenelz:
## PR Checklist

Based on a request we need to add back the HttpHeader component. Copied as is and exported it.


apps/tests/src/routes/http-header.tsx
```diff
@@ -0,0 +1,10 @@
+import { HttpHeader } from "../../../../packages/start/src/shared/HttpHeader";
```
brenelz: is this the right import?

Brendonovich: `@solidjs/start` is correct, have fixed it

PR #2020, opened 2025-11-24
Add oxfmt

DBozhinovski:
- Migrates all rules over to oxfmt config; 
- Updates CONTRIBUTING; 
- Formats full codebase; 
- Makes editorconfig as consistent as possible with oxfmt; 
- Removes prettierrc.


brenelz:
We should probably prioritize this. Otherwise we will keep getting conflicts

brenelz (approved):

PR #2022, opened 2025-11-24
fixtures should be `private: true` so they are not picked by changesets

atilafassina:


brenelz (approved):

huseeiin:
we should also change the tsconfig this way:
change `module` to `"esnext"` and `moduleResolution` to `"bundler"` (in `nitro-3` example) and delete `allowSyntheticDefaultImports` and `esModuleInterop` since they are not needed when `moduleResolution` is set to `"bundler"`

huseeiin:
also add `"types": ["vite/client"]` to tsconfig

PR #2042, opened 2026-01-04
feat(2.0): seroval json mode

lxsmnsyc:
Pretty straight forward. Adds an opt-in mode to use json mode (set by default) for seroval streaming, this is because the js stream is affected by CSP.

Also adds streaming support for client-to-server communication (except for url-encoded requests), assuming that the browser it comes from supports it.

### DIFF
--- a/packages/start/src/config/index.ts
+++ b/packages/start/src/config/index.ts
@@ -22,7 +22,14 @@ export interface SolidStartOptions {
   extensions?: string[];
   middleware?: string;
   serialization?: {
-    // This only matters for server function responses
+    /**
+     * The serialization mode to use for server functions/actions.
+     * The "js" mode uses a custom binary format that is more efficient than JSON, but requires a custom deserializer (with `eval()`) on the client.
+     * A strong CSP should block `eval()` executions, which would prevent the "js" mode from working.
+     * The "json" mode uses JSON for serialization, which is less efficient but can be deserialized with `JSON.parse` on the client.
+     *
+     * @default "json"
+     */
     mode?: "js" | "json";
   };
 }


packages/start/src/config/index.ts
```diff
@@ -21,6 +21,10 @@ export interface SolidStartOptions {
   routeDir?: string;
   extensions?: string[];
   middleware?: string;
+  serialization?: {
+    // This only matters for server function responses
+    mode?: 'js' | 'json';
```
lxsmnsyc: When possible, add a `plugins` option here that we can load into the serializer. The only problem is how do we bridge it from the config to the runtime (do we need a "runtime" config API?)

packages/start/src/server/serialization.ts
```diff
@@ -0,0 +1,253 @@
...
+  EventPlugin,
+  FormDataPlugin,
+  HeadersPlugin,
+  ReadableStreamPlugin,
+  RequestPlugin,
+  ResponsePlugin,
+  URLSearchParamsPlugin,
+  URLPlugin,
+];
+const MAX_SERIALIZATION_DEPTH_LIMIT = 64;
```
lxsmnsyc: Seroval has a default limit of 1000, but that's because of my poor judgment. Will probably port this default back.

packages/start/src/server/server-runtime.ts
```diff
@@ -154,15 +38,21 @@ async function fetchServerFunction(
...
         : createRequest(base, id, instance, {
-            ...options,
-            body: JSON.stringify(await Promise.resolve(toJSONAsync(args, { plugins }))),
-            headers: { ...options.headers, "Content-Type": "application/json" },
-          }));
+          ...options,
+          // TODO(Alexis): move to serializeToJSONStream
+          body: await serializeToJSONString(args),
+          // duplex: 'half',
+          // body: serializeToJSONStream(args),
```
lxsmnsyc: bleeding edge. TypeScript doesn't have the `duplex` property defined yet so this one is invalid for now. On top of that, this doesn't actually "stream" as in the server responds immediately on request. What the browser does currently is buffer the entire stream then sends in bulk (technically speaking, `duplex: half` describes this behavior, `duplex: full` describes the desired behavior).

Once all browsers are ready for this (and also TypeScript), we can swap to `duplex: full`

katywings:
Testing this branch on my project and uploading files through `formData` via https://github.com/hattipjs/hattip/tree/main/packages/base/multipart results in the following error 🙈:
`[unhandledRejection] Error: Invalid content type`

The error is throwed at this line: https://github.com/hattipjs/hattip/blob/15aa5ae4dec4478aa07785c85b2c85d4a753d0af/packages/base/multipart/src/form-data.ts#L79

It looks like with the changes in this branch, the upload doesn't set proper multipart boundaries anymore 🤔. The condition leading to the throw is `!match.directives.boundary` (https://github.com/hattipjs/hattip/blob/15aa5ae4dec4478aa07785c85b2c85d4a753d0af/packages/base/multipart/src/form-data.ts#L77C3-L77C29) - meaning: the `content-type` header misses the boundary flag 😅. E.g. the header should look as something like:

```
Content-Type: multipart/form-data; 
          boundary=gc0p4Jq0M2Yt08jU534c0p
```

(Reference: https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html)

lxsmnsyc:
Seemed like explicit definition of `multipart/form-data` overrides the derived content-type `fetch` does under the hood. Anyways fixed

brenelz:
Whats the latest on this pr?

lxsmnsyc:
@brenelz nothing much, but I haven't tested since the last commit lol

atilafassina (approved):

PR #2045, opened 2026-01-10
align dev port

huseeiin:
when you don't use nitro, as in:
```ts
import { solidStart } from "@solidjs/start/config";

export default defineConfig({
  plugins: [
    solidStart(),
  ],
});
```
instead of:
```ts
import { solidStart } from "@solidjs/start/config";
import { nitro } from "nitro/vite";

export default defineConfig({
  plugins: [
    solidStart(),
    nitro({
      preset: "node-server",
    }),
  ],
});
```
then solid start will use the 5173 port. now the port is aligned to be 3000 in both cases

### DIFF
--- a/.changeset/fast-hats-find.md
+++ b/.changeset/fast-hats-find.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/start": minor
+---
+
+make port 3000 all the time
--- a/packages/start/src/config/dev-server.ts
+++ b/packages/start/src/config/dev-server.ts
@@ -11,6 +11,9 @@ export function devServer(): Array<PluginOption> {
   return [
     {
       name: "solid-start-dev-server",
+      config() {
+        return { server: { port: 3000 } };
+      },
       configureServer(viteDevServer) {
         (globalThis as any).VITE_DEV_SERVER = viteDevServer;
         return async () => {


katywings (approved):

katywings:
@huseeiin Thank you 🥳

atilafassina (changes requested):
While I can understand wanting SolidStart to have a consistent port number by default, this is overriding user settings.

```ts
export defaultConfig({
   server: {
         port: 8888
    },
    plugins: [ solidStart() ]
})
```

will serve the app on `3000` instead of `8888`

huseeiin:
> While I can understand wanting SolidStart to have a consistent port number by default, this is overriding user settings.
> 
> ```ts
> export defaultConfig({
>    server: {
>          port: 8888
>     },
>     plugins: [ solidStart() ]
> })
> ```
> 
> will serve the app on `3000` instead of `8888`

amazing observation. how do we fix this?

katywings:
> > While I can understand wanting SolidStart to have a consistent port number by default, this is overriding user settings.
> > ```ts
> > export defaultConfig({
> >    server: {
> >          port: 8888
> >     },
> >     plugins: [ solidStart() ]
> > })
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > will serve the app on `3000` instead of `8888`
> 
> amazing observation. how do we fix this?

Maybe the solidStart plugin could add a different separate plugin (analog to our router, manifest, etc. plugins) for the port, that:
- uses `enforce: pre / post` (depending on your logic) 
- checks the `PORT` env (if needed)
- checks if the user has set a custom port via:

```tsx
// mutate the config directly (use only when merging doesn't work)
const mutateConfigPlugin = () => ({
  name: 'mutate-config',
  config(config) {
    if (!config.hasSomeValue) {
      config.hasSomeValue = "Hello World";
    }
  },
})

huseeiin:
since its not ready yet, will convert to draft, just in case someone merges is it on accident

atilafassina:
@katywings / @huseeiin 

the `config()` hook passes the configuration with only the user-set values. So if that `config` param above has the port defined, we don't touch it. If it doesn't,  we can set one. I haven't tried it, but I think that would do...

huseeiin:
> @katywings / @huseeiin
> 
> the `config()` hook passes the configuration with only the user-set values. So if that `config` param above has the port defined, we don't touch it. If it doesn't, we can set one. I haven't tried it, but I think that would do...

try it when you have time.

brenelz:
conflicts

brenelz:
Whats the latest here?

brenelz:
Closing as stale. We can revisit this in the future

PR #2047, opened 2026-01-12
refactor middleware

huseeiin:
i deleted some stuff like deprecation warnings please tell me if they should stay
currently in v2 when you do:
```ts
import { createMiddleware } from "@solidjs/start/middleware";

export default createMiddleware({
    onBeforeResponse: (e, response) => {
        console.log(response);
    },
});
```
`response.body` is undefined
that is fixed
also, there's a new breaking change (relative to start v1): response is now `Response` instead of `{ body: ReadableStream | string }` (`ReadableStream` if `mode` is `"stream"`, `string` otherwise)

### DIFF
--- a/.changeset/new-shirts-bake.md
+++ b/.changeset/new-shirts-bake.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/start": minor
+---
+
+refactor middleware. mainly fix onResponse
--- a/packages/start/src/middleware/index.ts
+++ b/packages/start/src/middleware/index.ts
@@ -14,11 +14,11 @@ export type ResponseMiddleware = (
   response: Response,
 ) => ReturnType<Parameters<typeof onResponse>[0]>;
 
-function wrapRequestMiddleware(onRequestFn: RequestMiddleware) {
+function wrapRequestMiddleware(onRequest: RequestMiddleware) {
   return async (h3Event: H3Event) => {
     // h3 onRequest doesn't allow returning a response, but we will for backwards compatibility with start v1
     const fetchEvent = getFetchEvent(h3Event);
-    const response = await onRequestFn(fetchEvent);
+    const response = await onRequest(fetchEvent);
     if (response) return response;
   };
 }
@@ -32,13 +32,14 @@ function wrapResponseMiddleware(onBeforeResponse: ResponseMiddleware): Middlewar
 }
 
 export function createMiddleware(
-  args:
+  args?:
     | {
         onRequest?: RequestMiddleware | RequestMiddleware[] | undefined;
         onBeforeResponse?: ResponseMiddleware | ResponseMiddleware[] | undefined;
       }
     | Middleware[],
 ): Middleware[] {
+  if (!args) return [];
   if (Array.isArray(args)) return args;
 
   const mw: Middleware[] = [];
--- a/packages/start/src/middleware/index.ts
+++ b/packages/start/src/middleware/index.ts
@@ -1,53 +1,40 @@
 // @refresh skip
 
-import type { H3Event, Middleware } from "h3";
+import { H3Event, type Middleware, onResponse } from "h3";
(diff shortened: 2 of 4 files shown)


katywings:
Imho the deprecation warnings should stay 🤔. Afaik the end goal is that users ideally will just use h3 middleware and we only keep `createMiddleware` for backwards compatibility.

huseeiin:
> Imho the deprecation warnings should stay 🤔. Afaik the end goal is that users ideally will just use h3 middleware and we only keep `createMiddleware` for backwards compatibility.

but how do they use h3 middleware?

huseeiin:
tip for users, when you do this:
```ts
export default createMiddleware({
    onBeforeResponse: async (e, r) => {
        console.log(await r.text());
    },
});
```
and you're streaming, your stream will be deferred.
to avoid that, you can use h3's `waitUntil` but by then you've left the request and can't respond with your modified response.
```ts
export default createMiddleware({
    onBeforeResponse: (e, r) => {
        e.nativeEvent.waitUntil((async () => {
            console.log(await r.text());
        })());
    },
});
```

huseeiin:
it seems this crashes with `ERROR Worker error: Invalid state: ReadableStream is locked` (only when Disable Cache is checked) when i refresh many times (will be fixed by https://github.com/h3js/srvx/pull/158). converting to draft

katywings:
> > Imho the deprecation warnings should stay 🤔. Afaik the end goal is that users ideally will just use h3 middleware and we only keep `createMiddleware` for backwards compatibility.
> 
> but how do they use h3 middleware?

To be honest, we still are debating it 😅. My plan is to write an RFC with a new API that will look somewhat like this:
https://github.com/katywings/solid-start/blob/727b61c0a0e13074dcf009298c08c2d391931f59/apps/fixtures/basic/src/entry-server.tsx#L20

In short: the user get's to create the `h3` instance themselves and can do with it whatever they want. I will go into more detail about backwards compatibility, composability and more in my RFC. That said, this new API is completely my own idea and I don't know how it will be perceived by the community and team. If you already have your own idea's how it should look, I'd love to hear em 🙂.

P.S. @ryansolid's long term goal/idea is, to basically make Start completely server-framework independent, so Start would just return a handler that can be integrated into any [WinterCG](https://wintertc.org/) compatible framework. We have to keep that in mind when thinking about middleware related API's.

huseeiin:
> > > Imho the deprecation warnings should stay 🤔. Afaik the end goal is that users ideally will just use h3 middleware and we only keep `createMiddleware` for backwards compatibility.
> > 
> > 
> > but how do they use h3 middleware?
> 
> To be honest, we still are debating it 😅. My plan is to write an RFC with a new API that will look somewhat like this: https://github.com/katywings/solid-start/blob/727b61c0a0e13074dcf009298c08c2d391931f59/apps/fixtures/basic/src/entry-server.tsx#L20
> 
> In short: the user get's to create the `h3` instance themselves and can do with it whatever they want. I will go into more detail about backwards compatibility, composability and more in my RFC. That said, this new API is completely my own idea and I don't know how it will be perceived by the community and team. If you already have your own idea's how it should look, I'd love to hear em 🙂.
> 
> P.S. @ryansolid's long term goal/idea is, to basically make Start completely server-framework independent, so Start would just return a handler that can be integrated into any [WinterCG](https://wintertc.org/) compatible framework. We have to keep that in mind when thinking about middleware related API's.

h3 v2 already exposes the app in `.nativeEvent.app` except it doesn't allow you to add routes (including middleware? i haven't checked) but i guess this is good in theory because more control is always better. but mixing config routes and file routes feels weird

huseeiin:
this is ready but don't merge before https://github.com/ryansolid/dom-expressions/pull/477

brenelz:
conflicts

brenelz:
@katywings is this still relevant to your middleware work or not really?

brenelz:
Closing as stale

PR #2052, opened 2026-01-18
feat: `dismantle`

lxsmnsyc:
This PR moves the current server function compiler from tanstack to dismantle.

> [!WARNING]
> Currently there's only one regression: since `dismantle` doesn't do an in-place replacement of the server function, it's not able to read local dependencies/side-effects except for module imports.
> Once `dismantle` is incorporated, we will require users to either move top-level variables/functions inside a server function or move those into separate module so `dismantle` can handle it properly.
>


lxsmnsyc:
This didn't align with the current intentions

PR #2053, opened 2026-01-19
fix onBeforeResponse

huseeiin:
this is like https://github.com/solidjs/solid-start/pull/2047 but only fixes the issue and doesn't introduce any new behavior. its your choice which to merge.


atilafassina (approved):
Thanks a lot!

PR #2060, opened 2026-01-29
feat: `vite preview`

huseeiin:
add `vite preview` support 🎉


brenelz:
Why are all our checks failing on this pr?

huseeiin:
> Why are all our checks failing on this pr?

because i did globalThis.CAN_SEND_FAST_NODE_STREAMS = true which is a type that doesn't exist, but i learned its not needed.

brenelz:
@lxsmnsyc Does this seem reasonable. Pretty small change overall

brenelz (approved):

PR #2061, opened 2026-02-03
fix: server id encoding

lxsmnsyc:
Current header value for `x-server-id` uses a `#` which is reserved. We have to use percent encoding.


lxsmnsyc:
> Seems reasonable to me. Do we need to do this in the new version as well?

no, 2.x uses a hash, so not much of an issue atm

atilafassina:
@copilot add a changest to this PR

PR #2065, opened 2026-02-14
Introduce `@solidjs/image`

lxsmnsyc:
This PR adds the `Image` component.

More details will be added soon.


birkskyum (approved):

.changeset/upset-vans-vanish.md
```diff
@@ -0,0 +1,5 @@
+---
```
birkskyum, line 1: No files in teh start package are modified, so start doesn't need a bump.

lxsmnsyc: changeset defaulted to semver choice iirc, thought that was weird for a sec

birkskyum: Are we sure the image component belong in the start repo? it doesn't appear to depend on the start workspace package.

lxsmnsyc: it's a co-feature. We just moved it out of the package so that anyone can still use it outside Start

PR #2073, opened 2026-02-21
Improve DX and test-cases on seroval mode

atilafassina:
- [x] Keep "js" as the default for V1
- [x] Add warnings and JSDoc for configuration options
- [x] Add a fixture to easily swap between modes and test them out


packages/start/config/index.d.ts
```diff
@@ -27,9 +27,16 @@ type SolidStartInlineConfig = {
     islands?: boolean;
   };
   serialization?: {
-    // This only matters for server function responses
-    mode?: 'js' | 'json';
-  }
+    /**
+     * The serialization mode to use for server functions.
+     * The "js" mode uses a custom binary format that is more efficient than JSON, but requires a custom deserializer (with `eval()`) on the client.
```
lxsmnsyc: probably need to mention the CSP effect

PR #2077, opened 2026-02-22
refactor(v2): use pnpm workspace for local packages

birkskyum:
Instead of relative paths for the start and nitro v2 plugin packages, we can just use the pnpm workspace.

The changes in the local plugins is still reflected on save, on the dev server of fixtures.

### DIFF
--- a/apps/tests/playwright.config.ts
+++ b/apps/tests/playwright.config.ts
@@ -5,15 +5,15 @@ export default defineConfig({
   testMatch: "**/*.test.ts",
 
   webServer: {
-    command: "pnpm run dev",
-    url: "http://localhost:3000",
+    command: "pnpm run dev --host 127.0.0.1 --port 3000 --strictPort",
+    url: "http://127.0.0.1:3000",
     reuseExistingServer: true,
     stdout: "pipe",
     stderr: "pipe",
   },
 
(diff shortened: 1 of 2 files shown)


lxsmnsyc:
hmmm, we probably also need `.npmrc` as a safety measure (despite the name, pnpm uses this):

```
prefer-workspace-packages=true
link-workspace-packages=true
```

birkskyum:
> hmmm, we probably also need `.npmrc` as a safety measure (despite the name, pnpm uses this):
> 
> ```
> prefer-workspace-packages=true
> link-workspace-packages=true
> ```

@lxsmnsyc  that won't be necessary. If the fixtures etc used

`"@solidjs/start": "2.0.0-alpha.1"`

Then the prefer-workspace [over the set version] would be needed. But the fixtures, landing, test etc. just have:

`"@solidjs/start": "workspace:*"`

So "prefer workspace" won't be preferred over anything - they're already set to look at the workspace first.

Browsing the pnpm lockfile shows this is working alreardy, with passages like:

```
      '@solidjs/vite-plugin-nitro-2':
        specifier: workspace:*
        version: link:../../../packages/start-nitro-v2-vite-plugin
```

PR #2091, opened 2026-03-04
WIP: Upgrade to solid 2 beta

brenelz:
## PR Checklist

Seeing what It take to update SolidStart to use Solid 2.0 beta. The examples work even with solid-router


lxsmnsyc:
We can probably use `^2.0.0-beta.3` based on my tests

lxsmnsyc:
needs rebasing

packages/start/src/shared/lazy.ts
```diff
@@ -1,52 +1,6 @@
...
-      },
-    };
-  };
-
-  return wrapper as T;
-};
-
-const lazy = !isServer
-  ? solidLazy
-  : <T extends Component<any>>(fn: () => Promise<{ default: T }>) => solidLazy(withAssets(fn));
```
katywings, line 50: Why are you removing all of my css work? 🙈

brenelz: I honestly can't remember. Maybe it had issues. Let me try and add it back

packages/start/src/server/handler.ts
```diff
@@ -13,6 +14,25 @@ import { handleServerFunction } from "../fns/handler.ts";
 import type { APIEvent, FetchEvent, HandlerOptions, PageEvent } from "./types.ts";
 import { getExpectedRedirectStatus } from "./util.ts";
 
+import { pageRoutes } from "./routes.ts";
+
+function buildDevManifest(): Record<string, any> {
```
brenelz: this is the most critical change that was made for v2 i think

packages/start/src/shared/lazy.ts
brenelz: added it back

katywings: Ahh I remember now 😅, Ryan is redoing all the css logic in Solid v2 core so the logic in start wont be necessary anymore. Unfortunately the new core logic seems to have many regressions, I gonna try to guide Ryan so that we can fix the holes 🙈.

packages/start/src/server/handler.ts
katywings: Where is this manifest gonna be used 😅?

brenelz: I don't think this exists anymore. Resolving

apps/tests/vitest.config.ts
```diff
@@ -1,15 +1,55 @@
...
     },
+    conditions: ["solid", "development", "browser"],
   },
   plugins: [solid()],
+  optimizeDeps: {
+    include: ["@solidjs/testing-library"],
+    esbuildOptions: {
+      plugins: [
+        {
+          name: "solid-compat",
```
brenelz, line 24: this can probably be removed

PR #2092, opened 2026-03-06
fix: explicitly set Content-Type to text/plain for json serialization mode

yinonburgansky:
Serialization mode:"json" uses Seroval special serialization format that is not valid JSON. Some runtimes (e.g. AWS Lambda) default to `application/json` when no `Content-Type` is set,  causing downstream `JSON.parse` errors such as: `SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data.`

When using "json" serialization mode, explicitly set `Content-Type` to `text/plain` for server-function responses to prevent incorrect parsing by runtimes. This restores the 2.x behavior, is safer for downstream environments, and is recommended best practice.


atilafassina (approved):

lxsmnsyc (approved):

brenelz:
Tests are failing

yinonburgansky:
> Tests are failing

they were already failing before, unrelated to the pr
if you look at the commits: https://github.com/solidjs/solid-start/commits/1.x/
they started failing from this commit: https://github.com/solidjs/solid-start/commit/a40029a9bf1936aaef9949e8d6dcc5d32c0904d1    
 (PR #2074 @atilafassina )

packages/start/src/runtime/server-runtime.ts
```diff
@@ -77,16 +77,16 @@ async function fetchServerFunction(
   const contentType = response.headers.get("Content-Type");
   const cloned = response.clone();
   let result;
-  if (contentType && contentType.startsWith("text/plain")) {
```
yinonburgansky: I debugged it stepping over line by line:
The bug stems from the fact this code assumes the content type will be empty for json type and will skip these 2 ifs which check if it is `text/plain` or `application/json`. 
The assumption doesn't hold due to the aws preset defaulting to `application/json` when content type is not set. which makes this code try to parse it as json even though it is not valid json (seroval json serialized format).
This didn't happen for `JS` serialization mode that set the content type to `text/javascript` which skips the 2 ifs and deserialize it correctly.
The fix consists of:
- Explicitly set content type so it won't be default set to the wrong content type (not a valid json).
- Move the serialization header check to the front.
I tested on my app - staging aws streaming server and it works now :)

PR #2094, opened 2026-03-08
test: add server-secret leak regression test

yinonburgansky:
Adds an integration test that scans source, server build output, and client build output for unique server-only secret markers. The test asserts that:

- markers exist in source files,
- each marker is defined once,
- markers are preserved in server output (not dead-code-eliminated),
- markers do not appear in client bundles (including .gz/.br assets).


yinonburgansky:
The output gzipped files which existed in 1.x are missing in 2.x.

PR #2104, opened 2026-03-14
Fix exports disappearing

### LINKED ISSUE #2100: [Bug?]: exported functions disappear
### Duplicates

### Latest version

### Current behavior 😯

ReferenceError: hello is not defined

### Expected behavior 🤔

"hello"

### Steps to reproduce 🕹

```ts
import type { APIHandler } from "@solidjs/start/server";

export const hello = () => "hello";

export const GET: APIHandler = async (e) => {
  return hello();
};
```

### Context 🔦

using v2 alpha

### Your environment 🌎

```shell

```

huseeiin:
fix #2100

### DIFF
--- a/.changeset/quiet-deserts-clap.md
+++ b/.changeset/quiet-deserts-clap.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/start": minor
+---
+
+fixed exports getting lost
--- a/packages/start/src/config/fs-routes/tree-shake.ts
+++ b/packages/start/src/config/fs-routes/tree-shake.ts
@@ -107,56 +107,73 @@ function treeShakeTransform({ types: t }: typeof Babel): PluginObj<State> {
                   });
                 }
               },
-              ExportDefaultDeclaration(exportNamedPath) {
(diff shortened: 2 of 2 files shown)


atilafassina:
this is one of the regressions from the Vinxi migration we always set on a fence about fixing.

While the disappearing export is an issue, a worse one would be to exposing server-only code accidentally to the client-bundle by adding an `export` in a file that is originally intended only for the server.

The workaround would be to export that variable somewhere else and have the API Route only consuming it instead.

---

Let's revisit this issue once #2070 and perhaps #2094  as well.

cc @lxsmnsyc

lxsmnsyc:
For the sake of loader modules this might be intentional. We probably would just recommend restructuring the module such that the said target exports is imported from a separate module, as mentioned by @atilafassina 

I'm inclined on a "wontfix"

brenelz:
So do we just close this pr?

lxsmnsyc:
I think we changed the behavior here. Needs re-testing.

brenelz:
Superseeded by this pr https://github.com/solidjs/solid-start/pull/2205

PR #2110, opened 2026-03-16
feat: add `env:*`

### LINKED ISSUE #2113: [Bug?]: 2.0.0-alpha.2 - process.env not available in server functions
### Duplicates

### Latest version

### Current behavior 😯

Loving the look of V2! I've noticed that when starting a project in dev mode with `vite dev`, none of the Private Environment Variables (those not prepended with VITE_) are available to Server Functions

.env
```
MY_SECRET=hello
```

routes/index.tsx
```
import { createAsync } from "@solidjs/router";

const myServerFn = async () => {
    "use server";

    if (typeof process.env.MY_SECRET === "undefined") {
        return "MY_SECRET is undefined";
    }

    return "OK";
};

export default function Component() {
    const result = createAsync(myServerFn);
    return (
        <>{result()}</> //Shows "MY_SECRET is undefined"
    );
}
```

The documentation on [Private Environment Variables](https://docs.solidjs.com/configuration/environment-variables#private-environment-variables) suggests that this should be OK.

### Expected behavior 🤔

`process.env`should be available in dev mode

### Steps to reproduce 🕹

Steps:

1.
2.
3.
4.

### Context 🔦

_No response_

### Your environment 🌎

```shell

```

lxsmnsyc:
This PR adds the `env:server` and `env:client` modules

```js
// Loads SERVER_* from .env
// Throws and omits all detected variables on client runtime
import { SERVER_EXAMPLE } from 'env:server';

console.log(SERVER_EXAMPLE);

// Loads CLIENT_* from .env
import { CLIENT_EXAMPLE } from 'env:client';

console.log(CLIENT_EXAMPLE);

// Dynamically loads environment variables (only on server)
import env from 'env:server/runtime';

console.log(env.NODE_ENV);
```

closes #2113

### DIFF
--- a/.changeset/smooth-foxes-taste.md
+++ b/.changeset/smooth-foxes-taste.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/start": minor
+---
+
+feat: add `env:*` runtime environment variables support


eru123:
Nice, this is a great feat, ensuring server and client env is loaded differently!

PR #2116, opened 2026-03-18
Fix Cloudflare preset not generating functioning applications

### LINKED ISSUE #2115: [Bug?]: `nitroV2Plugin` doesn't generate a valid output with Cloudflare presets
### Duplicates

### Latest version

### Current behavior 😯

When using the `"@solidjs/vite-plugin-nitro-2"` plugin Cloudflare presets (like `cloudflare-module`) don't generate a functioning application.

They generate an application that when run via `npx wrangler dev` returns a 500 response:
screenshot of 500 response from browser

Note that using directly the `nitro` plugin from `"nitro/vite"` does instead.

### Expected behavior 🤔

The `"@solidjs/vite-plugin-nitro-2"` plugin would produce valid Cloudflare Workers in the same way the official `nitro` plugin does. (especially since the "@solidjs/vite-plugin-nitro-2"` plugin does accept a `preset` in its options parameter).

### Steps to reproduce 🕹

Steps:

1. Create a new solid-start app via `npm create solid`, select:
   (the options not listed are not relevant)
     - ◇  What type of project would you like to create?
         │  SolidStart
     - ◇  Which version of SolidStart?
         │  v2 (pre-release, recommended)
     - ◇  Which template would you like to use?
         │  basic
2. cd into the project's directory, run `npm i`
3. in the `/vite.config.ts` file add the `cloudflare-module` preset to the `nitro` plugin:
    ```diff
   import { defineConfig } from "vite";
   import { nitroV2Plugin as nitro } from "@solidjs/vite-plugin-nitro-2";

   import { solidStart } from "@solidjs/start/config";

   export default defineConfig({
     plugins: [solidStart(),
   -   nitro()
   +   nitro({ preset: 'cloudflare-module' })
     ]
   });
    ```
4. run `npm run build`
5. run `npx wrangler dev .output/server/index.mjs --assets .output/public` (as suggested by the `npm run build` output)
6. Try visiting the page and it won't work

### Context 🔦

_No response_

### Your environment 🌎

```shell
System:
    OS: macOS 26.3.1
    CPU: (10) arm64 Apple M1 Pro
Binaries:
    Node: 22.21.1
    Yarn: 1.22.22
    npm: 10.9.4
    pnpm: 10.30.2
npmPackages:
   @solidjs/start: 2.0.0-alpha.2
   @solidjs/vite-plugin-nitro-2: 0.1.0
   (nitro: 3.0.260311-beta)
```

dario-piotrowicz:
## PR Checklist

- [x] Addresses an existing open issue: fixes #2115
  👆 No, I don't think that there are tests for the plugin package, if I'm wrong and some tests should be added please do let me know 🙂  

Setting up a Cloudflare preset in the plugin's configuration object results in builds not producing a functioning result (see https://github.com/solidjs/solid-start/issues/2115)

## What is the new behavior?

A Cloudflare preset can be passed to the plugin's configuration object and builds do produce functioning applications.

## Other information

To manually test the changes, create a solid start app via `npm create solid`, apply one of the cloudflare presets, run  `npm run build` follows by the `wrangler dev` command that the build command presents in its output and see that the resulting application does not render. Try again but using this PRs pkg-pr-new and see that now it works as intended.

Note: I only tested the changes with `cloudflare-module` since, as far as I know, it's the generally preferred preset for Cloudflare deployments


themavik (commented):
Reviewed the changes — the implementation looks correct and addresses the issue well.

themavik (commented):
**What this PR does:** Fixes Cloudflare preset builds by reconstructing full URLs from headers and using a custom virtual entry instead of fromWebHandler, which fails when Nitro strips URLs to path-only.

**Done well:** The getCloudflareVirtualEntryContent correctly rebuilds the URL from host, x-forwarded-proto, and path. The getHeader helper handles both Headers-like and plain object header formats. Setting noExternals for Cloudflare presets avoids conditional export resolution issues. Good separation in cloudflare.ts.

**Edge case:** The path construction uses event.path || event.req.url - consider x-forwarded-host for proxied setups.

brenelz (approved):

brenelz:
Is it possible to add tests for this? Not sure how feasible with it being so tied to cloudflare

brenelz (changes requested):
Im told this will break some other adapters

atilafassina:
> Im told this will break some other adapters

can you please elaborate, @brenelz / @huseeiin  ?
changes seem scoped to Cloudflare presets to me

huseeiin:
> can you please elaborate, @brenelz / @huseeiin ? changes seem scoped to Cloudflare presets to me

i said it will break netlify i was wrong it will break https://github.com/solidjs/solid-start/pull/2056 because noExternals: isCloudflarePreset(preset) will make noExternals only true if using cloudflare

dario-piotrowicz:
> Is it possible to add tests for this? Not sure how feasible with it being so tied to cloudflare

I asked the same in the PR's description 🙂 

I think that the logic of the plugin is not currently being tested and having full e2e/integration tests for this could be a bit involved 🤔

But If we wanted to, I think we could somewhat easily have some unit tests to ensure that the various functions return the expected values 

Would you like me to give that a go and actually set up tests for the package here?

atilafassina:
@dario-piotrowicz  / @huseeiin 

I'm still seeing `500`s in Start v2. Both in nitro v3 and in this version of the plugin.

in the Nitro v2 plugin (this PR), we get multiple issues related to `noExternal` (imports getting blocked because `sideEffects`)

in Nitro v3, I get `{"status":500,"unhandled":true,"message":"HTTPError"}`. I couldn't find the full stack trace in CF dashboard (🙈 ) but I ran `pnpx wrangler dev` and got

```sh
[wrangler:info] ✨ Parsed 1 valid header rule.
[wrangler:info] Ready on http://localhost:8787
✘ [ERROR] Error: Attempting to use server context in non-server build
```

**UPDATE**

this config works
```ts
import { nitro } from "nitro/vite";

import { solidStart } from "@solidjs/start/config";

export default defineConfig({
  plugins: [solidStart(), nitro()],
  nitro: {
    preset: "cloudflare_module",
    exportConditions: ["worker"]
  },
});
```

big thanks  to @pi0

brenelz:
So what should we do with this pr?

birkskyum:
The custom nitro v2 vite plugin has been deprecated in favor of the official Nitro v3 vite plugin

PR #2117, opened 2026-03-19
fix: strip baseURL prefix before matching API routes

### LINKED ISSUE #2106: [Bug?]: API routes not matched when server.baseURL is set
### Duplicates

### Latest version

### Current behavior 😯

When `server.baseURL` is configured in the Vinxi/SolidStart config (e.g. `server: { baseURL: '/admin' }`), API routes defined in `src/routes/` are never matched. Requests to `/admin/api/my-route` fall through to the page renderer and return a 404 HTML response instead of executing the API handler.

The root cause is in `@solidjs/start/dist/server/handler.js`:

```js
// line 29
const match = matchAPIRoute(new URL(event.request.url).pathname, event.request.method)
```

`event.request.url` contains the **full** pathname including the `baseURL` prefix — e.g. `/admin/api/my-route`. However, `matchAPIRoute` looks up routes registered from the file system, which are stored **without** the prefix (e.g. `/api/my-route`). The lookup always fails, so all API routes become unreachable.

### Expected behavior 🤔

API routes defined in `src/routes/api/` should be matched and executed correctly regardless of the `server.baseURL` setting. `matchAPIRoute` should strip the `server.baseURL` prefix from the pathname before performing the route lookup — the same way `@solidjs/router`'s `base` prop strips it for page route matching on the client.

### Steps to reproduce 🕹

Steps:

1. Create a SolidStart app with `server.baseURL` set:
```ts
// app.config.ts
import { defineConfig } from '@solidjs/start/config'
export default defineConfig({
  server: { baseURL: '/app' },
})
```

2. Add `base` to the Router so page navigation works:
```tsx
// src/app.tsx
<Router base={import.meta.env.SERVER_BASE_URL}>
  <FileRoutes />
</Router>
```

3. Create an API route:
```ts
// src/routes/api/hello.ts
export function GET() {
  return new Response(JSON.stringify({ ok: true }), {
    headers: { 'Content-Type': 'application/json' },
  })
}
```

4. Start the dev server and request `GET /app/api/hello`

**Result:** `404` — the SPA catch-all (`[...404]`) renders instead of the API handler.  
**Expected:** `200 { "ok": true }`

### Context 🔦

**Workaround:** Move the route file to mirror the full path including the base — `src/routes/app/api/hello.ts` — so the registered route path `/app/api/hello` matches what `matchAPIRoute` receives. This is unergonomic and forces duplication of the base prefix in the file structure.

### Your environment 🌎

```shell

```

themavik:
## Summary
Fixes #2106.
**Root cause:** `matchAPIRoute` used the full pathname (e.g. `/admin/api/my-route`) while routes are registered without the base prefix (e.g. `/api/my-route`). When `server.baseURL` is configured, the base prefix was not stripped before lookup, causing all API routes to return 404.
**Fix:** Strip the `baseURL` prefix from the pathname before matching API routes.
## Changes
- `packages/start/src/server/handler.ts`: Updated `stripBaseUrl` to use `SERVER_BASE_URL` (from `server.baseURL`) when available, with fallback to `BASE_URL`
- `packages/start/src/config/index.ts`: Added `import.meta.env.SERVER_BASE_URL` to the define block when `server.baseURL` is configured
- `packages/start/src/env.d.ts`: Uncommented `SERVER_BASE_URL` type definition
## Testing
- Verified fix addresses reported 404 scenario with baseURL
- Change is minimal and follows existing code patterns

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changeset/fix-baseurl-api-routes.md
+++ b/.changeset/fix-baseurl-api-routes.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/start": patch
+---
+
+Strip baseURL prefix before matching API routes.
--- a/packages/start/src/config/index.ts
+++ b/packages/start/src/config/index.ts
@@ -73,7 +73,7 @@ export function solidStart(options?: SolidStartOptions): Array<PluginOption> {
           },
         };
       },
-      async config(_, env) {
+      async config(config, env) {
         const clientInput = [handlers.client];
         if (env.command === "build") {
           const clientRouter: BaseFileSystemRouter = (globalThis as any).ROUTERS.client;
@@ -143,6 +143,9 @@ export function solidStart(options?: SolidStartOptions): Array<PluginOption> {
             "import.meta.env.START_CLIENT_ENTRY": JSON.stringify(handlers.client),
             "import.meta.env.START_DEV_OVERLAY": JSON.stringify(start.devOverlay),
             "import.meta.env.SEROVAL_MODE": JSON.stringify(start.serialization?.mode || "json"),
+            "import.meta.env.SERVER_BASE_URL": JSON.stringify(
+              (config as { server?: { baseURL?: string } })?.server?.baseURL ?? "",
+            ),
           },
           builder: {
             sharedPlugins: true,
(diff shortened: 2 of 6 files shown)


themavik (commented):
Reviewed the changes — the implementation looks correct and addresses the issue well.

brenelz (approved):

brenelz:
Could you please add a changeset and if possible add some tests?

themavik:
Added changeset and unit tests for stripPathBase (new strip-path-base.spec.ts).

PR #2126, opened 2026-04-03
Use `h3/generic` for runtime agnostic methods

huseeiin:
now cloudflare works when not using nitro.
previously doing `wrangler dev dist/server/entry-server.js` (when using `@cloudflare/vite-plugin`, of course) the server would return a srvx `NodeResponse` instead of a standard `Response` which doesn't work.
## PR Checklist


brenelz:
Are we good with this change?

lxsmnsyc (approved):

huseeiin:
@brenelz it seems "node" is still picked even when using "/generic". weird.
its not because of `externalConditions: ["solid", "node"]` or `noExternal: ["@solidjs/start", "h3", "cookie-es"]`, i tested

PR #2127, opened 2026-04-04
fix: server functions inclusivity + runtime exports

lxsmnsyc:
- Fix server functions to allow configurable file filters. I neglected to add this on the original PR.
- Fix runtime to be exported package rather than accessing it directly by file. Resolution-related issues so this is much better and direct to the point

I've also moved  and renamed server fn files since the current ones are too confusing


packages/start/src/fns/server.ts
katywings, line 22: Running this branch in dev (with vite 8) I now get the following error 🙈:

`TypeError: Cannot read properties of undefined (reading 'BASE_URL')
    at query (file:///...node_modules/.pnpm/...@solidjs/start/dist/fns/server.js:12:35)`

lxsmnsyc: @katywings got that initially too, but I tried running again and it worked. Did you also try clearing node_modules?

katywings: Just did a full clear and multiple restarts of the dev server, the error occurs every time 🤔

katywings: I think it might be some duplicated code being run here yet again. A top level console.log in registrations is executed twice during server init. I do only have one @solidjs/start in node_modules though... Not sure yet where the second instance is coming from - but I do have some idea, let me check...

katywings: Yeah no... If I had to guess my thesis is that this has something to do with the whole server functions preload shenanigans, if the preload means that the code doesn't go through vite's loading behavior that might by the reason for duplicate instances and also for missing `import.meta.env`. Unfortunately I do not really have the time right now to dig deeper 😅

brenelz:
Whats the status of this pr?

lxsmnsyc:
@brenelz not much, except maybe there's a recent push on main that needs to be synced. I'll probably take a look later this week

packages/start/src/fns/server.ts
katywings: The `TypeError: Cannot read properties of undefined (reading 'BASE_URL') at query (file:///...node_modules/.pnpm/...@solidjs/start/dist/fns/server.js:12:35)` error can be fixed by setting:

```ts
resolve: {
  noExternal: ['@solidjs/start']
},
```

I wonder if it makes sense to set this config by default in our plugin :thinking:

katywings (approved):

PR #2136, opened 2026-04-12
fix: move esbuild to devDependencies

### LINKED ISSUE #2135: esbuild should not be a production dependency
## Description

`@solidjs/start@2.0.0-alpha.2` lists `esbuild` in `dependencies`, which causes it to be installed in production environments even though it's only needed at build time.

https://github.com/solidjs/solid-start/blob/4cc999b698abcf4d934e2e0e24e9dd13be454c42/packages/start/package.json#L51

## Why this matters

1. **Docker image bloat**: esbuild ships a platform-specific Go binary (~9 MB). In multi-stage Docker builds that run `npm ci --omit=dev` for the production stage, esbuild still gets installed because it's a transitive production dependency of `@solidjs/start`.

2. **False-positive security vulnerabilities**: Security scanners (e.g., Trivy) flag Go stdlib CVEs in the esbuild binary. Since these CVEs exist in the Go binary that ships with esbuild and not in any JavaScript code, they are not exploitable at runtime — but they still require triage and `.trivyignore` entries. Current examples: CVE-2025-58183, CVE-2025-61726, CVE-2025-61728, CVE-2025-61729, CVE-2025-68121, CVE-2026-25679.

3. **Not needed at runtime**: The SolidStart build output (`.output/server/`) is a standalone bundle. esbuild is only imported in build-time config code (`src/config/fs-routes/router.ts`) — it is never used by the production server.

## Suggested fix

Move `esbuild` from `dependencies` to `devDependencies` (or `peerDependencies` with `optional: true`), since it's only used during the Vite build step.

## Additional note

`vite` is listed as a non-optional `peerDependency`, which has a similar effect — it forces vite to be installed in production even though the built output doesn't need it. Making it `"optional": true` in `peerDependenciesMeta` would allow `npm ci --omit=dev` to skip it.

## Environment

- `@solidjs/start`: 2.0.0-alpha.2
- Node.js: 24.x
- npm: 11.x

frenzzy:
## Summary

- Move `esbuild` from `dependencies` to `devDependencies` in `packages/start/package.json`

## Motivation

`esbuild` is only imported in build-time Vite plugin code (`src/config/fs-routes/router.ts`) and is never used by the production server output (`.output/server/`). Having it in `dependencies` causes:

1. **Docker image bloat**: esbuild ships a platform-specific Go binary (~9 MB) that gets installed even with `npm ci --omit=dev`
2. **False-positive security findings**: Security scanners (e.g., Trivy) flag Go stdlib CVEs in the esbuild binary that are not exploitable at runtime

Since `vite` (which itself depends on `esbuild`) is already a `peerDependency`, esbuild will always be available during the build step.

Closes #2135

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/packages/start/package.json
+++ b/packages/start/package.json
@@ -47,7 +47,6 @@
     "defu": "^6.1.4",
     "error-stack-parser": "^2.1.4",
     "es-module-lexer": "^2.0.0",
-    "esbuild": "^0.27.4",
     "fast-glob": "^3.3.3",
     "h3": "2.0.1-rc.16",
     "html-to-image": "^1.11.13",
@@ -68,6 +67,7 @@
     "node": ">=22"
   },
   "devDependencies": {
+    "esbuild": "^0.27.4",
     "@types/babel__core": "^7.20.5",
     "@types/node": "^25.5.0",
     "vite": "^7.3.1",


agoldstein03 (changes requested):
check CI, need to update lockfile

katywings:
As explained in my comment https://github.com/solidjs/solid-start/issues/2135#issuecomment-4818326271, this is the wrong approach to get rid of `esbuild` in production envs, as `@solidjs/start` depends on it 😅. An alternative solution is included in said comment.

PR #2143, opened 2026-05-12
fix(1.x) - hash server function ids

lxsmnsyc:
This PR obfuscates function ids generated for server functions by hashing (using xxhash32). This is to hide the file structure being exposed by the transform.


packages/start/config/index.js
```diff
@@ -37,6 +38,11 @@ function solidStartServerFsRouter(config) {
     );
 }
 
+function getFunctionId(id) {
+  console.log(id);
```
mizulu:
```suggestion

```

brenelz (approved):

PR #2147, opened 2026-05-20
Patch css hot reload ss2

madaxen86:
## PR Checklist

- [x] Addresses an existing open issue: fixes [#19818 at tailwindlabs](https://github.com/tailwindlabs/tailwindcss/issues/19818)

When adding classes to file routes/layouts tailwind does not hot reload the css file. 

This PR invalidates css modules when a file route is changed.

## REPRODUCTION
1. pnpm create solid,  choose `with-tailwindcss` template
2. go to `src/routes/index.tsx`
3. change e.g. h1 class `text-sky-500` to `text-purple-500`
==> blue color is removed ✅, purple is not added ❌

### DIFF
--- a/packages/start/src/config/fs-routes/fs-watcher.ts
+++ b/packages/start/src/config/fs-routes/fs-watcher.ts
@@ -1,9 +1,4 @@
-import type {
-  EnvironmentModuleNode,
-  FSWatcher,
-  PluginOption,
-  ViteDevServer,
-} from "vite";
+import type { EnvironmentModuleNode, FSWatcher, PluginOption, ViteDevServer } from "vite";
 import { VITE_ENVIRONMENTS } from "../constants.ts";
 import { moduleId } from "./index.ts";
 import type { BaseFileSystemRouter } from "./router.ts";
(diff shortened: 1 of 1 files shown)


katywings:
Hey @madaxen86, thank you a lot for the fix! What are the exact reproduction steps needed, were the hot reload for css did not work? You linked an issue regarding Tailwind, does this only relate to Tailwind or can it be reproduced without it? 

When looking at the implementation, it seems as if this fix reloads each and every css file, not only the ones related to routes? Why is it necessary to reload all css 😅?

madaxen86:
Added the repro to the description.

NOTE: PR is not ready to be merged. Just needed a pkg.pr.new to test on stackblitz.
Interestingly, this PR works locally but not on stackblitz!?

This is specifically related to tailwindcss (and probably pandacss) where there are no css files changed directly, you just apply the atomic classes in tsx/jsx files.

Hot updates work with Vite v6.

I am still investigating and I'll ping you when I find a solutions that's actually applicable. Reloading all css modules is overkill of course. I just don't get why even that does not work in stackblitz!?

madaxen86:
I digged into it and it's very ambiguous.
Updating the @tailwindcss/vite plugin totally breaks the hot reloading. But not just for solid-start, also for tanstack-start-solid.
Invalidating the app.css or even all css modules does also not work anymore. 
The only thing that triggers and update is to save the app.css or "fake" it with a vite plugin.

packages/start/src/config/fs-routes/fs-watcher.ts
```diff
@@ -31,21 +26,31 @@ function createRoutesReloader(
...
       const seen = new Set<EnvironmentModuleNode>();
       devEnv.moduleGraph.invalidateModule(mod, seen);
     }
-
+    
     if (environment !== "ssr") {
       if (mod) {
         devEnv.reloadModule(mod);
+        
+        const extensions = [".css", ".scss", ".sass.", ".less"];
```
lxsmnsyc, line 43: not a requirement, but we could probably use a RegExp here (`/\.(c|sc|sa|le)ss$/i`)

katywings: @lxsmnsyc I think we can close this PR 😅, its superseded by #2171

lxsmnsyc: yeah I believe it solves both problems lol

PR #2152, opened 2026-05-31
fix: thread Vite `base` through SSR manifest and Nitro baseURL

### LINKED ISSUE #2151: SolidStart 2.0 (alpha) ignores Vite `base` for prerendered/SSR script and asset URLs
> _written by claude_

## Summary

Under any non-root Vite `base` (e.g. `base: "/repo/"` for a GitHub Pages subpath deploy), `@solidjs/start@2.0.0-alpha.*` produces SSR HTML whose `<script>`, `<link rel="modulepreload">`, and `<link rel="stylesheet">` tags are root-absolute (`/_build/...`) instead of base-prefixed (`/repo/_build/...`). The page returns 200 from the right URL, but every asset 404s, hydration never starts, and the app appears blank.

There's also a secondary papercut: `@solidjs/vite-plugin-nitro-2` defaults Nitro's `baseURL` to `/` regardless of Vite's `base`, so users have to write the prefix twice — once as `base` in their Vite config and again as `nitroV2Plugin({ baseURL: "/repo/" })` — to get Nitro to prerender and serve at the right path.

## Reproduction

1. Scaffold a SolidStart 2.0 project, set `base: "/sub/"` in `vite.config.ts`.
2. `pnpm build` → `node .output/server/index.mjs`.
3. `curl http://localhost:3000/sub/` → HTML returns, but every asset reference is `/_build/...` instead of `/sub/_build/...`.

## Where the URLs are emitted

- `packages/start/src/server/manifest/prod-ssr-manifest.ts`:
  - `path(id)` → `join("/", viteManifestEntry.file)`
  - `getAssets(id)` → `href: "/" + asset`
  - `json()` → `output: join("/", ...)`
- `packages/start/src/server/manifest/dev-client-manifest.ts`:
  - `import(join("/", id))` (same bug in dev under a base)
- `packages/start-nitro-v2-vite-plugin/src/index.ts`:
  - `resolvedNitroConfig` has no `baseURL`, so Nitro defaults to `/`.

The dev-ssr manifest already does the right thing (`normalize(join(import.meta.env.BASE_URL, id))`), which is the pattern to copy.

## v1 comparison

In v1, Vinxi owned the build manifest and emitted `output.path` with the router base already prefixed, so `StartServer` could read it raw and it Just Worked. The v2 migration off Vinxi dropped that wiring. The fix isn't new API — it's restoring base-awareness in the two spots that lost it.

## Proposed fix

PR linked below. Three small commits:

1. Failing unit tests for `prod-ssr-manifest`.
2. Replace `"/"` with `import.meta.env.BASE_URL` in the prod SSR manifest and the dev client manifest (using `pathe.join`, which collapses to `/foo` when base is `/`).
3. Default `nitroConfig.baseURL` to the resolved Vite `config.base` in the nitro-2 plugin. User-supplied `baseURL` still wins.

## Out of scope (follow-up)

`<Router>` in the user's `app.tsx` still requires manual `base={import.meta.env.BASE_URL.replace(/\/$/, "")}`. Auto-passing it from `StartClient`/`StartServer` would close the loop but touches the router integration and is worth a separate discussion.

bigmistqke:
Hey all 👋 

I wanted to deploy `solidbase` to gh-pages, but I needed to add basePath awareness to `SolidStart` so I vibecoded it with `claude`.

_[after this line all is written by claude]_

Fixes #2151.

## Summary

Under a non-root Vite `base`, the prod SSR manifest emits root-absolute URLs for scripts, modulepreload, and stylesheets, and `@solidjs/vite-plugin-nitro-2` ignores the same base for Nitro routing. Subpath deploys (GitHub Pages, reverse-proxied paths, embedded mounts) ship HTML that 404s every asset.

This restores the v1 behavior — base flows through the build manifest — without changing any public API or adding a new helper.

## Changes

- **`prod-ssr-manifest.ts`** — replace `join("/", ...)` with `join(import.meta.env.BASE_URL, ...)` in `path`, `getAssets`, and `json`. Matches the existing pattern in `dev-ssr-manifest.ts`.
- **`dev-client-manifest.ts`** — same fix for the dev-mode dynamic import.
- **`vite-plugin-nitro-2`** — capture Vite's resolved `config.base` in `configResolved` and use it as the `nitroConfig.baseURL` default. A user-supplied `nitroV2Plugin({ baseURL })` still wins because the spread comes after.

`pathe.join` collapses correctly when `BASE_URL === "/"`, so the no-base case is unchanged (verified with a smoke build of `apps/tests`).

## Tests

New `packages/start/src/server/manifest/prod-ssr-manifest.spec.ts` covers:

- `path(id)` returns BASE_URL-prefixed paths.
- `getAssets(id)` hrefs all sit under BASE_URL.
- `json()` output paths are BASE_URL-prefixed.
- BASE_URL `"/"` still emits single-slash URLs.

6/6 pass after the fix.

## Manual verification

`apps/tests` built twice — once with `base: "/sub/"`, once default. With base, every script/link tag is correctly prefixed and the Nitro server listens at `/sub`. Without base, output is byte-identical to main.

## What this PR does NOT cover

This fixes framework-emitted URLs. User-authored root-absolute paths still need to be base-aware on the user's side — the framework can't safely rewrite arbitrary string literals in user code. Specifically:

- **`entry-server.tsx` static `<head>` references** — `<link rel="icon" href="/favicon.ico" />`, `<link rel="manifest" href="/site.webmanifest" />`, `<meta property="og:image" content="/og.png" />`, etc. These live in user JSX, not the build manifest, so they ship as-is. Authors should use `${import.meta.env.BASE_URL}favicon.ico` (or a small helper).
- **User-authored `<a href="/...">`** — Solid Router's `<A>` prepends base; plain `<a>` does not unless the click is intercepted. Recommend `<A>` for in-app navigation.
- **CSS `url(/foo)` and ``** — same story. Vite handles asset imports (`import logo from "./logo.png"`) and `new URL("./asset", import.meta.url)` correctly; raw string paths are the user's responsibility.
- **Sitemap / robots / canonical URLs** generated by themes or user code — these typically depend on a separate `siteUrl` config, not Vite `base`.

A follow-up could ship a tiny `withBase(path)` helper from `@solidjs/start` to make the user-side calls less noisy, but that's a separate discussion.

## Out of scope (also follow-up)

- Auto-passing `base` to `<Router>` in `app.tsx`. Touches the router integration; happy to follow up if maintainers want it.
- Any change to `<Router base>` semantics in `@solidjs/router`.

## Commits

1. `0147d17` test(start): fail when prod SSR manifest ignores deploy base
2. `e872ac8` fix(start): thread BASE_URL through SSR manifest paths
3. `4bb7d84` fix(vite-plugin-nitro-2): inherit Vite config.base as the Nitro baseURL


brenelz (approved):

lxsmnsyc:
Is this only necessary for nitro v2? what is the behavior for v3

birkskyum:
The Nitro v2 vite plugin is now deprecated. If you experience this on Nitro v3 with SolidStart v2 as well, feel free to open a new ticket

PR #2153, opened 2026-06-01
solid start alpha running with deno

ashtonjurgens:
When trying to get solid start alpha running with deno, I encountered a few problems. This is an informational PR of what I had to change to get things running. I can't guarantee nothing has been broken, (especially with 3efff0c63a2e42f7810b979752ba27c430a74a14) which is why I'm calling it an informational PR.

There are 3 commits (all very small), and each commit has a thorough message to explain the change.


packages/start/src/config/manifest.ts
```diff
@@ -16,6 +16,8 @@ export function manifest(start: SolidStartOptions): PluginOption {
       devServer = server;
     },
     async resolveId(id) {
+      if (id.startsWith("/@manifest/")) return id;
```
brenelz, line 20: this seems like it would have issues?

birkskyum:
Things have improved since. Where deno 2.6.3 breaks, deno 2.9.3 works much better. If you still experience issues with out latest pre-release SolidStart 2.0.0-beta.6 or newer, and latest deno, and a fresh template that use nitrov3, then feel free to open a new ticket.

PR #2156, opened 2026-06-11
feat: routeTransforms option for non-TS route languages

### LINKED ISSUE #1549: [Bug?]:  Page components - weird behaviour with transpiling plugins
### Duplicates

### Latest version

### Current behavior 😯

Page components (in `/routes` directory) don't get compiled using plugins configured in `app.config.ts`, unlike other components.

### Expected behavior 🤔

Just like any other component, page components should be transpiled by application plugins.

### Steps to reproduce 🕹

Steps:

1. Create a new Solid Start application.
2. Install `@danielx/civet` library.
3. Setup config:
```ts
// app.config.ts

import civetVitePlugin from '@danielx/civet/vite'
import { defineConfig } from '@solidjs/start/config'

export default defineConfig({
  extensions: ['civet', 'tsx', 'ts'],
  vite: {
    plugins: [
      civetVitePlugin({
        ts: 'civet',
      }),
    ]
  }
});

```
4. Create a civet page in `/routes` directory.
```civet
export default function Civet
  <>
    This is a Civet page.<br />
    This doesn't work.<br />
    This would be the perfect way to write Civet pages.

```
5. Minimal reproduction repo: https://codesandbox.io/p/github/IgorSzymanski/solid-start-plugin-issue/civet-plugin

### Context 🔦

This a minor issue, because there is a workaround, but the behaviour is very weird and inconsistent.
I use `@danielx/civet/vite` plugin to write my SolidJS application using Civet syntax (https://civet.dev/). 

> Civet is a programming language that compiles to [TypeScript](https://www.typescriptlang.org/) or [JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript), so you can [use existing tooling](https://civet.dev/integrations) (including VSCode type checking, hints, completion, etc.) while enabling concise and powerful syntax. It starts with [99% JS/TS compatibility](https://civet.dev/comparison), making it easy to transition existing code bases.

Basically, the plugin actually works, as I can write Civet components (as seen in the reproduction repo) and they get transpiled to TS just fine. The only issue is that page components (or route components) get ignored and they're treated like they're just components written in TS. Perfectly, page components should get the same treatment that other components get and if they're written in Civet, they should be transpiled as well.

There is a workaround - I can just write `.civet` components, but stick to TS syntax in page components (which is still compatible with civet), co it's in no way a critical issue, but an inconvenience nevertheless.

I don't know the codebase of Solid Start, so I'm not sure what's causing the issue or what could potentially fix it.

### Your environment 🌎

_No response_

STRd6:
Fixes #1549 (on `main`/v2; the vinxi-based 1.x line has the same issue — happy to backport, see notes below).

## Problem

Route modules are analyzed for their exports by reading the file off disk and parsing it with esbuild's tsx loader (`analyzeModule` in `fs-routes/router.ts`, vendored from vinxi). Languages that *compile to* JS/TS — Civet, CoffeeScript, etc. — can't be parsed that way, so `toRoute` throws, `addRoute` swallows the error, and the route is **silently dropped** even though the user registered the extension via `extensions` and added a bundler plugin that handles the language everywhere else.

Reported here as #1549 and to Civet as [DanielXMoore/Civet#1319](https://github.com/DanielXMoore/Civet/issues/1319). Civet currently works around it by monkey-patching `toRoute` and substituting `fs.readFileSync` during analysis ([DanielXMoore/Civet#2148](https://github.com/DanielXMoore/Civet/pull/2148)) — this PR is the seam that makes that workaround unnecessary.

## Approach

A per-extension source transform that runs before export analysis, so all the existing pick/route-config/HTTP-method logic is reused and the transform author never sees es-module-lexer's shapes:

```ts
import civet from "@danielx/civet";

solidStart({
  extensions: ["civet"],
  routeTransforms: {
    civet: (code, sourcePath) => civet.compile(code, { filename: sourcePath, sync: true }),
  },
})
```

- `FileSystemRouterConfig` gains `routeTransforms?: Record<string, RouteModuleTransform>` (extension without dot, matching how `extensions` is spelled), threaded through `solidStart()`.
- `analyzeModule(src)` is split into `analyzeCode(code, sourcePath)` plus a new `analyzeRouteModule(sourcePath, config)` that applies the matching transform first (and `await init`s the lexer, which watcher-triggered updates could otherwise race).
- `toRoute` becomes async; its only callers (`addRoute` / `updateRoute`) were already async and now await it.

The hardcoded `.md`/`.mdx` carve-out in `toRoute` is untouched, but could conceivably migrate to this mechanism later.

## Verification

`tsc --noEmit` on `packages/start` is clean. `vitest run` shows 2 failures that also fail for me on unmodified `main` (server-fns related), 5 passing.

## Notes

- Naming (`routeTransforms`, `sourcePath`) and shape are open to feedback — e.g. this could alternatively hang off `extensions` entries in a tuple form like vite-plugin-solid's `extensions` option.
- The released 1.x line hits the same problem through vinxi's `analyzeModule`; since `toRoute` lives in this repo there too, a near-identical backport is possible without touching vinxi.

### DIFF
--- a/.changeset/route-transforms-non-ts-languages.md
+++ b/.changeset/route-transforms-non-ts-languages.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/start": minor
+---
+
+Add `routeTransforms` option: per-extension source transforms applied before route modules are analyzed for their exports, so compile-to-JS languages (Civet, CoffeeScript, ...) can be used in the route directory
--- a/packages/start/src/config/fs-router.ts
+++ b/packages/start/src/config/fs-router.ts
@@ -1,6 +1,6 @@
 import type { ExportSpecifier } from "es-module-lexer";
 import {
-  analyzeModule,
+  analyzeRouteModule,
   BaseFileSystemRouter,
   cleanPath,
   type FileSystemRouterConfig,
@@ -25,7 +25,7 @@ export class SolidStartClientFileRouter extends BaseFileSystemRouter {
     return routePath?.length > 0 ? `/${routePath}` : "/";
   }
 
-  toRoute(src: string) {
+  async toRoute(src: string) {
     const path = this.toPath(src);
 
     if (src.endsWith(".md") || src.endsWith(".mdx")) {
@@ -41,7 +41,7 @@ export class SolidStartClientFileRouter extends BaseFileSystemRouter {
       };
     }
 
-    const [_, exports] = analyzeModule(src);
+    const [_, exports] = await analyzeRouteModule(src, this.config);
     const hasDefault = !!exports.find(e => e.n === "default");
     const hasRouteConfig = !!exports.find(e => e.n === "route");
     if (hasDefault) {
@@ -114,7 +114,7 @@ export class SolidStartServerFileRouter extends BaseFileSystemRouter {
     return routePath?.length > 0 ? `/${routePath}` : "/";
   }
 
-  toRoute(src: string) {
+  async toRoute(src: string) {
     const path = this.toPath(src);
     if (src.endsWith(".md") || src.endsWith(".mdx")) {
       return {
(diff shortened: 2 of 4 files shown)


brenelz:
Closing as stale with conflicts

PR #2165, opened 2026-06-25
feat(config): add server-only and client-only virtual module plugin

### LINKED ISSUE #2162: [Feature?]: Built-in server-only & client-only virtual module
### Duplicates

### Latest version

### Summary 💡

Add a built-in server-only virtual module that throws a build-time error when imported in client-side code. When imported inside a "use server" file or any server-side module, it is a no-op.

This can be implemented as a first-class Vite plugin inside `@solidjs/start/config`:

``` typescript
{
    name: 'server-only',
    resolveId(id, importer, { ssr }) {
      if (id !== 'server-only') return null;
      if (!ssr) {
        this.error(
          `[server-only] "${importer}" cannot import a server-only module on the client`,                                                                                                                                     
        );
      }
      return '\0server-only';
    },
    load(id) {
      if (id === '\0server-only') return 'export {}';
    },
  },
```
'client-only' is similar to this.

### Examples 🌈

``` typescript
import 'server-only';

// anything below is guaranteed never to reach the client bundle
export const db = createDatabaseClient();
```

### Motivation 🔦

SolidStart has no way to enforce that a module never reaches the client bundle. "use server" marks functions as server RPCs, but it does not prevent accidental client-side imports of modules containing secrets, DB credentials, or Node.js-only APIs.

The workaround today is a runtime check using import.meta.env.SSR, but that only throws in the browser at runtime — it won't catch the leak in CI or during pnpm build. A build-time error via a Vite resolveId hook is both trivial to implement and already proven in Next.js. Shipping it as part of @solidjs/start would give the same safety guarantee without requiring users to wire up the plugin themselves.

tsushanth:
Closes #2162

SolidStart has no build-time mechanism to prevent a module from being bundled on the wrong side. `"use server"` marks functions as RPCs but does not stop a module containing DB credentials or Node-only APIs from leaking into the client bundle. The only workaround today is a runtime `import.meta.env.SSR` check that only fires in the browser, too late to be caught in CI.

This PR adds a `solid-start:boundary-modules` Vite plugin to the array returned by `solidStart()`. It resolves two bare specifiers:

- `server-only` — calling `this.error()` during `resolveId` when `ssr` is false turns the import into a hard build error pointing at the importing file. In the SSR environment the specifier resolves to a virtual empty module.
- `client-only` — the mirror: errors in the SSR environment, resolves to an empty module on the client.

Both produce zero runtime overhead (empty export). The build error surfaces the violating file path, so the mistake is caught at `pnpm build` or in CI rather than at runtime in the browser.

The implementation follows the shape suggested in the issue and mirrors what Next.js ships in its `server-only` and `client-only` packages, with no new dependencies.

### DIFF
--- a/.changeset/fix-windows-preview-path-to-file-url.md
+++ b/.changeset/fix-windows-preview-path-to-file-url.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/start": patch
+---
+
+Convert absolute path to file:// URL before dynamic import in preview server to fix Windows compatibility
--- a/packages/start/src/config/dev-server.ts
+++ b/packages/start/src/config/dev-server.ts
@@ -1,4 +1,5 @@
 import { NodeRequest, sendNodeResponse } from "srvx/node";
+import { pathToFileURL } from "node:url";
 import {
   type Connect,
   isRunnableDevEnvironment,
@@ -18,7 +19,7 @@ export function devServer(): Array<PluginOption> {
             const webReq = new NodeRequest({ req, res });
             const def: {
               default: { fetch: (req: Request) => Promise<Response> };
-            } = await import(process.cwd() + "/dist/server/entry-server.js");
+            } = await import(pathToFileURL(process.cwd() + "/dist/server/entry-server.js").href);
             sendNodeResponse(res, await def.default.fetch(webReq));
           });
         };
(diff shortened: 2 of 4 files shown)


brenelz:
This has 3 fixes all mixed together so I am going to close

PR #2167, opened 2026-06-26
feat: support `server-only` and `client-only` modules (#2162)

### LINKED ISSUE #2162: [Feature?]: Built-in server-only & client-only virtual module

YanAnghelp:
Closes #2162


packages/start/src/config/index.ts
```diff
@@ -200,6 +200,29 @@ export function solidStart(options?: SolidStartOptions): Array<PluginOption> {
       },
       filter: options?.serverFunctions?.filter,
     }),
+    {
+      name: "solid-start:boundary-modules",
+      enforce: "pre",
+      resolveId(id, importer, { ssr }) {
```
brenelz, line 215: this is all that is needed?

YanAnghelp: Yes, I have already tested it.

YanAnghelp: There's a known limitation: when a client component imports a file that uses `server-only`, even if all functions in that file are marked with `"use server"`, the build will incorrectly treat this as the client component directly calling `server-only` functions and throw an error. The only workaround is to apply `"use server"` at the file level.

However, `"use server"` already guarantees that the functions execute on the server — in that case, `server-only` is redundant. `server-only` is primarily meant to protect functions that must run on the server and cannot be called directly from outside (i.e., not exposed as RPC endpoints).

brenelz (approved):

packages/start/src/config/index.ts
```diff
@@ -209,6 +209,29 @@ export function solidStart(options?: SolidStartOptions): Array<PluginOption> {
...
+              `Attempt to import 'client-only' in a server module: ${importer}`,
+            );
+        } else {
+          return null;
+        }
+        return "\0solid-start:boundary-modules:id";
+      },
+      load(id) {
+        if (id === "\0solid-start:boundary-modules:id") return "export {}";
+      },
```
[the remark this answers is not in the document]
YanAnghelp: Because the returned value is just the fixed "export {}", this will not cause any actual bugs. This can be regarded as a style suggestion.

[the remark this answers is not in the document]
YanAnghelp: Starting from Vite 6, [Each environment has an isolated module graph.](https://vite.dev/guide/api-environment-instances#separate-module-graphs) However, here it is specifically bound to the environment, so server-only and client-only actually point to different modules.

YanAnghelp:
To facilitate #2207, this plugin should be moved to a separate file. Should the changes be made directly in this PR or in #2207?

lxsmnsyc:
@brenelz I think instead of a separate plugin since this one is a superset of the server import guard, perhaps we can mark Start's server-only modules via import rather than the plugin in #2207 ?

lxsmnsyc (approved):

YanAnghelp:
@lxsmnsyc In that case, the `importer` would be the internal module itself, making it difficult to pinpoint the error.

lxsmnsyc:
@YanAnghelp I'm not sure I understood. My point was that the #2207 change has a "server-only" module list, and I feel like instead we can utilize your plugin. Does it work or does the two plugin have to co-exist

YanAnghelp:
@lxsmnsyc I think the same way. We only need to check the module list at the beginning of this plugin. We don't need two plugins. That's why I proposed that this plugin should be moved to a separate file to keep the code concise.

PR #2174, opened 2026-07-04
fix: debounce fs route hmr events, avoid reload for existing routes

katywings:

- When starting the server, multiple reloads and invalidations are triggered per added route
- When renaming routes, multiple reloads and invalidations are triggered (file rename = "remove file + add file")
- When just changing an existing route, it still triggers a full client reload

https://github.com/user-attachments/assets/41e2475f-0c24-49e3-b581-b25c7e772360

## What is the new behavior?

- Reloads and invalidations are debounced
- Changing an existing valid route triggers a soft HMR, no client reload
- Adding a default export to an existing invalid route file, adds the route and reloads the client
- Removing a default export from an existing valid route file, removes the route and reloads the client 

https://github.com/user-attachments/assets/f2f970f6-f874-42a0-8d2b-3b2ba2b1a6d8

## Other information

From what I can tell the route HMR has been broken atleast since https://github.com/solidjs/solid-start/pull/2125. I tried to make sure that I didn't introduce a regression. The routes graph is still updated when adding new files.

### DIFF
--- a/packages/start/src/config/fs-routes/fs-watcher.ts
+++ b/packages/start/src/config/fs-routes/fs-watcher.ts
@@ -1,22 +1,10 @@
-import type {
-  EnvironmentModuleNode,
-  FSWatcher,
-  PluginOption,
-  ViteDevServer,
-} from "vite";
+import type { EnvironmentModuleNode, FSWatcher, PluginOption, ViteDevServer } from "vite";
+import { debounce } from "../../utils/debounce.ts";
 import { VITE_ENVIRONMENTS } from "../constants.ts";
 import { moduleId } from "./index.ts";
 import type { BaseFileSystemRouter } from "./router.ts";
 
-interface CompiledRouter {
-  removeRoute(path: string): void;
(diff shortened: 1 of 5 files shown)


packages/start/src/config/fs-routes/fs-watcher.ts
```diff
@@ -1,56 +1,52 @@
...
-): () => void {
-  routes.addEventListener("reload", handleRoutesReload);
-  return () => routes.removeEventListener("reload", handleRoutesReload);
+) {
+  const devEnv = server.environments[environment]!;
+  if (!devEnv?.moduleGraph) return;
+
+  /**
+   * Debounce catches multiple route changes in a row
+   * Short timeout for inexpensive invalidations11
```
lxsmnsyc: "11"

katywings: Thanks! Fixed :)

lxsmnsyc (approved):

PR #2187, opened 2026-07-16
fix: require vite 8 newer

birkskyum:
Rolldown is Stable, Vite 8 is Stable.

Being able to assume vite 8+ makes it easier to maintain this library, because we cut out the entire rollup/esbuild dance, and because of it's strong rollup plugin backwards compat there's a clear migration path.

Running vite build will show

```
❯ pnpm run build
$ vite build
[@solidjs/vite-plugin-nitro-2] This plugin is deprecated. Migrate to Nitro v3 using the official `nitro/vite` plugin. See https://github.com/solidjs/templates/tree/main/solid-start-v2 for examples.
vite v8.1.4 building client environment for production...
✓ 14 modules transformed.
computing gzip size...

```


lxsmnsyc:
I think we should lock it Vite 8 first no? are we not expecting some big changes

birkskyum:
> I think we should lock it Vite 8 first no? are we not expecting some big changes

I don't understnd - this does exactly that, lock it to vite 8 and up. Which big changes do you have in mind?

lxsmnsyc:
@birkskyum the version range applies to 9 and above, which is what I'm concerned about.

birkskyum:
Okay, I've set it vite 8 only

birkskyum:
I can follow up with a dedicated vite 9 forward compat pr, so Start v2 has some runway.

lxsmnsyc (approved):

PR #2194, opened 2026-07-16
fix: add files list to pnpm 11 bundle

birkskyum:
This PR ensures `@solidjs/start` includes its compiled output when published with pnpm 11.

The pnpm 11 upgrade changed the package contents because `packages/start` did not define a `files` field. The root `.gitignore` excludes `dist`, causing the compiled output to be omitted even though the published exports point to `./dist`.

This adds an explicit package allowlist matching the contents previously published with beta.0, including both `dist` and `src`.

### DIFF
--- a/.changeset/curvy-carrots-dream.md
+++ b/.changeset/curvy-carrots-dream.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/start": patch
+---
+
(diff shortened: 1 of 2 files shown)


birkskyum:
Use this instead
- #2193

PR #2204, opened 2026-07-17
fix: run onBeforeResponse middleware in declared order

### LINKED ISSUE #2131: [Bug?]: V2 middleware functions do not execute in order
### Duplicates

### Latest version

### Current behavior 😯

onRequest functions execute in order
onBeforeResponse functions execute in inconsistent (backwards?) order

### Expected behavior 🤔

onBeforeResponse functions execute in order

### Steps to reproduce 🕹

Steps:

1. create new V2 project
2. create middleware file with this code
```ts
import { createMiddleware } from '@solidjs/start/middleware'

const one = async () => {
  console.log('begin one')
  await new Promise(resolve => setTimeout(resolve, Math.random() * 1000))
  console.log('end one')
}

const two = async () => {
  console.log('begin two')
  await new Promise(resolve => setTimeout(resolve, Math.random() * 1000))
  console.log('end two')
}

const three = async () => {
  console.log('begin three')
  await new Promise(resolve => setTimeout(resolve, Math.random() * 1000))
  console.log('end three')
}

const four = async () => {
  console.log('begin four')
  await new Promise(resolve => setTimeout(resolve, Math.random() * 1000))
  console.log('end four')
}

const five = async () => {
  console.log('begin five')
  await new Promise(resolve => setTimeout(resolve, Math.random() * 1000))
  console.log('end five')
}

const six = async () => {
  console.log('begin six')
  await new Promise(resolve => setTimeout(resolve, Math.random() * 1000))
  console.log('end six')
}

const seven = async () => {
  console.log('begin seven')
  await new Promise(resolve => setTimeout(resolve, Math.random() * 1000))
  console.log('end seven')
}

const eight = async () => {
  console.log('begin eight')
  await new Promise(resolve => setTimeout(resolve, Math.random() * 1000))
  console.log('end eight')
}

const nine = async () => {
  console.log('begin nine')
  await new Promise(resolve => setTimeout(resolve, Math.random() * 1000))
  console.log('end nine')
}

const ten = async () => {
  console.log('begin ten')
  await new Promise(resolve => setTimeout(resolve, Math.random() * 1000))
  console.log('end ten')
}

export default createMiddleware({
  onRequest: [
    one,
    two,
    three,
    four,
  ],
  onBeforeResponse: [
    five,
    six,
    seven,
    eight,
    nine,
    ten,
  ]
})
```
3. observe how it is different than V1

### Context 🔦

trying to migrate to V2 and realized some middleware functions were breaking because values they expected were supposed to be populated by previous functions, but those functions were not called yet because it is happening out of order

### Your environment 🌎

```shell

```

brenelz:
Fixes #2131

- [x] Addresses an existing open issue: fixes #2131

`onBeforeResponse` functions passed to `createMiddleware` execute in reverse array order. h3 v2 middleware compose onion-style, so logic after `await next()` unwinds innermost-first — and `wrapResponseMiddleware` does its work after awaiting `next()`, as noted by @sabercoy in the issue.

## What is the new behavior?

The response wrappers are registered reversed, so the `onBeforeResponse` callbacks run in the order they were declared, matching v1 semantics. `onRequest` ordering is unchanged.

## Other information

Added `packages/start/src/middleware/index.spec.ts` covering onRequest order, onBeforeResponse order (fails without this fix), request→handler→response sequencing, and response replacement from an onBeforeResponse middleware.


packages/start/src/middleware/index.ts
```diff
@@ -85,7 +85,10 @@ export function createMiddleware(
   if (typeof args.onBeforeResponse === "function") {
     mw.push(wrapResponseMiddleware(args.onBeforeResponse));
   } else if (Array.isArray(args.onBeforeResponse)) {
-    mw.push(...args.onBeforeResponse.map(wrapResponseMiddleware));
+    // h3 middleware compose onion-style: logic after `await next()` unwinds
+    // innermost-first (reverse registration order). Register the wrappers
+    // reversed so onBeforeResponse functions run in their declared order.
+    mw.push(...args.onBeforeResponse.map(wrapResponseMiddleware).reverse());
```
brenelz, line 93: @katywings you probably know the most about middleware. Does this change make sense?

packages/start/src/middleware/index.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+
+import { createMiddleware } from "./index.ts";
+
+const createApp = (middleware: Middleware[]) => {
+  const app = new H3();
+  app.use(defineHandler({ middleware, handler: () => "ok" }));
+  return app;
+};
+
+const delay = () => new Promise(resolve => setTimeout(resolve, 5));
```
katywings: Node has a built-in promisified setTimeout: 

```
import {
  setTimeout,
} from 'node:timers/promises';
```

birkskyum:
Got some feedback from fable as well:

### Suggestions (minor, none blocking)
- Add a chaining test: nothing currently pins that a later-declared onBeforeResponse sees an earlier one's replacement response as body. That is the other observable behavior the reversal changes (before the fix, the chaining direction was also backwards), and a regression here would be silent.
- Cheap coverage gaps if you want them: the single-function onBeforeResponse form and the raw Middleware[] passthrough are untested.
- Release-note consideration: any app that worked around #2131 by pre-reversing its onBeforeResponse array will now get flipped behavior. The changeset describes the fix, which is probably sufficient for a beta line, but it is worth a -sentence in the release notes.
- Out of scope but worth a follow-up look: because response wrappers are registered after request wrappers, an onRequest that returns a Response short-circuits past all onBeforeResponse callbacks entirely. If v1 ran response hooks even for early-returned responses, that is a separate remaining semantic gap; this PR neither causes nor fixes it.

brenelz:
Addressed the feedback

birkskyum (approved):

packages/start/src/middleware/index.ts
katywings: Makes sense yuppa 👍.

What I wonder is why we arent using h3's `onRequest` and `onResponse` helpers 🤔. (onRequest does pretty much nothing, but onResponse would simplify our wrapResponseMiddleware)

Issue #2229, opened 2026-07-22
[Bug]: [v2] HttpStatusCode with deferStream no longer works


katywings:
### Duplicates

- [x] I have searched the existing issues

### Latest version

- [x] I have tested the latest version

### Current behavior 😯

Setting the response status code with `<HttpStatusCode code={404} />` after waiting for some `query` data with `deferStream: true`, stopped working with Start v2. It instead just responds with 200. The same regression probably applies to `<HttpHeader>`.

### Expected behavior 🤔

The response status code should be properly sended, even if `<HttpStatusCode code={404} />` is server-rendered late with `deferStream: true`.

### Steps to reproduce 🕹

1. Set the following as content of `src/routes/[...404].tsx`:

```tsx
import { Title } from "@solidjs/meta";
import { createAsync, query } from "@solidjs/router";
import { HttpStatusCode } from "@solidjs/start";
import { Show } from "solid-js";

const getData = query(async () => "test", "data");

export default function NotFound() {
  const data = createAsync(() => getData(), { deferStream: true });

  return (
    <main>
      {data()}
      <Title>Not Found</Title>
      <Show when={data()}>
        <HttpStatusCode code={404} />
      </Show>
      
      <h1>Page Not Found</h1>
      <p>
        Visit{" "}
        <a href="https://start.solidjs.com" target="_blank">
          start.solidjs.com
        </a>{" "}
        to learn how to build SolidStart apps.
      </p>
    </main>
  );
}
```

2. Open http://localhost:3000/notfound
3. Check the status code

### Your environment 🌎

```shell
- node: v24.18
- Start: ^2.0.0-rc.1
```

PR #2231, opened 2026-07-24
fix: defer handler response until stream shell payload

katywings:

h3 v2 no longer waits sending the response until the first stream payload arrives, and ends up throwing away response status/header modifications.

## What is the new behavior?

The h3 handler response is deferred again until the first `renderToStream` payload with response status/headers arrives. This is accomplished with the `iterable` h3 utility. h3 added back the functionality to wait for the first stream payload in h3js/h3#1512 as part of v2.0.1-rc.26.

h3 v2.0.1-rc.26 also fixed some other oddities with headers, which in return broke our header workarounds - in particular the test `should preserve multiple Set-Cookie headers on redirect (RFC 6265)`. With all of the header fixes introduced in the past couple h3 releases, all our header workarounds seem to be not necessary anymore. Our tests now pass without `produceResponseWithEventHeaders` (which included the workarounds).

This PR also slightly overhauls the `should preserve multiple Set-Cookie headers on redirect (RFC 6265)` test asserts, so that they result in more understandable error messages in case of a fail.

## Other information

Refs: #2229, h3js/h3#1510

### DIFF
--- a/packages/start/src/server/handler.ts
+++ b/packages/start/src/server/handler.ts
@@ -1,6 +1,13 @@
 import middleware from "solid-start:middleware";
-import { defineHandler, getCookie, H3, type H3Event, redirect, setCookie } from "h3/generic";
-import { join } from "pathe";
+import {
+  defineHandler,
+  getCookie,
+  H3,
+  type H3Event,
+  iterable,
+  redirect,
+  setCookie,
+} from "h3/generic";
 import type { JSX } from "solid-js";
 import { sharedConfig } from "solid-js";
 import { getRequestEvent, renderToStream, renderToString } from "solid-js/web";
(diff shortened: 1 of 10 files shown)


packages/start/src/server/handler.ts
```diff
@@ -123,7 +123,13 @@ export function createBaseHandler(
...
       // h3 expects a standard web ReadableStream across runtimes. The adapter
       // also tolerates cancellation while Solid finishes outstanding work.
-      return toWebReadableStream(stream);
+      const [webStream, shellRendered] = toWebReadableStream(stream);
+
+      // Waiting for the first stream payload
+      // This blocks h3 from sending the Response without deferred response status/headers
+      await shellRendered;
+
+      return webStream;
```
lxsmnsyc: why does the previous implementation allow returning the two, what's the diff

lxsmnsyc: nvm didn't saw the next file 😴

birkskyum (approved):

brenelz (approved):

PR #2241, opened 2026-07-26
fix: don't serialize error stack traces to the client in production

### LINKED ISSUE #1967: [Bug?]: [security] Server stack trace sent to client on error in server action

birkskyum:
- Fixes #1967

When a server function throws, the error is serialized with seroval and rethrown on the client. Seroval includes `Error.prototype.stack` unless the `ErrorPrototypeStack` feature is disabled, and `serialization.ts` only disabled `Feature.RegExp`, so production builds ship server file paths and internal function names to anyone who can trigger a throw (CWE-209).

Disables the feature on the two serializer paths when `import.meta.env.PROD`. Development is unchanged, so the stack is still there when it's useful. The parse side keeps the original feature set, so a payload that does carry a stack still deserializes.

Note `serializeToJSStream` passed no `disabledFeatures` at all, so it leaked regardless of the JSON path's settings. It gets its own constant rather than reusing the JSON one, to avoid picking up the `Feature.RegExp` downgrade it never had.

### DIFF
--- a/.changeset/wild-pots-repeat.md
+++ b/.changeset/wild-pots-repeat.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/start": patch
+---
+
+Don't send server error stack traces to the client in production builds. When a server function throws, the error is serialized and rethrown on the client, and seroval included `Error.prototype.stack` by default, leaking server file paths and internal function names. Stacks are still serialized in development.
--- a/packages/start/src/fns/serialization.spec.ts
+++ b/packages/start/src/fns/serialization.spec.ts
@@ -1,0 +1,78 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+async function loadSerialization(prod: boolean) {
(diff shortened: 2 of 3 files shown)


brenelz (approved):
LGTM but might want Alexis to make sure

lxsmnsyc (approved):

Issue #1918, opened 2025-07-10
[Bug?]: unplugin-macros doesn't work


huseeiin:
### Duplicates

- [x] I have searched the existing issues

### Latest version

- [x] I have tested the latest version

### Current behavior 😯

`routes/test.ts:`
```ts
import { buildTime, getRandom } from "../lib/test.ts" with { type: "macro" };

export function GET() {
  return getRandom();
}

console.log(buildTime);
```
when you go to `/test` you'll see a different value every time.

### Expected behavior 🤔

its supposed to be the same value every time, since its a macro.

### Steps to reproduce 🕹

i couldn't reproduce in vinxi. this bug is only in solidstart.

### Your environment 🌎

```shell

```

huseeiin:
this is because of how api routes are imported under the hood by solid start, with a `?pick=GET` (or `?pick=POST`) query.

brenelz:
Do you have suggestions how to handle this? I assume its uncompatible right now

huseeiin:
> Do you have suggestions how to handle this? I assume its uncompatible right now

we need a way to import routes without a query. sveltekit does this. i don't know how tho

PR #2242, opened 2026-07-26
fix: end route module ids in the source extension so unplugin filters match

### LINKED ISSUE #1374: [Bug?]: unplugin-icons + unplugin-auto-import don't work within filesystem routes
### Duplicates

### Latest version

### Current behavior 😯

Inside files in `/routes` that have a default export - ie those that export a route component - the imports that would usually be provided by `unplugin-auto-import` are (i assume) not present, so instead of rendering the icons there is an attempt to render `undefined`.
In files that don't export a default component this is apparently not an issue.

### Expected behavior 🤔

The imports from `unplugin-auto-import` should always be present so that auto-imports can be used in route files.

### Steps to reproduce 🕹

Steps:

1. Go to [this StackBlitz](https://stackblitz.com/edit/github-htmjmk-dt5yga?file=src%2Fapp.tsx,src%2Froutes%2Findex.tsx)
2. Notice that using the icon works in `/app.tsx` and `/routes/SeparateFileComponent.tsx`, but not `/routes/index.tsx`

### Context 🔦

I want to have auto imported icons throughout my whole app.

### Your environment 🌎

_No response_

birkskyum:
- Closes #1374

Route files are requested with a `?pick=...` query, leaving ids like `src/routes/index.tsx?pick=default&pick=$css`. That does not end in `.tsx`, so plugins filtering with an end-anchored extension regex (`/\.[jt]sx?$/`, the default in `unplugin-auto-import` and friends) silently skip every route file.

The query now ends with `&lang.<ext>`, the same convention `@vitejs/plugin-vue` uses for `?vue&type=script&lang.ts`. The three places that built this string share a `toRouteModuleId()` helper.

That change moves the query into Rollup's chunk base name, so `sanitizeRouteChunkName` strips it back off via `output.sanitizeFileName`. Output filenames are unchanged from before this PR.

Verified with stock `AutoImport({ dirs: [...] })` and no `include` override: the import is now injected into route chunks, in both build and dev. Existing suites pass (64 unit, 30 e2e, 30 bundled-dev), all fixtures build.


birkskyum:
Superseded by #2249, which landed the same fix on `main`: the same `lang.<ext>` marker on route module ids, plus the same chunk-name sanitizer to keep filenames unchanged. Its `toPickId` also guards on the script extensions, so `.md`/`.mdx` routes keep the old id shape, which is better than what this PR did.

That leaves nothing here but a changeset, which is now #2254.

#1374 (the `unplugin-auto-import` report this PR was aimed at) is fixed by #2249 for the same reason #1918 was.

PR #2250, opened 2026-07-26
feat: custom seroval plugins via `serialization.plugins`

### LINKED ISSUE #1474: [Feature?]: Custom Seroval plugins
### Duplicates

### Latest version

### Summary 💡

It would be useful to be able to define a list of your own plugins used by Seroval, as far as I know, this is the only way to properly parse custom classes.

### Examples 🌈

_No response_

### Motivation 🔦

I'm struggling with trying to return the ObjectId (used in MongoDb) from a server function, instance cannot be parsed by Seroval.

birkskyum:
- Closes #1474

Values seroval has no built-in support for (Mongo `ObjectId`, Prisma `Decimal`, `Temporal`, custom classes) threw `SerovalUnsupportedTypeError` when crossing a server function boundary, with no way to register a plugin. Seroval has supported `createPlugin` since v1; SolidStart just hardcoded its plugin list.

```ts
// vite.config.ts
solidStart({
  serialization: { plugins: "src/seroval-plugins.ts" },
});

// src/seroval-plugins.ts
import { createPlugin } from "@solidjs/start/serialization";
export default [MyPlugin];
```

The path resolves through a new `solid-start:seroval-plugins` virtual module, wired like `solid-start:middleware`. One module feeds both `fns/client.ts` and `fns/handler.ts`, so client and server can't drift out of sync. Built-in plugins keep precedence (seroval takes the first passing `test()`, user plugins are appended).

Also adds a `@solidjs/start/serialization` entrypoint re-exporting `createPlugin`, `OpaqueReference`, and plugin types. Without it users need their own `seroval` dep, and a version mismatch wouldn't fail the build, it would emit plugin nodes the other end can't interpret.

Two limits, both documented on the option:
- Only server function and action payloads. The SSR hydration payload is serialized by `solid-js/web` and is unaffected.
- Under `mode: "js"` the payload is eval'd on the client, so a plugin's `serialize` may reference globals only, not the plugin module's imports. `deserialize` (`mode: "json"`, the default) has no such limit.

Verified: 6 new unit tests, 1 new e2e round-tripping a custom class in both directions, and a manual browser check under both serialization modes. Full suites green (67 unit, 31 e2e, 31 bundled-dev).

### DIFF
--- a/.changeset/lemon-plugins-serialize.md
+++ b/.changeset/lemon-plugins-serialize.md
@@ -1,0 +1,25 @@
+---
+"@solidjs/start": minor
+---
+
+Add `serialization.plugins` to configure custom Seroval plugins for server functions.
+
+Values Seroval has no built-in support for (Mongo's `ObjectId`, Prisma's `Decimal`, `Temporal`, and other custom classes) previously threw when returned from or passed to a server function. Point the new option at a module whose default export is an array of plugins:
+
+```ts
+// vite.config.ts
+solidStart({
+  serialization: {
+    plugins: "src/seroval-plugins.ts",
+  },
+});
+```
+
+```ts
+// src/seroval-plugins.ts
+import { createPlugin } from "@solidjs/start/serialization";
+```
+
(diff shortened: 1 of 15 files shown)


lxsmnsyc (approved):
lgtm

PR #2251, opened 2026-07-26
feat: include server function id in the request URL

### LINKED ISSUE #1795: [Feature?]: show function name in _server routes

birkskyum:
Closes #1795

Server function calls all went to a single `_server` URL, so access logs, traces and the network panel collapsed every function into one entry.

Requests now go to `_server/<id>`. In development the id ends with the source name, so you get `POST /_server/13375fbc-0-ping` with no configuration.

Production ids stay opaque by default. Opt in to keep the names:

```js
solidStart({ serverFunctions: { readableIds: true } })
```

Notes:
- `action.url` changes from `/_server?id=<id>` to `/_server/<id>`. The handler still accepts the `?id=` form, so hand-written URLs keep working.
- The route check was `pathname.startsWith("/_server")`, which also matched unrelated routes like `/_serverless`. It is now an exact match on `/_server` or a `/_server/` prefix.

Tested: 12 new unit tests for the URL helpers, both e2e suites pass (30/30 dev, 30/30 bundled-dev), and verified against a production build that the new URL, the legacy `?id=` URL, and `readableIds` all behave as expected.

### DIFF
--- a/.changeset/khaki-forks-shave.md
+++ b/.changeset/khaki-forks-shave.md
@@ -1,0 +1,25 @@
+---
+"@solidjs/start": minor
+---
+
+Include the server function id in its request URL
+
+Server function calls now go to `_server/<id>` instead of a bare `_server`, so
+access logs, traces and the network panel can tell functions apart instead of
+collapsing every call into one entry. Ids contain the function's source name in
+development, e.g. `POST /_server/a1b2c3-0-getUser`.
+
+Production ids stay opaque by default. Set `serverFunctions.readableIds` to keep
+the names in production builds too:
+
+```js
+import { solidStart } from "@solidjs/start/config";
+import { defineConfig } from "vite";
+
+export default defineConfig({
+  plugins: [solidStart({ serverFunctions: { readableIds: true } })],
+});
+```
+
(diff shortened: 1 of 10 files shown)


lxsmnsyc:
Hmmm this needs discussion. Pre-1.0 we had this kind of format then shifted to what we have today. I'm honestly fine with this change, but perhaps @ryansolid can weigh in on this

birkskyum:
This issue has already waited for an opinion from Ryan for [over a year](https://github.com/solidjs/solid-start/issues/1795#issuecomment-2657298579)... so it's not something we can hold back a release of Start 2 for. Closing for now.

PR #2255, opened 2026-07-27
fix: reload the browser when the server entry or middleware changes in dev

### LINKED ISSUE #1611: [Feature]: HMR for `entry-server.tsx`
### Duplicates

### Latest version

### Summary 💡

As with other files such as front-end files, it would be nice for the app to refresh when we save `entry-server.tsx` or `entry-server.jsx`. For now we have to manually refresh.

### Examples 🌈

_No response_

### Motivation 🔦

Convenience

birkskyum:
- Closes #1611

### Problem

`entry-server.tsx` and the configured middleware are server-only, so they have no module in the client graph and Vite has nothing to send the browser when they change. The dev server re-imports the server entry per request, so the edit is live on the next request, but the open page keeps showing the shell rendered from the previous version until you refresh by hand. Reproduced on the current RC: editing `src/entry-server.tsx` in `apps/fixtures/bare` produces zero messages on the client HMR socket.

### Fix

Adds a `serverOnlyReload()` plugin to `devServer()`. On `hotUpdate` for the `ssr` environment, if the changed file is the resolved server entry or the configured middleware, it sends a `full-reload` on the client hot channel (debounced so one save touching both only reloads once). Both paths are resolved through the ssr environment's plugin container, so an extensionless or aliased `middleware` path still matches.

Only those two entries are handled: modules that also exist on the client (routes, components, server functions) are left to Vite, since reloading on top of a client HMR update would needlessly throw away the page's state. This is also why the check is not the broader "file has no module in the client graph" rule, which would reload and reset the current page whenever you edit a lazy route you have not navigated to yet.

### Verification

Browser check on `apps/fixtures/bare`: before the change the page never updates and needs a manual refresh; after it, the edit appears on its own and the hydrated counter resets, confirming a real reload. An HMR socket probe shows exactly one `full-reload` for `entry-server.tsx` and for `middleware.ts` (in `apps/fixtures/experiments`), while editing `src/app.tsx` stays silent as before. New unit tests cover both entries plus the cases that must stay silent. `tsc --noEmit`, `vitest run` (77 passed), `oxfmt --check`, and the `apps/tests` Playwright suite (35 passed) are green.

### DIFF
--- a/packages/start/src/config/dev-server.ts
+++ b/packages/start/src/config/dev-server.ts
@@ -1,17 +1,22 @@
 import { existsSync } from "node:fs";
-import { join } from "node:path";
+import { join, resolve } from "node:path";
 import { pathToFileURL } from "node:url";
 import { NodeRequest, sendNodeResponse } from "srvx/node";
 import {
   type Connect,
+  type DevEnvironment,
   isRunnableDevEnvironment,
+  normalizePath,
   type PluginOption,
   type ViteDevServer,
 } from "vite";
+import { debounce } from "../utils/debounce.ts";
 import { VITE_ENVIRONMENTS } from "./constants.ts";
+import { parseIdQuery } from "./utils.ts";
 
-export function devServer(serverEntryPath: string): Array<PluginOption> {
+export function devServer(serverEntryPath: string, middlewarePath?: string): Array<PluginOption> {
   return [
+    serverOnlyReload(serverEntryPath, middlewarePath),
     {
       name: "solid-start-dev-server",
       configurePreviewServer(server) {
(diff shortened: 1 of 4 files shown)


packages/start/src/config/dev-server.ts
```diff
@@ -1,17 +1,22 @@
...
   type ViteDevServer,
 } from "vite";
+import { debounce } from "../utils/debounce.ts";
 import { VITE_ENVIRONMENTS } from "./constants.ts";
+import { parseIdQuery } from "./utils.ts";
 
-export function devServer(serverEntryPath: string): Array<PluginOption> {
+export function devServer(serverEntryPath: string, middlewarePath?: string): Array<PluginOption> {
   return [
+    serverOnlyReload(serverEntryPath, middlewarePath),
```
brenelz, line 19: probably good if someone else reviews this but overall looks good

lxsmnsyc:
Doesn't the refresh reload pragma work

PR #2259, opened 2026-07-27
fix: surface errors from lazily loaded route modules instead of hanging the SSR stream

birkskyum:
A syntax error in a route (or anything it imports) left the browser on a blank page with no error shown anywhere.

Every file route loads through `lazy()`. When the module fails to transform, its dynamic import rejects, and Solid's server-side `lazy` has no rejection path:

```js
p = fn();
p.then(mod => p.resolved = mod.default);   // no .catch
```

So `_loading` stays true, the Suspense boundary never resolves, and `renderToStream` never closes. The document shell is flushed and the connection then hangs open forever, which is also why no Vite overlay appears: as far as the browser is concerned the page loaded fine.

`withAssets` already wraps the import on the server for asset collection, so this catches the rejection there and returns a module that throws during render. Solid routes that through `catchError` to the nearest `ErrorBoundary`, i.e. the dev overlay. Also covers a lazy chunk failing to load in production, which hung the stream the same way.

Errors in eagerly imported files (`app.tsx`, `entry-server.tsx`) were already handled and are unchanged.

Verified on `apps/fixtures/basic` with a broken component:

| | before | after |
|---|---|---|
| load page while broken | request never completes, blank page | 500, dev overlay shows the SyntaxError |

Added `lazy.spec.ts`, which times out with "the SSR stream never closed" without the fix.


lxsmnsyc:
perhaps we can move this as an upstream issue?

Issue #1434, opened 2024-04-06
[Feature?]: Error Handling with solid-router


alvechy:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Allow for semantic error handling with server actions.
This is a continuation of https://github.com/solidjs/solid-router/discussions/365 discussion, but focused on `solid-start` server actions.

The goal is to allow:
- return non-`200` (`401`, `500`, etc) responses from server actions for observability platforms to monitor how many responses are not `200`.
- handle such responses in code with reactive `useSubmission().error` when using without the form.

Proposal: use [Response.ok](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok) to populate `useSubmission().error`.

```tsx
const serverAction = action(() => {
  'use server';
  try {
    return perform();
  } catch(error) {
    // Error is serialised, no queries are revalidated (unless specified otherwise)
    // Response.ok is `false`, so `submission.error` is serialised Error.
    // This will make the action use to fail (as the error is rethrown), so needs try/catch
    return json(new Error('Something went wrong'), { code: 500 });
  }
});

const Component = () => {
  const performAction = useAction(serverAction);
  const submission = useSubmission(serverAction);

  // See example, QR Code scanning
  const userActionWithoutForm = () => {
    try {
      await performAction(); // same as fetch().catch() when the request has failed
    } catch {}
  }

  return (
    <Switch>
      <Match when={submission.pending}>Checking</Match>
      <Match when={submission.error}>Retry with another input</Match>
      <Match when={true}>Perform action</Match>
    </Switch>
  )
}
```

If the action returns non-`Error` data, the action does not fail:
```
// action, does not throw an Error
  return json('Something went wrong', { code: 500 });
// Component
  // similar to fetch(response => response.json()) without checking for response.ok
  await performAction();
  // rely on submission.error being populated with 'Something went wrong' string
```

### Examples 🌈

In the app, users can scan a QR code, when valid code is scanned – `action` is executed. If scanned code is invalid – error message is shown with Retry logic:
```tsx
const checkCode = action(async code => {
  "use server";
  try {
    return validity(code);
  } catch(error) {
    // ===Unclear===
  }
});
// component:
const submission = useSubmission(checkCode);
const checkCode = useAction(checkCode);

const onScanned = code => {
  try {
    await checkCode(code);
  } catch(error) {
    // do something, or just ignore it, same way we would do with using fetch directly.
  }
[...]

PR #2261, opened 2026-07-28
fix: reject server function calls on 5xx responses without X-Error

adipascu:
## PR Checklist

- [x] Addresses an existing open issue: relates to #1434 (covers its 5xx case)

Server function calls only reject when the response carries an `X-Error` header. Two cases slip past that check in `packages/start/src/fns/client.ts`:

- A 5xx produced below the handler (a server function module crashing at import time, a proxy or CDN error, a dead worker) has no `X-Error`, so the promise **resolves** and callers mistake server failure for success.
- An error response whose body yields nothing throws `undefined`, so callers have no error to inspect or match on.

Minimal reproduction: https://github.com/adipascu/solid-start-server-fn-500-repro

## What is the new behavior?

Responses with status >= 500 throw the deserialized result the same way `X-Error` does, falling back to an `Error` carrying the status when the body yields nothing. 4xx responses are untouched, since server functions can legitimately return them without `X-Error`.

## Other information

This is the follow-up to #2159, which was merged onto `1.x` where I asked whether to retarget it at `main`. v2's rewritten client has the same logic, so the same fix applies.

Added `packages/start/src/fns/client.spec.ts` beside the existing `handler.spec.ts`. Both regression tests were confirmed to fail without the source change.

### DIFF
--- a/packages/start/src/fns/client.ts
+++ b/packages/start/src/fns/client.ts
@@ -94,8 +94,8 @@ async function fetchServerFunction(
   }
 
   const result = await extractBody(instance, true, response.clone());
-  if (response.headers.has("X-Error")) {
-    throw result;
+  if (response.headers.has("X-Error") || response.status >= 500) {
+    throw result ?? new Error(`Server function call failed with status ${response.status}`);
   }
   return result;
 }
--- a/.changeset/tidy-donkeys-reject.md
+++ b/.changeset/tidy-donkeys-reject.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/start": patch
+---
+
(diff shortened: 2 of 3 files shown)


packages/start/src/fns/client.ts
```diff
@@ -94,8 +94,8 @@ async function fetchServerFunction(
   }
 
   const result = await extractBody(instance, true, response.clone());
-  if (response.headers.has("X-Error")) {
-    throw result;
+  if (response.headers.has("X-Error") || response.status >= 500) {
+    throw result ?? new Error(`Server function call failed with status ${response.status}`);
```
birkskyum, line 98: Does it need the `result ??` part at all?

birkskyum (approved):
LGTM

PR #2262, opened 2026-07-28
feat: add a server function error handler

adipascu:
## PR Checklist

`handleServerFunction` catches everything a server function throws and writes it straight into the response. Nothing else observes it, so the crash is invisible to monitoring, and whatever was thrown reaches the client as-is.

## What is the new behavior?

A new `serverFunctions.onError` option names a module whose default export is called with the thrown value before it is serialized. Returning a value replaces what is sent, returning `undefined` changes nothing.

```ts
solidStart({ serverFunctions: { onError: "src/server-fn-error.ts" } });
```

Naming the module in the config rather than registering a handler at runtime keeps the app in sole control of it, so no dependency can reach the running server and take over reporting. It follows `serialization.plugins` from #1474, with its own virtual module because this one is bundled into the server only and may import a monitoring SDK.

How other frameworks do this:

- SvelteKit `handleError`, which observes and replaces: [docs](https://svelte.dev/docs/kit/hooks#handleError) ([src](https://github.com/sveltejs/kit/blob/b27c82f9c6896d7127a4023a8b1868c68e4a27b9/packages/kit/src/runtime/server/utils.js#L117)). Same `handler(...) ?? fallback` shape as this PR.
- Next.js `onRequestError`, observe only: [docs](https://nextjs.org/docs/app/api-reference/file-conventions/instrumentation#onrequesterror-optional) ([src](https://github.com/vercel/next.js/blob/ad618bf13fbc4be57d6b6136a20547af50708eb2/packages/next/src/server/lib/router-utils/instrumentation-globals.external.ts#L74))
- React Router `handleError`, observe only: [docs](https://reactrouter.com/api/framework-conventions/entry.server.tsx#handleerror) ([src](https://github.com/remix-run/react-router/blob/8bc59bc7b027d5f2aba15a6dd72f44e757dbd8f9/packages/react-router/lib/server-runtime/server.ts#L74))
- Nitro `error` hook, observe only: [docs](https://nitro.build/docs/lifecycle#error-handling) ([src](https://github.com/nitrojs/nitro/blob/77b77ffb19fd522400b517f6773b78832ff436b7/src/build/virtual/app.ts#L42))
- TanStack Start has no dedicated hook. Global `functionMiddleware` wraps every server function, so an app catches around `next()` itself: [docs](https://tanstack.com/start/latest/docs/framework/react/guide/middleware) ([src](https://github.com/TanStack/router/blob/179d9b9b5eec50dc7833b4e2ef91b735d34bbc07/packages/start-client-core/src/createServerFn.ts#L233))

Replacing matters for adapters whose runtime wraps failures, such as Effect's `FiberFailure`. Unwrapping here keeps a `Response` thrown through the wrapper recognised as control flow by the `x instanceof Response` branch.

### DIFF
--- a/packages/start/src/fns/error-handler.ts
+++ b/packages/start/src/fns/error-handler.ts
@@ -1,31 +1,8 @@
+import onServerFunctionError from "solid-start:server-fn-error-handler";
+
 export type ServerFunctionErrorHandler = (thrown: unknown) => unknown;
-
-let errorHandler: ServerFunctionErrorHandler | undefined;
-
-/**
- * Registers a handler for errors thrown by server functions, called before the
- * error is serialized into the response.
- *
- * Return a value to send it in place of what was thrown, or `undefined` to
- * send the original. Pass `undefined` to unregister.
- *
- * @example
- * ```ts
- * import { setServerFunctionErrorHandler } from "@solidjs/start/fns/server";
- *
- * setServerFunctionErrorHandler(error => {
- *   reportToMonitoring(error);
- *   return new Error("Internal server error");
- * });
- * ```
- */
-export function setServerFunctionErrorHandler(
-  handler: ServerFunctionErrorHandler | undefined,
-): void {
-  errorHandler = handler;
-}
 
 /** @internal */
 export function applyServerFunctionErrorHandler(thrown: unknown): unknown {
-  return errorHandler?.(thrown) ?? thrown;
+  return onServerFunctionError?.(thrown) ?? thrown;
 }
--- a/packages/start/src/fns/error-handler.ts
+++ b/packages/start/src/fns/error-handler.ts
@@ -1,0 +1,31 @@
+export type ServerFunctionErrorHandler = (thrown: unknown) => unknown;
+
+let errorHandler: ServerFunctionErrorHandler | undefined;
+
+/**
+ * Registers a handler for errors thrown by server functions, called before the
+ * error is serialized into the response.
+ *
+ * Return a value to send it in place of what was thrown, or `undefined` to
+ * send the original. Pass `undefined` to unregister.
+ *
+ * @example
+ * ```ts
+ * import { setServerFunctionErrorHandler } from "@solidjs/start/fns/server";
+ *
+ * setServerFunctionErrorHandler(error => {
+ *   reportToMonitoring(error);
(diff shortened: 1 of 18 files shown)


packages/start/src/fns/handler.ts
```diff
@@ -126,6 +126,10 @@ export async function handleServerFunction(h3Event: H3Event) {
     }
     return serializeToJSONStream(result);
   } catch (x) {
+    // Nothing else observes a server function failure: it is caught here and
+    // written straight to the response, so monitoring never sees the crash and
+    // the thrown value reaches the client as-is.
+    x = globalThis.__transformServerFnError?.(x) ?? x;
```
lxsmnsyc: I don't agree with a globalThis implementation, that can be tampered with by another code (hoping its not malicious actors)

adipascu: Agreed, dropped it. The handler is now registered through `setServerFunctionErrorHandler`, exported from `@solidjs/start/fns/server`, so nothing reads or writes `globalThis`.

packages/start/src/fns/error-handler.ts
```diff
@@ -0,0 +1,31 @@
...
+ * setServerFunctionErrorHandler(error => {
+ *   reportToMonitoring(error);
+ *   return new Error("Internal server error");
+ * });
+ * ```
+ */
+export function setServerFunctionErrorHandler(
+  handler: ServerFunctionErrorHandler | undefined,
+): void {
+  errorHandler = handler;
```
lxsmnsyc: Functionally, I think it's still the same as the last one since another package can just import this function and call it through peer dependency.

I think the go-to solution here is somewhere similar to #1474 , I think we can change that where instead of just exporting the plugins, we can also export something like an `onServerFunctionError` that the runtime can call.

Example:

```js
// serialization.ts
export function onServerFunctionError(error: unknown) {
  return doStuff(error);
}
```

```js
import { onServerFunctionError } from 'solid-start:serialization';

// ...
x = onServerFunctionError(err);
```

lxsmnsyc (approved):
AI works wonders

adipascu:
Thanks for the #1474 pointer, `serverFunctions.onError` came out much cleaner than the setter. Branch is green and up to date, and good to merge from my side whenever someone with the rights can.

PR #2266, opened 2026-07-29
docs: clean up the `serverFunctions.onError` JSDoc

adipascu:
## PR Checklist

Cleanup of my own JSDoc from #2262. `serverFunctions.onError` is documented too thinly to write a handler from, and `ServerFunctionErrorHandler` has no doc at all.

## What is the new behavior?

Behavior unchanged, documentation only. The option now states that:

- a thrown `Response` reaches the handler, so replacing everything breaks `redirect()`
- the return value is not awaited
- only calls arriving over `/_server` reach it

`ServerFunctionErrorHandler` gets `@param`, `@returns`, and an `@example`, and the inert `/** @internal */` is dropped.

## Other information

Comments only, so no tests and no changeset.

Paired with solidjs/solid-docs#1610, which expands the config reference page for this option.


birkskyum (approved):

packages/start/src/fns/error-handler.ts
```diff
@@ -1,8 +1,38 @@
...
+ *   Sentry.captureException(thrown); // or console.error, or any reporter
+ *   return new Error("Something went wrong");
+ * };
+ *
+ * export default onServerFunctionError;
+ * ```
+ */
 export type ServerFunctionErrorHandler = (thrown: unknown) => unknown;
 
-/** @internal */
```
birkskyum, line 5: Is this removal of "internal" from applyServerFunctionErrorHandler intentinoal?

adipascu: Yes, I added that by accident in #2262. It's the only `@internal` in the repo, and a no-op without `stripInternal`.

PR #2275, opened 2026-08-01
fix: detect shared chunks via moduleIds and imported css instead of isDynamicEntry

katywings:

Shared chunk css hasn't been server rendered in production as of late, resulting in flashes of unstyled content. Probably caused by the upgrade to vite 8, shared chunks no longer are being flagged as dynamic entry, causing them to not be picked up by the shared chunk id remapping logic.

Screenshot_20260801_215023

## What is the new behavior?

Shared chunk css is being server rendered again in production. The remapping logic has being updated, to detect shared chunks based on quantity of included modules + imported css, instead of `isDynamicEntry`.

Screenshot_20260801_215156

### DIFF
--- a/packages/start/src/config/lazy.ts
+++ b/packages/start/src/config/lazy.ts
@@ -73,12 +73,16 @@ const lazy = (): PluginOption => {
       if (this.environment.name !== VITE_ENVIRONMENTS.client) return;
 
       for (const chunk of Object.values(bundle)) {
-        if (chunk.type !== "chunk" || !chunk.isDynamicEntry || chunk.facadeModuleId) continue;
+        if (chunk.type !== "chunk" || chunk.facadeModuleId) continue;
+        if (!chunk.viteMetadata?.importedCss.size) continue;
+
+        const moduleIds = chunk.moduleIds.filter(id => !id.endsWith("css"));
+        if (moduleIds.length <= 1) continue;
 
         // Has to follow Vites implementation:
         // https://github.com/vitejs/vite/blob/4be37a8389c67873880f826b01fe40137e1c29a7/packages/vite/src/node/plugins/manifest.ts#L179
         const chunkName = `_${basename(chunk.fileName)}`;
-        for (const id of chunk.moduleIds) {
+        for (const id of moduleIds) {
           sharedChunkNames[id] = chunkName;
         }
       }
--- a/.changeset/deep-sails-peel.md
+++ b/.changeset/deep-sails-peel.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/start": patch
+---
+
+Fixed shared chunk css not being server rendered in production (Vite 8 regression).
--- a/apps/fixtures/css/vite.config.ts
+++ b/apps/fixtures/css/vite.config.ts
@@ -13,13 +13,14 @@ export default defineConfig({
          * Creates a shared chunk with two components. Needed for the "SharedChunk" test!
          * The vite manifest behaves differently for such shared chunks.
          * More info: packages/start/src/config/lazy.ts
-         *
-         * TODO: When switching to Rolldown, migrate this to advancedChunks
-         * https://vite.dev/guide/rolldown.html#manualchunks-to-advancedchunks
          */
-        manualChunks(id) {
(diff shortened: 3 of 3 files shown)


katywings:
P.S. @ryansolid, if this is being accepted, the same changes probably have to be applied in the Solid v2 manifest logic

packages/start/src/config/lazy.ts
```diff
@@ -73,12 +73,16 @@ const lazy = (): PluginOption => {
       if (this.environment.name !== VITE_ENVIRONMENTS.client) return;
 
       for (const chunk of Object.values(bundle)) {
-        if (chunk.type !== "chunk" || !chunk.isDynamicEntry || chunk.facadeModuleId) continue;
+        if (chunk.type !== "chunk" || chunk.facadeModuleId) continue;
```
birkskyum, line 76: @katywings I looked over this PR and didn't catch anything - LGTM. For thoroughness i figured, since we don't have coderabbit or greptile in this repo, i better run it by an LLM as well, to see if it finds somehting - and it output all this. 

<------ AI output below --------> 

From line 75 in lazy.ts do the following:

```ts
      for (const chunk of Object.values(bundle)) {
        if (chunk.type !== "chunk" || chunk.facadeModuleId) continue;

        // Has to follow Vites implementation:
        // https://github.com/vitejs/vite/blob/4be37a8389c67873880f826b01fe40137e1c29a7/packages/vite/src/node/plugins/manifest.ts#L179
        const chunkName = `_${basename(chunk.fileName)}`;
        for (const id of chunk.moduleIds) {
          if (!fileEndingRegex.test(id)) continue;
          sharedChunkNames[id] = chunkName;
        }
      }
```

**Explanation**

The loop does not decide how Vite names manifest entries. Vite has already made that decision. The loop’s job is to teach SolidStart which Vite manifest key belongs to each source module.

Vite uses two kinds of manifest keys:

- If a chunk has a `facadeModuleId`, Vite keys it by the source module path, such as `/src/components/Lazy.tsx`.
- If a chunk has no `facadeModuleId`, Vite keys it by its generated filename, such as `_shared-C30ClGHB.js`.

SolidStart must copy that relationship into `sharedChunkNames`. Later, while building the server bundle, it injects the corresponding key as `id$$`:

```ts
export const id$$ = "_shared-C30ClGHB.js";
```

During SSR, SolidStart uses `id$$` to look up the client manifest and find the component’s CSS.

The version of this PR identifies facade-less chunks using two additional signals:

```ts
chunk.viteMetadata.importedCss.size > 0
moduleIds.length > 1
```

Those signals describe the specific shared chunk she observed, but they do not control how Vite keys the chunk. This can miss:

- A facade-less chunk containing only one application module.
- A facade-less chunk whose CSS belongs to one of its imported chunks rather than directly to itself.

In both cases, Vite still uses `_generated-filename.js` as the manifest key, so SolidStart still needs the mapping.

The suggested loop instead follows Vite’s decision directly:

```ts
if (chunk.type !== "chunk" || chunk.facadeModuleId) continue;
```

This means: “Vite will use a generated manifest key for this chunk, so record that key for its relevant modules.”

The inner filter:

```ts
if (!fileEndingRegex.test(id)) continue;
```

only limits the map to JSX/TSX modules that SolidStart’s later server transform can process. It is not being used to decide whether the chunk is shared.

So the change is essentially: use Vite’s actual naming condition rather than inferring it from CSS and module count.

katywings: Thank you, but to be honest, I am not willing to debate with an LLM or its output 😅, especially if it doesn't link to the sources of it's understanding, for validation.

As to why I didn't just use `facadeModuleId` as sole signal: Generally speaking you want to apply this special mapping to as few files as possible. If you know about a specific case you can add it to the css fixture as a test and then fix the matching case by case. That way you avoid over-matching unrelated cases and creating new bugs / regressions. Right now I don't wanna break our css implementation with what might be days before we release stable, so I used defensive programming. My main knowledge about facadeModuleId comes from https://github.com/rollup/rollup/issues/3772#issuecomment-689955168. Vite 8 & Rolldown might apply slightly different logic and I didn't have the time to go through the whole code of those today.

birkskyum (approved):

PR #2276, opened 2026-08-01
feat: implement css.filter plugin option

katywings:

1. Bug: File paths with special characters such as `+` are skipped during CSS asset collection in development
2. Feature: CSS Files in `node_modules` are always skipped in development asset collection, resulting in FOUC's

## What is the new behavior?

1. File paths with special characters such as `+` are now being processed properly in the assets manifest
2. The vite plugin has a new `css.filter` option analog to `serverFunctions.filter`, e.g. allowing the inclusion of certain dependencies

## Other information

- The reason for both commits being in one PR is that `pnpm` uses `+` in dependency paths. It was just easier to implement the `css.filter` feature, with the `+`-bug already being fixed 😅.
- The `css` fixture has a new `lazy+` test, making sure that regressions related to special characters will show up in the future

### CSS fixture before fix

Screenshot_20260801_155607

### CSS fixture after fix

Screenshot_20260801_160106

### DIFF
--- a/.changeset/stale-goats-go.md
+++ b/.changeset/stale-goats-go.md
@@ -1,0 +1,16 @@
+---
+"@solidjs/start": minor
+---
+
+The file filter logic used for CSS crawling in development can now be configured with the vite plugin option `css.filter` analog to `serverFunctions.filter`:
+
+```ts
+solidStart({
+  css: {
+    filter: {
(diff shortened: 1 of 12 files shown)


katywings:
P.S. @ryansolid This also might be a great addition for the Solid v2 manifest 😅

birkskyum (approved):

PR #2282, opened 2026-08-04
fix: dev toolbar throwing SourceMapConsumer import error on every dev page load

adipascu:
## PR Checklist

With the dev toolbar enabled, every dev page load throws:

```text
The requested module '/@fs/.../source-map-js/source-map.js' does not provide an export named 'SourceMapConsumer'
```

`source-map-js` and `error-stack-parser` are CommonJS with neither an `exports` nor a `module` field, so Vite serves them untransformed unless they are pre-bundled. `ErrorBoundary` imports `DevToolbar` statically, so this fires on healthy pages rather than only on errors.

This is a v2 regression. #1231 added both to `optimizeDeps` in 2024 and every published v1 still carries it. The v2 `src/config/index.ts` is a rewrite that never did, so every v2 release is affected, `2.0.0-alpha.0` through `2.0.0`.

## What is the new behavior?

Both are pre-bundled again, on the client environment, gated on `devOverlay`.

## Other information

The `@solidjs/start > ` prefix is required. Both are dependencies of `@solidjs/start` rather than of the app, so a bare specifier does not resolve from the project root under pnpm and Vite skips the entry silently.

---

Aside: I am currently available for web contracting or full-time work. Contact: adrian@pascu.be.

### DIFF
--- a/packages/start/src/config/index.ts
+++ b/packages/start/src/config/index.ts
@@ -199,6 +199,11 @@ export interface SolidStartOptions {
   };
 }
 
+const DEV_TOOLBAR_COMMONJS_DEPENDENCIES = [
+  "@solidjs/start > source-map-js",
+  "@solidjs/start > error-stack-parser",
+];
+
 const absolute = (path: string, root: string) =>
   path ? (isAbsolute(path) ? path : join(root, path)) : path;
 
@@ -291,6 +296,9 @@ export function solidStart(options?: SolidStartOptions): Array<PluginOption> {
           environments: {
             [VITE_ENVIRONMENTS.client]: {
               consumer: "client",
+              ...(start.devOverlay
+                ? { optimizeDeps: { include: DEV_TOOLBAR_COMMONJS_DEPENDENCIES } }
+                : {}),
               build: {
                 write: true,
                 manifest: true,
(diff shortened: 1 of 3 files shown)


packages/start/src/config/index.ts
```diff
@@ -199,6 +199,11 @@ export interface SolidStartOptions {
   };
 }
 
+const DEV_TOOLBAR_COMMONJS_DEPENDENCIES = [
+  "@solidjs/start > source-map-js",
+  "@solidjs/start > error-stack-parser",
+];
```
brenelz, line 205: is there a better fix than this i wonder

lxsmnsyc: currently I don't think so. There are no good alternatives imo, so this is fine

brenelz (approved):

brenelz:
Is this good to go then?

adipascu:
@brenelz good to go. If you'd rather drop the CommonJS dependencies from the toolbar entirely instead, I can do that as a follow-up.

lxsmnsyc:
@adipascu if there's a good alternative for source-map processing, that would be great

adipascu:
@lxsmnsyc There is: `@jridgewell/trace-mapping` for the source maps and `error-stack-parser-es` for the stack parsing, both ESM. I would merge the current PR as is and swap the libraries in a follow-up PR.

lxsmnsyc:
@adipascu I've made the swap in #2308 except the error stack parser

lxsmnsyc (approved):

PR #2304, opened 2026-08-20
fix: defer hydration until suspense stream is parsed

### LINKED ISSUE #2297: [Bug]: two pages render simultaneously
### Duplicates

### Latest version

### Current behavior 😯

_No response_

### Expected behavior 🤔

_No response_

### Steps to reproduce 🕹

```tsx
import { createAsync, query, Route, Router } from "@solidjs/router";
import { For, Suspense } from "solid-js";
const getVideos$ = query(async () => {
  "use server";
  await new Promise((r) => setTimeout(r, 1000));

  return [{ title: "hi", thumb: "https://picsum.photos/200/300" }];
}, "videos");

const getVideo$ = query(async () => {
  "use server";
  await new Promise((r) => setTimeout(r, 1000));
  return { title: "hi", thumb: "https://picsum.photos/200/300" };
}, "video");

  return (
    <Router>
      <Route
        path="/"
        component={() => {
          const videos = createAsync(() => getVideos$());

          return (
            <Suspense fallback={<p>Loading...</p>}>
              <For each={videos()}>
                {(video) => (
                  <a href="/video">
                    <h1>{video.title}</h1>
                    <img src={video.thumb} />
                  </a>
                )}
              </For>
            </Suspense>
          );
        }}
      />
      <Route
        path="/video"
        component={() => {
          const video = createAsync(() => getVideo$());
          return (
            <Suspense fallback={<p>Loading...</p>}>
              {video()?.title}
              <img src={video()?.thumb} />
            </Suspense>
          );
        }}
      />
    </Router>
  );
}
```
do exactly what's in the video:

https://github.com/user-attachments/assets/94d33e7e-09e7-4846-a799-4f2b6385bb25

### Context 🔦

_No response_

### Your environment 🌎

```shell

```

birkskyum:
## PR Checklist

- [x] Addresses an existing open issue: fixes #2297

The client entry module is marked `async`, so it can hydrate while the HTML parser is still waiting for streamed Suspense content. Hydration can retain references to fallback route DOM that the stream later replaces. After reloading a detail route and navigating back, both the stale detail route and the new root route can remain rendered.

## What is the new behavior?

The client module now uses the default deferred execution of module scripts. It is still discovered and fetched when the parser encounters it, but hydration waits until the streamed document has been parsed.

A Playwright regression reproduces the reload-and-back sequence from #2297 and verifies that only the root route remains.

## Other information

Validation completed:

- `pnpm --filter @solidjs/start typecheck`
- `pnpm --filter @solidjs/start test:ci` (98 tests)
- `pnpm --filter tests build`
- `pnpm --filter tests unit:ci` (5 tests)
- Playwright Chromium e2e suite (38 tests)
- Focused regression in standard and bundled-dev configurations

### DIFF
--- a/packages/start/src/server/StartServer.tsx
+++ b/packages/start/src/server/StartServer.tsx
@@ -34,7 +34,6 @@ export function StartServer(props: { document: Component<DocumentComponentProps>
               <script
                 type="module"
                 nonce={nonce}
-                async
                 src={getSsrManifest("client").path(import.meta.env.START_CLIENT_ENTRY_URL)}
               />
             </>
--- a/.changeset/calm-streams-hydrate.md
+++ b/.changeset/calm-streams-hydrate.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/start": patch
+---
+
+Defer client hydration until streamed Suspense content has been parsed to prevent stale route DOM after history navigation
--- a/apps/tests/src/app.tsx
+++ b/apps/tests/src/app.tsx
@@ -2,9 +2,11 @@ import { MetaProvider, Title } from "@solidjs/meta";
 import { Router } from "@solidjs/router";
 import { FileRoutes } from "@solidjs/start/router";
 import { Suspense } from "solid-js";
+import { getRequestEvent, isServer } from "solid-js/web";
 import "./app.css";
+import Repro2297App from "./repro-2297-app";
 
-export default function App() {
+function OriginalApp() {
   return (
     <Router
       root={props => (
@@ -69,4 +71,12 @@ export default function App() {
       <FileRoutes />
     </Router>
   );
+}
+
+export default function App() {
+  const pathname = isServer
+    ? new URL(getRequestEvent()!.request.url).pathname
+    : window.location.pathname;
+
+  return pathname.startsWith("/suspense-back") ? <Repro2297App /> : <OriginalApp />;
 }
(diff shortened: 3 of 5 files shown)


packages/start/src/server/StartServer.tsx
```diff
@@ -34,7 +34,6 @@ export function StartServer(props: { document: Component<DocumentComponentProps>
               <script
                 type="module"
                 nonce={nonce}
-                async
```
brenelz, line 37: i feel usually this is async. This is what was causing problems?

birkskyum: Seems like it @huseeiin  can you verify this works?

huseeiin: > Seems like it @huseeiin can you verify this works?

yes its fixed

ryansolid:
DO NOT MERGE.

This is indictive of another bug.. a core bug most likely and probably one annoying to fix. But fundamentally we definitely do not want to block hydration until document completion. It basically negates the value of streaming as the uncanny valley covers the whole time window. Script entry should be async not deferred.

In 2.0 we have smart snapshotting. In 1.x early interaction is supposed to bail out of hydration and resume the boundary as client rendering. But in all cases hydrating as soon as possible is definitely desired. This is part of the design contract.

So glad I noticed before merge. Vercel accidentally did this on their first RSC edge streaming demos and people attributed it to Server Components being slow. React core team scrambled to figure out the issue. I actually identified it for them in their PR. Honestly Im not sure perception ever completely recovered from that.

ryansolid:
If guys want .. make an issue against Dom Expressions probably with the context of the original bug this was trying to fix.

PR #2318, opened 2026-09-11
fix(directives): report server functions that cannot work

lxsmnsyc:
A server function is moved to the top level of its module, and several shapes stop working once it is. They compiled without complaint before.

- Fail the build when a server function reads a variable from an enclosing function, uses `this` or `arguments` in an arrow, uses `super`, or reads a private class member.
- Fail the build when the directive is in an object or class method. The transform ignored it, which shipped the method body and the modules it imports to the browser.
- Warn when a directive string is not the first statement of a module or a function body, where it has no effect.
- Warn for each export a "use server" module cannot serve, naming it. Those exports are still left out of the client build.
- Support an anonymous default export from a "use server" module.
- Compile server functions in `.mts` and `.cts` files.
- Point build errors at the full path of the file.


lxsmnsyc:
Also replaces the `getDescriptiveName` for `getHierarchicalName`, which fixes an issue where two server functions of the same name but of different scopes (e.g. FooComponent > foo vs BarComponent > foo) gets a compiler conflict since the generated ids matches .

brenelz (approved):

Issue #1132, opened 2023-11-28
[Bug]: Solid-Start should not reuse Vite's "base" setting for the router


indeyets:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

Vite's "base" setting, which is originally intended as the base-url for assets is reused for solid-router here: https://github.com/solidjs/solid-start/blame/main/packages/start/entry-client/StartClient.tsx#L98

current behaviour is originally defined by @atk and then fine-tuned by @devinrhode2 

### Expected behavior 🤔

Solid-Start should provide a separate independent setting for router's base-path to support cases when assets are served from one path, but actual app is served from another.

### Steps to reproduce 🕹

Steps (sorry, its "fake" data, as I don't know how to make real-life reproduction for this):

1. Set "base" setting in "vite.config.ts" to `https://cdn.example.com/custom/path`
2. Deploy app to `https://app.example.com/`
3. Open the page
4. See that all of dynamic JS-based elements defined "inside" of router are NOT dynamic (hydration didn't run)

### Context 🔦

My situation looks like this:

We serve assets (js, css, … files) from the CDN and use both custom host and custom path-prefix there. Vite's "base" setting is perfect for configuring this.

So, for example, it looks like: `https://cdn.example.com/custom/path`

Solid start reads this setting and reuses the path for the base setting of solid-router.

But the problem is, that the app is served from: `https://app.example.com/` and `base=/custom/path` SILENTLY breaks the routing and while server-side renders the pages, client-side doesn't even try to hydrate them. No error or warning here.

indeyets:
my temporary local solutions is this [YARN patch](https://yarnpkg.com/cli/patch):

```patch
diff --git a/entry-client/StartClient.tsx b/entry-client/StartClient.tsx
index a9b6ecb4bdb13d07ceae0d1246745ee527638b4f..e8d551f9b6e5609f3c20d830dafd9a4af99655a0 100644
--- a/entry-client/StartClient.tsx
+++ b/entry-client/StartClient.tsx
@@ -95,7 +95,7 @@ export default ({
     );
   }
 
-  const BASE_URL = import.meta.env.BASE_URL;
+  const BASE_URL = "/"; 
   let basePath = BASE_URL;
   if (BASE_URL.startsWith("http")) {
     try {
```

devinrhode2:
[...]

PR #2319, opened 2026-09-11
CDN assets support

indeyets:
## PR Checklist

Vite accepts a full URL as `base` (`https://cdn.example.com/some/prefix/`) so that a build addresses its assets on a CDN. With @solidjs/start 2.0.5 that config produces a broken build:

- The SSR manifest joins `BASE_URL` with the asset path via `pathe.join`, which folds `https://` into `https:/`. Every `<link>`, modulepreload, the entry `<script>` and the serialized manifest come out as `https:/cdn.example.com/…`. The prefixing itself was added in #2305; it only works for a path base.
- Server functions post to `${BASE_URL}_server`, i.e. to the CDN. The no-JS form fallback and the single-flight referer resolve against `origin + BASE_URL`, which for a full URL yields hosts like `http://app.example.comhttps/…`.
- `SERVER_BASE_URL` has existed since #2218 as the app's mount path, but only API route matching read it, and only when `server.baseURL` was set explicitly. Otherwise it was `""`, and everything fell back to the asset base.

Related: #1132 (closed, 1.x) described the same conflation of asset base and app base.

## What is the new behavior?

`BASE_URL` addresses assets only. `SERVER_BASE_URL` is the app's mount path and is always defined:

- `server.baseURL` if set;
- otherwise `/` when Vite's `base` is a full or protocol-relative URL;
- otherwise Vite's `base` itself (a plain path).

It is always wrapped in slashes, so relative redirect `Location`s resolve under it. Server functions, no-JS redirects, single flight and API route matching all use it. The SSR manifest joins asset paths with a URL-aware helper, so a full-URL base comes through intact. The type is published via `@solidjs/start/env`, since apps pass it to `<Router base>` (the documented setup for an app under a path).

Verified with a build using `base: "https://cdn.example.com/"`: all asset tags, imported assets and CSS `url()`s point at the CDN, `_server` stays at `/`. With `base: "/app/"` everything stays under `/app/` as before.

---

The changes are split into six commits, each meant to be reviewed on its own; the manifest fix comes first and is independent of the rest.


lxsmnsyc (approved):

packages/start/src/config/index.ts
```diff
@@ -202,6 +202,27 @@ export interface SolidStartOptions {
 const absolute = (path: string, root: string) =>
   path ? (isAbsolute(path) ? path : join(root, path)) : path;
 
+// the same test Vite applies to an external base
+const externalUrlRE = /^([a-z]+:)?\/\//;
+
+// A mount path is always wrapped in slashes: `new URL("page", origin + "/app")`
+// resolves to /page, losing the segment, and a bare "app/" would glue onto
+// the origin.
+const withSlashes = (path: string) =>
```
lxsmnsyc, line 211: not sure if this needs to handle an empty string, might not be an issue.

brenelz:
conflicts here now

packages/start/src/config/index.ts
indeyets: I don't think it's possible to have empty string there, but I handled the case anyway. future-proofness. thank you! 👍

indeyets:
@brenelz done ✅

PR #2320, opened 2026-09-16
fix(fns): guard the no-JS flash cookie against non-form arguments

lxsmnsyc:
A no-JS server function POST with an empty or non-form body returned a 500. The
flash cookie called `.entries()` on an argument that was not a FormData and
threw, and the same code runs as the error handler, so it rethrew and the
request died.

Changes:
- Spread `.entries()` only when the last argument is a FormData.
- Wrap the flash cookie build so a bad value degrades to a plain redirect.
- Add regression tests for a non-form body, an empty body, and the form case.


brenelz (approved):

brenelz:
## Review

**Intent:** No-JS server function POSTs with empty/non-form bodies were 500ing because the flash cookie build called `.entries()` on a non-FormData last arg — and that same path is the error handler, so it rethrew. Guarding the entries call and wrapping cookie serialize in try/catch is the right shape.

**Verdict:** No blockers. Already approved; leaving the written notes here.

### Blockers
None. CI green.

### Nits
1. **Duck-type `last?.entries`** — correctly covers both `FormData` and `URLSearchParams` (urlencoded goes through `extractBody` as the latter). Prefer `instanceof FormData || instanceof URLSearchParams` so a stray `Array`/`Map` doesn’t get turned into index pairs. Unlikely on this path, but clearer.
2. Tests cover non-form, empty, and form echo — good. Optional: on the crash cases, assert there is **no** `flash=` Set-Cookie (or that `input` is empty) so the “best effort” path stays honest.

### What’s strong
- Root cause matches the code; try/catch correctly protects the redirect/error path.
- Changeset scoped to `@solidjs/start`.
- Regression tests hit the real no-JS gate (no `X-Server-Instance`).

